PR #16, opened 2017-04-30
change directory mode before deleting them.

ardovm:
Apparently, rdiff-backup does not take into account the possibility that folders could not be deletable.
This pull request is about adding a call to os.chmod(), that should render modifiable any directory, before attempting to delete it.

Please note that I only tested this patch under Win32.

A simple use case is with folders with custom icons. When the user sets a custom icon on a folder, Windows also sets the "system" attribute on the folder.

I hope this can be useful.


ardovm:
Closed, as I used a wrong branch. I apologize for the noise.

PR #28, opened 2018-03-11
add iso, vob and mdf to the --no-compression list

orangenschalen:
Extend the default --no-compression file extension list by some more
common extensions that are used for DVD backups.

As proposed by
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=607257

### DIFF
--- a/src/rdiff_backup/Globals.py
+++ b/src/rdiff_backup/Globals.py
@@ -191,8 +191,8 @@ compression = 1
 # compiled version of the first.
 no_compression_regexp_string = (
     b"(?i).*\\.(gz|z|bz|bz2|tgz|zip|rpm|deb|"
-    b"jpg|jpeg|gif|png|jp2|mp3|ogg|avi|wmv|mpeg|mpg|rm|mov|flac|shn|pgp|"
-    b"gpg|rz|lzh|zoo|lharc|rar|arj|asc)$")
+    b"jpg|jpeg|gif|png|jp2|mp3|mp4|ogg|avi|wmv|mpeg|mpg|rm|mov|flac|shn|pgp|"
+    b"gpg|rz|lzh|zoo|lharc|rar|arj|asc|vob|mdf)$")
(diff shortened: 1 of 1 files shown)


anomaly256:
Thanks for the PR, though I'm not sure adding iso to the list is necessarily a good idea.  The majority of iso images I encounter are data volumes and benefit from compression specially the kinds of iso files people wish to back up.  Compressed media isos (DVD Video for example) tend to be the exception not the norm I think.

orangenschalen:
You are right, while "my isos" are usually compressed, that is not a sound assumption in general.

I've removed .iso from the list. Is the patch acceptable that way, or should I refrain from extending the default list?

ottok:
This patch was also suggested in Debian as https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=607257

ottok:
Thanks for contributing to rdiff-backup!

As you might have noticed, the project has been stale for some years now. We intend to revitalize it now. I hope you can be involved in the effort!

The biggest ongoing change is the rewrite to Python 3 (in https://github.com/rdiff-backup/rdiff-backup/pull/40) and once that is merged we will have to ask you to rebase this contribution on the new Python 3 code base. Thanks for understanding and your help!

ottok:
Hello!

Most of the code base is now rewritten in Python 3. We also have a CI pipeline now. Could you please rebase this PR on the latest upstream master?

We would also be very grateful if you want to contribute to the project fixing other issues as well besides this PR.

orangenschalen:
Yes, I'll rebase it against master, it'll just take a few days. Nice to
see rdiff-backup to come back to life. Thanks a lot for you work!

orangenschalen:
The branch is rebased, the only conflict was the 'b' prefix for Bytestream vs. String literals. Now they are proper bytestreams.

ericzolf:
Thanks for the contribution! Let's stay conservative :-)

Issue #58, opened 2019-08-13
Set up rdiff-backup on PyPi


ottok:
The package rdiff-backup is listed at https://pypi.org/project/rdiff-backup/ but no versions after 0.1.1 have been released and the author anyway is unknown. We should try to get access to this package on PyPi and integrate it to be automatically published from tags released on Github.

ottok:
Current PyPi maintainer @AvdN responded to me by e-mail and assured this will be fixed once the new people doing the Python 3 rewrite have established themselves and #52 has been fixed.

AvdN:
I have added Otto as maintainer on PyPI based on the data that at least one of the members on savannah.nongnu.org (Andrew Foster) has been involved in the mailing list thread and that the project page there has been updated with information on the move of the project.

ottok:
Seems Travis-CI has PyPi integration, so we could push each tagged master branch there automatically in the future: https://docs.travis-ci.com/user/deployment/pypi/

ericzolf:
I wouldn't want to rely too much on Travis' capabilities as it makes us dependent on GitHub/Microsoft. AFAICJ `setup.py upload` does the trick as well, could be called from within Travis but also outside of it.

ikus060:
`setup.py upload` is not recommended to upload something to pypi.

The preferred and simple way is to use twine. Also as I mention mutiple
time, the automated build running by travis is not valid for pypi and it's
not generating the binaries using manylinux. Will need to address this
first. Pypa repository provide a very good example to get started with
manylinux. And allow to quickly generate the binaries for all the python
versions.

--
Patrik Dufresne Service Logiciel inc.
http://www.patrikdufresne.com <http://patrikdufresne.com/>/
514-971-6442
130 rue Doris
St-Colomban, QC J5K 1T9

ikus060:
I want to bring this topic back and try to follow a bit more our contribution guide line. I want to improve the current travis CICD pipeline to build and test packages compatible with PYPi.

Since rdiff-backup has C modules, the recommended way is to use manylinux images to compile & test the pacakges.

Here some of the highlight changes I'm proposing:
1. Replace the debian:sid docker image by manylinux images
2. Remove the Dockerfile
3. Change travis pipeline to have stages:
  * test : where all our tox.ini test runs with a matrix for py35,36,37
[...]

PR #87, opened 2019-08-16
Take package and distribution version from Git

### LINKED ISSUE #66: Get rid of $version and replace with other mean to centralize versioning
<a href="https://github.com/ericzolf"></img></a> **Issue by [ericzolf](https://github.com/ericzolf)**
_Sunday Apr 14, 2019 at 11:06 GMT_
_Originally opened as https://github.com/ericzolf/rdiff-backup/issues/8_

----

Sounds to me like an SVN-ism that can't work with Git.

ericzolf:
Replace distutils.core with setuptools
Introduce setuptools_scm (thanks to ikus060 for the hint)
We must now tag releases with vA.B.C for it to work
Closes #66


ericzolf:
`./setup.py --version` gives a version dependent on the state of the repository, and based on the latest tag. After a `./setup.py build`, `PATH=$PWD/build/scripts-3.7:$PATH PYTHONPATH=$PWD/build/lib.linux-x86_64-3.7 python3 rdiff-backup --version` also gives a new version; you can try to set the tag `git tag v1.3.5`, touch a file and see the variations (don't forget to do `git tag -d v1.3.5` after). `./setup.py sdist` also creates a correctly versioned tarball.

ottok:
I don't understand how the overall process is supposed to work. So when we make release 2.0.0 we tag version v2.0.0 and release a .tag.gz package with the source. When that tar.gz package is imported to Debian or PyPi, how will the build then know what version it was since the .git/tags are not there anymore?

How do we make development versions? If we tag v1.9.0 in git, then Github (and other tools) will automatically make releases of those and systems will start to upgrade since v.1.9.0 is bigger than current stable release v1.3.3.

ericzolf:
I would ask @ikus060 to comment on this one as he surely has more experience than I do.

What I can observe:

* The way I did it, the version is only (re-)generated if you use `setup.py`, rdiff-backup itself gets the version statically from the generated `src/rdiff-backup/Version.py` file.
* PyPI shouldn't be the issue because `setup.py` can directly push the module with the correct version directly from Git (so what I understand from `setup.py --help-commands`).
* Debian (and any RPM...) - I'm not sure, `./setup.py sdist` packages the `Version.py` generated but I'm not sure it's used or generated if you do `./setup.py install` (which you probably do as part of your packaging, I guess)... Possibly we need to set a default version (there is an option) and we use Git only if it exists.

OK, let me mark this PR as WiP until we've clarified all aspects.

ottok:
I reviewed a bunch of random Python packages that are both in Debian and available via PyPi. Most of them seem to simply define a `__version__` in their main file and then use that variable elsewhere, including in `setup.py`.

```
black/black.py:__version__ = "19.3b0"
black/setup.py:    _version_re = re.compile(r"__version__\s+=\s+(?P<version>.*)")

membernator/membernator.py:__version__ = "1.0.1"
membernator/membernator.py:    args = docopt(__doc__, version="membernator %s" % __version__)
membernator/setup.py:from membernator import __version__
membernator/setup.py:    version=__version__,

radon/docs/conf.py:version = radon.__version__
radon/radon/__init__.py:__version__ = '3.0.3'
radon/radon/cli/__init__.py:program = Program(version=sys.modules['radon'].__version__)
radon/radon/complexity.py:    version = __import__('radon').__version__
radon/setup.py:      version=radon.__version__,
```

Some hade a separate VERSION.txt and others had a VERSION variable. By far however the design pattern above was the most common and I would recommend adhering to it.

ericzolf:
OK, I created a source package using `./setup.py sdist`, took it somewhere else, unpacked it, and the version remains the same even without Git around. So for me it sounds like it gives us everything we want, especially in regard to traceability, as it worked with or without a recent tag in the correct format.

@ottok can't you quickly verify that it also works for a Debian package?

ericzolf:
Oh, and it means we only need to care about the version when we really want to create a release, and we do this by tagging the Git repo. It can't get easier in my opinion.

We just need to adapt our versioning scheme:
- tag v1.3.5 → "official" version 1.3.5
- no new tag, add commits and do tests → nightly or debug or test or snapshot version 1.3.6.dev0+g889ee14.d20190818 (for example)
- next version must then be 1.3.7 to make sure that the version numbers are growing, but that would make sense (and not a must because the debug versions haven't been officially released)

Exactly the same of course with 2.0.0, we could actually agree to only release _even_ versions and the _odd_ ones are reserved for non official versions.

ikus060:
I did not follow the full story, but if you are still using scm_version, it working nicely. When generating a package either binary or source distribution, scm_version does it a trick to keep track of the version in both pacakges. Same goes when pushing to pypi or generating RPM or DEB. 
The only inconvenience is building from source. Users cannot download a tar.gz from github and run `python setup.py xxx`, The users must git clone the repository.

ottok:
Related, I came across versioneer used in https://github.com/warner/magic-wormhole/blob/master/versioneer.py – in it's `setup.py sdist`  a `_version.py` is created and included in the distributed .tar.gz source file (example: https://salsa.debian.org/debian/magic-wormhole/blob/debian/src/wormhole/_version.py).

This project however does not build any RPM or DEB packages directly at upstream from upstream sources, so it does not show an example how to solve the 'version not found in any file' problem for nightly builds.

ericzolf:
I'm not sure how you do the nightly builds, aren't you doing them on the basis of Git directly? And/or on the basis of the output of `setup.py sdist/bdist`?
In this case, the secret is actually according to my research a file `PKG-INFO` which is "injected" into the tarball. That should IMHO solve all our issues at once.

ottok:
Which version for test builds will the CI be using in this scheme? The latest git tag version + date + commit id maybe?

I am still inclined to use a actual explicit version file in the sources in git somewhere, but I see that e.g. wormhole uses Versioneer and they manage to package it in Debian, so probably this concept is OK then. Versioneer CI does not make deb packages for every commit tough. I plan to do that as part of our CI.

ericzolf:
Probably without date as date is only if your work tree is dirty, but else something like this.

And exactly if you're building packages on the fly should they have a unique version or we might get bored with support requests of nightly for which we wouldn't know the actual state used, as we wouldn't be able to change any explicit version on a daily basis.

ottok:
Date is needed to ensure old test systems update to newer test releases.

For reference, this is how it works for MariaDB:
- upstream sources has one file that defines version, eg. 10.3.17
- then 10.3.17 is released, that file is bumped to include 10.3.18. So while in development, all test builds will have 10.3.18 before the real 10.3.18 is made.
- test builds and test debs will have version string 10.3.18~<date>-<short commit hash>
- any system running 10.3.17 will experience an update if it sees 10.3.18~<date>-<short commit hash>
- any system with 10.3.18~<date>-<short commit hash> will update if it sees are 10.3.18~<date>-<short commit hash> release where <date> is bigger
- any system with 10.3.18~<date>-<short commit hash>  will update to 10.3.18 if it sees it, as the tilde ~ is considered a pre-release sign and X is bigger than X~something

If we use a git tag as the version number, it makes the test build complicated, as the test builds don't know what version is about to be prepared and what they need to simulate

ericzolf:
Sorry wrong statement: between tag/version and commit  ID there is a commit distance to make sure the version grows.
Also I'd like to note that ~ is a package debianism and isn't generically understood (and should IMHO be reserved for distribution packages and not for upstream, which we are).

ericzolf:
Issue #115 shows the kind of problem I'm trying to address with this PR: user has used the "latest version from Git", no clue which version it is, and it will become worse when we start to have nightlies.

ottok:
I don't know how PyPi handles versions (#58) but if the nightly version is installed via apt, then the version number will be easy to check. In issue 115 the real probles is however that the reported did not report the version in any explicit way.

ericzolf:
But these are not arguments against merging this PR, I've addressed all your requirements and I'm slowly sick re-merging again and again this PR, so what's the actual issue?

src/rdiff_backup/Globals.py
```diff
@@ -21,8 +21,10 @@
 import re
 import os
 
+from . import Version
```
ericzolf: Once your personal / packaging requirements are clearer we can add here a try/except and do some magic in case the Version.py file doesn't exist.

PR #91, opened 2019-08-19
Master fix librsynctest for librsync 0.9.7

bigbear3001:
This includes https://github.com/rdiff-backup/rdiff-backup/pull/90

Currently the librsynctest.py tests fail on Ubuntu 18.04 LTS because the rdiff command has no -H parameter yet.

This fix checks in the help page of the rdiff command if there is any mention of `-H` and runs the test the old way in case there is no support for `-H`

### DIFF
--- a/DEVELOP.adoc
+++ b/DEVELOP.adoc
@@ -7,7 +7,7 @@ Some notes for developers and other people willing to help development.
 - Install:
 * tox
 * libacl-devel (for sys/acl.h)
-* rdiff (with support for -H parameter)
+* rdiff
 - unpack the test files previously available from
 https://download-mirror.savannah.gnu.org/releases/rdiff-backup/testfiles.tar.gz and now temporarily available as
 release under https://github.com/ericzolf/rdiff-backup/releases[]:
--- a/testing/librsynctest.py
+++ b/testing/librsynctest.py
@@ -1,4 +1,4 @@
-import unittest, random
+import unittest, random, subprocess
 from commontest import *
 from rdiff_backup import librsync, log
 
@@ -32,8 +32,12 @@ class LibrsyncTest(unittest.TestCase):
 		for i in range(iterations):
 			MakeRandomFile(self.basis.path, file_len)
 			self._clean_file(self.sig)
-			assert not os.system(b"rdiff -b %i -H md4 signature %s %s" %
-								 (blocksize, self.basis.path, self.sig.path))
+			if b'-H' in subprocess.check_output(["rdiff", "--help"]):
+				assert not os.system(b"rdiff -b %i -H md4 signature %b %b" %
+									 (blocksize, self.basis.path, self.sig.path))
+			else:
+				assert not os.system(b"rdiff -b %i signature %b %b" %
+									 (blocksize, self.basis.path, self.sig.path))
 			with self.sig.open("rb") as fp:
 				rdiff_sig = fp.read()
 


testing/librsynctest.py
```diff
@@ -32,8 +32,12 @@ def sig_file_test_helper(self, blocksize, iterations, file_len = None):
 		for i in range(iterations):
 			MakeRandomFile(self.basis.path, file_len)
 			self._clean_file(self.sig)
-			assert not os.system(b"rdiff -b %i -H md4 signature %s %s" %
-								 (blocksize, self.basis.path, self.sig.path))
+			if b'-H' in subprocess.check_output(["rdiff", "--help"]):
+				assert not os.system("rdiff -b %i -H md4 signature %s %s" %
```
ericzolf: Needs to be bytes

testing/librsynctest.py
```diff
@@ -32,8 +32,12 @@ def sig_file_test_helper(self, blocksize, iterations, file_len = None):
...
+									 (blocksize, self.basis.path, self.sig.path))
+			else:
+				assert not os.system("rdiff -b %i signature %s %s" %
```
ericzolf: Needs to be bytes

testing/librsynctest.py
```diff
@@ -32,8 +32,12 @@ def sig_file_test_helper(self, blocksize, iterations, file_len = None):
 		for i in range(iterations):
 			MakeRandomFile(self.basis.path, file_len)
 			self._clean_file(self.sig)
-			assert not os.system(b"rdiff -b %i -H md4 signature %s %s" %
-								 (blocksize, self.basis.path, self.sig.path))
+			if b'-H' in subprocess.check_output(["rdiff", "--help"]):
```
ericzolf: Needs to be bytes (I assume). Python will not complain if you search for bytes in strings but it will also never find it.

DEVELOP.adoc
```diff
@@ -7,7 +7,7 @@ Some notes for developers and other people willing to help development.
 - Install:
 * tox
 * libacl-devel (for sys/acl.h)
-* rdiff (with support for -H parameter)
+* rdiff
```
ericzolf: We need to see but there might be a merge conflict as the diff is based on your other MR and not on master.

ericzolf: Forget it, your other change was already merged in master.

testing/librsynctest.py
bigbear3001: Why would `["rdiff", "--help"]` need to be bytes? 

(check_output returns bytes if you do not specify an encoding, see https://docs.python.org/3/library/subprocess.html#subprocess.check_output and https://docs.python.org/3/library/subprocess.html#subprocess.run (for stdout))

From my point of view (and my tests) everything that needs to be bytes (`b'-H'`) is already correct.

Maybe i missinterpreted your comment, could you elaborate what exactly in the statement you would like to see as bytes?

testing/librsynctest.py
bigbear3001: fixed the missing bytes in os.system call

testing/librsynctest.py
bigbear3001: fixed the missing bytes in os.system call

testing/librsynctest.py
ikus060: @ericzolf `subprocess.check_output(["rdiff", "--help"])` is returning bytes. Not str.
https://docs.python.org/3/library/subprocess.html#subprocess.check_output

Current implementation is ok

ericzolf: Ah, sorry, all over the place it's "bytes in, bytes out/str in, str out" but not here, you're right.

testing/librsynctest.py
ericzolf: It's a trifle but replace %s with %b: b'%s' is deprecated, but will not be removed during the 3.x series.
See https://docs.python.org/3/library/stdtypes.html#printf-style-bytes-formatting

testing/librsynctest.py
ericzolf: It's a trifle but replace %s with %b: b'%s' is deprecated, but will not be removed during the 3.x series.
See https://docs.python.org/3/library/stdtypes.html#printf-style-bytes-formatting

ericzolf:
I also just checked your code and it looks good, just fix the %s/%b thingy (yes, I know, I make you fix things that were wrong before, but this way the code becomes better) and I'll merge:

    git remote add bigbear3001 https://github.com/bigbear3001/rdiff-backup.git
    git fetch bigbear3001 
    git checkout master-fix_librsynctest_librsync_097
    git diff master  # looks good
    ./setup.py build
    PATH=$PWD/build/scripts-3.7:$PATH PYTHONPATH=$PWD/build/lib.linux-x86_64-3.7 python3 testing/librsynctest.py # success

testing/librsynctest.py
bigbear3001: didn't know the `b'%s'` deprecation, thanks for the hint, fixed in the touched lines

PR #92, opened 2019-08-19
Fix selectiontest where /boot is not a seperate mount

bigbear3001:
EFI systems now often have /boot/efi as seperate mountpoint instead of
/boot

Currently the selectiontest fails on systems where /boot is not its own mountpoint. On my Notebook (Tuxedo Infinitybook) /boot/efi is now the seperate mountpoint.

With this commit we check that /boot is actually mounted as a seperate mountpoint before asserting that it is detected as a different filesystem

### DIFF
--- a/testing/selectiontest.py
+++ b/testing/selectiontest.py
@@ -1,5 +1,4 @@
-
-import re, io, unittest, types, os
+import re, io, unittest, types, os, subprocess
 from commontest import *
 from rdiff_backup.selection import *
 from rdiff_backup import Globals, rpath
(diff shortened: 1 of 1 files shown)


ericzolf:
Actually the 2nd commit doesn't look good to me "merge master into branch". What did you try to do?

You should always rebase your branch upon master (Otto described how in another pull request) and not the other way around (if it was what you were trying to do), as this is messing up our history in strange ways. Else the test stayed successful on my computer so fine to merge.

bigbear3001:
@ericzolf will do a rebase. i did resolve the conflict online on github.com that creates a merge commit.

PR #93, opened 2019-08-20
Rename README to be a Markdown file

ottok:
This will allow more styling, headers, links and screenshots in the future.


ikus060:
I would way we should reformat the README to markdown. Otherwise, I don't get the benefit of renaming the file only.

ericzolf:
There are quite a lot of paths and command inline in the text and they should be marked as code (e.g. `/usr` and `--prefix`). After that no issue to merge (even if I had started to use Asciidoc which is even more powerful than Markdown).

ottok:
The point with Markdown is that Github will automatically render it.

Current Github main repo page:
image

After this change:
image


I did not edit the content of the README. It surely needs and overhaul but this PR is just about changing the format. I don't know if AsciiDoc would be better but at least it seems that Github renders it as well. I have no prior experience or knowledge of AsciiDoc.

afoster:
You may also find some useful words in https://github.com/rdiff-backup/rdiff-backup/blob/8ccc5a3b44c996ecd810f8d5d586d0da6435cc32/README.md which has since been deleted. 

My original thought was that it was more useful to explain what rdiff-backup is and some more common ways to install for users rather than jumping straight into build-from-source stuff. But I might be overloading this PR.

ottok:
To me https://github.com/rdiff-backup/rdiff-backup/blob/8ccc5a3b44c996ecd810f8d5d586d0da6435cc32/README.md looks good. Please make a new PR to apply those changes on top of this and master.

PR #95, opened 2019-08-20
Extend Tox to include flake8 static Python code analysis

ottok:
We want our code to be PEP-8 compliant, right?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,7 @@
 # Use tox_slow.ini for longer running tests.
 
 [tox]
-envlist = py35, py36, py37, py38
+envlist = py35, py36, py37, py38, flake8
 
 [testenv]
 deps =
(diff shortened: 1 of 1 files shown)


ericzolf:
Two comments:
- build and dist must be ignored, as well as .tox.root and .package.root (and .package).
- this raises the question if it's not easier to just point flake8 at:
1. src
2. testing
3. setup.py
4. rdiff-backup and rdiff-backup-statistics (both would actually belong under `src`)
5. misc (optional)
- shouldn't we wait until the files are clean before we introduce the test? Currently it's pretty much useless (and _very_ verbose)/

ottok:
Flake8 automatically scans only Python files, so we do not need to manually ignore unrelated non-Python folders. Ignoring should be used only for paths that have Python we want to ignore, or paths that are very large and we want to save I/O by not scanning them at all. Those are already in the `exclude`.

I think it would be OK to merge this now, and then proceed with fixing the syntax stuff. We cannot however merge the Travis-CI integration (#96) until the code is clean. Code can be made clean quickly en mass with YAPF,

ericzolf:
OK but:
1. `build` and `.tox.root` contain copied Python code and hence must be excluded as well.
2. I'm also using tox to check my code so I don't really wan't 1000s of lines of error, at least not without perspective to get it quickly fixed, which can only happen after my bytes path PR has been merged (don't make do a rebase of this one).

ottok:
Added `build` and `.tox.root` to excludes.

PR #94, opened 2019-08-20
Update Dockerfile with exact steps on how to run the Tox tests

ottok:

PR #99, opened 2019-08-23
Add .tox and build folder to docker ignore

bigbear3001:
To prevent docker copying those dirs into the context upon building
the docker images


bigbear3001:
added in #94

Issue #60, opened 2019-08-13
Clean-up the code with pep8 and replace tabs with spaces


ericzolf:
The code looks strange and more difficult to read and edit currently. That would be my next task after having got the bytes MR merged into master.

ericzolf:
I would proceed like this:

- first commit with non important files `yapf --recursive --parallel --in-place misc setup.py python-rdiff`
- second commit with the test files `yapf --recursive --parallel --in-place testing` after checking that it doesn't break the results
- third commit with the actual source files `yapf --recursive --parallel --in-place src`, also checking that the tests aren't broken

The question is when to do it as it would conflict with any on-going pull request.

ottok:
I suggest you do those changes now. There are no major code PRs pending, and the very old ones should be rebased on the new Python 3 codebase anyway soon.

ericzolf:
I need #110 to be merged before I can do it, so that I'm sure I start from a sane base (tox tests wise).

ottok:
This is now done for `testing/` and still needs to be done for `src/`. Any takers?

ottok:
PR for `misc/`: https://github.com/rdiff-backup/rdiff-backup/pull/114

ottok:
The PR #114 now also takes care of parts of `src/`. Great work by @elMor3no ! :muscle:

elMor3no:
Which others need to be format?

ericzolf:
That's an easy one :-)
Just run `tox -e flake8` and fix the code until tox and flake8 are happy and the test result is green.

ottok:
@elMor3no If you don't have Tox and Flake installed, you can also run `make test-static` to run the tests inside a Docker container. This is all documented in the sources. Just make sure you have the latest upstream master pulled to the source you are reading.

The test results, including Flake8, are also visible on Travis-CI.org, e.g. https://travis-ci.org/rdiff-backup/rdiff-backup/jobs/580715485

ericzolf:
Any progress on this topic? @elMor3no ?

elMor3no:
Ohh Sorry... I dont see this notification until now... My inbox is full of trash mail now... 
I will check this in a couple of days

ericzolf:
@elMor3no are you able to assign the issue to yourself in order to show that you're taking care of it?

[...]

PR #111, opened 2019-08-25
Make the test scripts PEP8/flake8 compatible

ericzolf:
Addresses #60 but only for the test scripts under `testing/` so that I can be sure that I don't break anything, then someone else can handle the actual code.

### DIFF
--- a/testing/FilenameMappingtest.py
+++ b/testing/FilenameMappingtest.py
@@ -1,60 +1,72 @@
 import unittest
-from commontest import *
+import os
+from commontest import MakeOutputDir, abs_output_dir, abs_test_dir, old_test_dir, \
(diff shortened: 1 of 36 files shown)


ottok:
This is marked "WIP" so I assume it is not ready for merging yet.

ericzolf:
The testing scripts are now all flake8-clean and you can review.

I had to push one commit to fix the tox.ini and ignore also W503 or W504, because both together make formatting of certain long lines mostly impossible. According to https://lintlyci.github.io/Flake8Rules/rules/W503.html and a few other pages, "line breaks should occur before the binary operator because it keeps all operators aligned" hence W503 can/should be ignored.

Issue #54, opened 2019-08-12
Decide where to store test files


ottok:
The next Tox framework uses test files from https://github.com/ericzolf/rdiff-backup/releases

Where should we host this in the future?

Options:
- in gh-pages branch as a tar.gz for download via rdiff-backup.net/testfiles.tar.gz to keep the master branch / main repo "clean" from binary stuff
- in master branch as a tar.gz and ensuring the test suite and the test data evolve lockstep in the same commits
- somewhere else?

Alternatively, do we even have to have the files as a binary package? Could it be stored in an uncompressed plain files way in the master branch and then some kind of test-prepare.sh script would set the file permissions and attributes (since git does not store them and the special attributes is required for the test files, as that is almost the whole point in them).

ericzolf:
I would keep it as I did it until now (but of course in this repo, not mine), as release with tag meaning basically, at this stage the code / tests worked with those testfiles.

Not a nice alternative but I think the others are worse until we have a better solution:

- packed or unpacked, the files are big and binaries, nothing I'd like to have in our main Git repo.
- unpacked in a separate Git repo: the issue is that the files also contain ACLs, extended attributes, hard and soft links, root and user rights, device & other non regular files, etc i.e. nothing with which Git could cope, and writing a script that would fix all this isn't that easy. And the problem with big binary files and Git remains.
- web page would be an alternative, but again it's a Git repo, etc...

My favoured approach would be to create a script/library which, based on a textual description e.g. YAML, could create any file structure so that we only need to describe which files with need with which characteristics and they would be created on the fly. Not an easy approach but the best IMHO, and the script itself would be small.

I had already started to think about it: https://github.com/ericzolf/rdiff-backup/issues/20 - wanted also to check if nobody else had the same idea.

davekempe:
We could use ansible to abstract that away perhaps? or perhaps whole vagrant boxes setup with local/remote different OSes and filesystems? Sol1 is happy to help with hosting - we are just in the process of upgrading our own hosting facilities, so soon should be able to have plenty of resources we can spare.

ericzolf:
I also thought about Ansible, I should it's my specialty in real life, but I just postponed the topic as it's a big one. It would be cool to have Windows, and MacOS, machines for testing.

But I'd split the topic:
- short term: stay as is, just move release to the official repo
- long term: introduce "test files as code"

ericzolf:
[...]

PR #116, opened 2019-09-01
Add Travis-CI integration to project

ottok:
Utilize Docker image and Makefile so that a developer running 'make'
can easily replicate the exact failures Travis-CI sees (if any).

Also test for PEP-8 coding style compatibility but allow failures.

Disable failing tests so the rest of the test suite passes.

Failures can be forbidden, tests extended and the scope of platforms
extended later on once we have this base set 100% working and good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+# Since we create our own virtualens via Tox, don't use any of Travis-CI Python
+# templates/features and could as well have "language: generic" here.
+language: python
+
+# Don't run tests on WIP branches
+branches:
+  except:
+    - /^WIP-.*$/
+
+# We utilize Docker images, thus must have a Docker service
+services:
+  - docker
+
+# Build Docker image for development and testing
+before_script:
+  - make container
+
+# Each env becomes a separate job. Add more Python versions to have a matrix.
+env:
+  - MAKE_STEP=test-static
+  - MAKE_STEP=build
+  - MAKE_STEP=test-runtime-base
+  - MAKE_STEP=test-runtime-root
+  - MAKE_STEP=test-runtime-slow
+
+matrix:
+  allow_failures:
+    - env: MAKE_STEP=test-static
+
+script: make $MAKE_STEP
+
+# Travis-CI cannot cache the plain testfiles due to permission errors, thus we
+# have the .tar.gz stored in the cache dir to aboid too many re-downloads.
+cache:
+  directories:
+    - $HOME/build/rdiff-backup/cache
+# Runs in path /home/travis/build/rdiff-backup/rdiff-backup (git repo work directory)
+# One level up is an empty directory that can be used for artifacts, cache etc.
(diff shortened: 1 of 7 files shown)


ottok:
See the CI result at https://travis-ci.org/rdiff-backup/rdiff-backup/builds/579527732 or by clicking at the CI links above embedded in the commits and PR status box.

ikus060:
What is the value of the Makefile ? 
1. It's miss leading cause usually makefile are called with `./configure`, `make`, `make install`. Having a makefile is the repository is miss leading the users that are used to this patterns.
2. It's yet another file to manage, while Travis yml provide all the feature you need to call the proper command line.

It should not be required to build our own rdiff-backup image. `debian:stretch` should be the right image. And the local source code should be mounted into the image using `-v`.
So here again. No need for Dockerfile.

I would rename `setup-testfiles.sh` to `build.sh` or something similar.

As mentioned in an earlier comments, we should prever the use of manylinux here, the image provided contains multiple version of python, so it ease the testing for different python versions and the binaries generate might be pushed to pypi.

ottok:
True, it is not using `configure` and `install` as Makefiles usually do. I could not add these as the Python program does not really have steps like this. I could add some mockup of them though.

The point of the Makefile is that I (or anybody) can easily run on their own computer the exact steps that Travis-CI runs. Instead of manually copy-pasting lines from `.travis-ci.yml` one can just run `make X`. With this structure there is just one place to update the commands, and that is the Makefile. With any other version the same Docker commands might need to be in two places, one for Travis-CI and one for local use.

The `testfiles` are only for development, it would not be appropriate to call them `build.sh` since the actual build does not need those files when making the Python package for distribution.

Debian Sid is used for development and it has Python 3.7. Debian Stretch/Buster is a released version for stable use cases, but here we target development and the latest of latest, now stability.

Local source dir IS mounted using `-v`.

I am not familiar with manylinux. I agree that the next logical step is to extend the number of platforms and Python versions we test for as said in the commit message. This is just the first version so we actually get CI in the workflow.

ikus060:
Let me rephrase this. 

If we want to make rdiff-backup more robust, we clearly need a CICD pipeline as soon as possible. So your implementation to address this item is great.

But if we also want to make rdiff-backup more accessible and easy to contribute we should also embrace the python conventions that already exists. In that regards we need to improve the situation.
As a first start, python conventions make uses of `setup.py` to manage many of the functionnalities that you have placed in `Makefile` and `setup-testfiles.sh`. As an example, to run the test, it should be possible to run them using `python setup.py test`. The `test` command may prepare the environment as needed. The final way to call this for docker might look like this: `docker run -v `pwd`:/build python:3 python /build/setup.py test` or something similar. A Makefile or an additional shell script should not be required.

I also recommend to call tox in a similar way like this: `python setup.py tox`.

We should probably tackle all of this later and track this with an issue if we want to more forward.

ottok:
I agree that the setup.py should be used more, but since our current testfiles story is what it is (see #54)  we also need some extra shell script snippets.

PR #118, opened 2019-09-05
Fix regression of local tests on ACLs and uncomment rpath char dev test

### LINKED ISSUE #117: Regression: ACL tests failing on iteration and mapping
We have a regression somewhere: if I checkout my last commit and run the tests, they are successful, if I do the same on master, the tests do fail:

```
$ git status
HEAD detached at cd6eac9
nothing to commit, working tree clean
$ git checkout master
Previous HEAD position was cd6eac9 Make the test scripts PEP8/flake8 compatible (2)
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ tox -e py37
[...]
FAIL: testIterate (__main__.ACLTest)
Test writing several records and then reading them back
  File "testing/eas_aclstest.py", line 483, in testIterate
    assert acl1_reread == self.acl1
AssertionError

FAIL: test_acl_mapping (__main__.ACLTest)
Test mapping ACL names
  File "testing/eas_aclstest.py", line 600, in test_acl_mapping
    assert get_perms_of_user(out_acl, self.current_user) == 7
AssertionError

Ran 20 tests in 8.479s

FAILED (failures=2)
ERROR: InvocationError for command '/home/ericl/Comp/rdiff-backup/.tox/py37/bin/python testing/eas_aclstest.py' (exited with code 1)
__________________________________________________________________________________________________________________ summary ___________________________________________________________________________________________________________________
ERROR:   py37: commands failed
```

I won't have time quickly to analyse the issue so if someone could look into it before we change too much the code again.

ericzolf:
This is basically a revert of 5954f6059ff49bcb426daa0eed73d81665c99bad

- User and Group for ACL tests are now gathered optionally from environment
  variables RDIFF_TEST_USER and RDIFF_TEST_GROUP.
- Rpath tests implying existence of /dev/tty2 are skipped if it doesn't exist

Closes #117

### DIFF
--- a/testing/eas_aclstest.py
+++ b/testing/eas_aclstest.py
@@ -272,9 +272,8 @@ user.empty
 class ACLTest(unittest.TestCase):
     """Test access control lists"""
 
-    # @TODO: Use the instrumented testuser here in case tests are run as root
-    current_user = 'testuser' # pwd.getpwuid(os.getuid()).pw_name
-    current_group = 'testuser' # grp.getgrgid(os.getgid()).gr_name
+    current_user = os.getenv('RDIFF_TEST_USER', pwd.getpwuid(os.getuid()).pw_name)
+    current_group = os.getenv('RDIFF_TEST_GROUP', grp.getgrgid(os.getgid()).gr_name)
 
     sample_acl = AccessControlLists((), """user::rwx
 user:root:rwx
--- a/testing/rpathtest.py
+++ b/testing/rpathtest.py
@@ -56,11 +56,11 @@ class CheckTypes(RPathTest):
         assert rpath.RPath(self.lc, self.prefix, ("fifo", )).isfifo()
         assert not rpath.RPath(self.lc, self.prefix, ()).isfifo()
 
-    # @TODO: One cannot really assume tty2 exists everywhere
-    #def testCharDev(self):
-    #    """Char special files identified"""
-    #    assert rpath.RPath(self.lc, "/dev/tty2", ()).ischardev()
-    #    assert not rpath.RPath(self.lc, self.prefix, ("regular_file", )).ischardev()
+    @unittest.skipUnless(os.path.exists('/dev/tty2'), "Test requires /dev/tty2")
+    def testCharDev(self):
+        """Char special files identified"""
+        assert rpath.RPath(self.lc, "/dev/tty2", ()).ischardev()
+        assert not rpath.RPath(self.lc, self.prefix, ("regular_file", )).ischardev()
 
     @unittest.skipUnless(
         os.path.exists('/dev/sda') or os.path.exists('/dev/nvme0n1'),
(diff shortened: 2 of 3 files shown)


ottok:
You can't simply revert this, as test will start failing for Travis-CI and (all/most) contributors locally. What we need is two separate commits/fixes for both of those issues.

ottok:
You can test this locally with `make test`. Currently Travis-CI fails with the same error message:
```
FAIL: test_acl_mapping (__main__.ACLTest)
Test mapping ACL names
  File "testing/eas_aclstest.py", line 599, in test_acl_mapping
    assert get_perms_of_user(out_acl, self.current_user) == 7
AssertionError
```

The value `7` is what root has, so it is not picking up the user name/group correctly now. Code looks OK to me though.

ericzolf:
As which user do the tests actually run in the Docker container? Can it be that they run as root? I don't see any change of user in the Makefile or in the travis file...

ericzolf:
Yep, see https://docs.docker.com/engine/reference/builder/#user
You need to add `USER testuser:testuser` to Dockerfile, make sure testuser can do sudo without password, and adapt the Makefile accordingly to use sudo for `test-runtime-files` and `test-runtime-root`.

I can't test it locally, you need to do the changes and I'll rebase this PR upon master once your PR is merged.

ottok:
I tested printing out `os.getenv('RDIFF_TEST_USER')`  in this test, and it seems empty. Apparently the code does some user changing stuff without preserving the ENV.

I was however unable to track down where the user is changed and what happens there.

ericzolf:
Ah, yes, we must tell tox to pass necessary variables through. tox.ini now fixed through latest push.

My comments around proper setup of the container still holds though. Running the slow and base tests as root isn't meant to work.

ericzolf:
mmh, no more approval required? Is this wanted or an indirect consequence of the Travis checks?

ottok:
Why do we have then RDIFF_TEST_USER and RDIFF_TEST_GROUP if running the testsuite as a user is a requirement anyway and don't want to support running the testsuite as root? Should they be removed then?

ericzolf:
I just added them to make the tests run in a container, I didn't need them before. They're only a work around in case the automated discovery of the current user doesn't work.

And I think it's best practice to not develop or test as root unless absolutely required.

ottok:
> I just added them to make the tests run in a container, I didn't need them before.

You could have just told me that *not* running as root is a requirement to begin with.. The value of the Dockerfile is exactly to show what are the concrete commands and in what context to run them so that from now on everybody will know how to exactly repeat the test runs.

ericzolf:
I didn't realise that the tests weren't run as non-root, given the comments in the makefile, correctly marking what needs to run as root and what not. 

Anyway, it doesn't hurt, the variables are unnecessary but that's all. The Dockerfile and the Makefile should just be fixed to align with the comments.

PR #123, opened 2019-09-13
Revert and improve versioning procedure

ericzolf:
Revert "Take package and distribution tool from Git"
    
This reverts commit c46e7fb2b498b430a9b0e27fdb8b0e5cb6a1a4ea.
    
It appears that the approach leads to Git structures being owned by
root when calling `sudo ./setup.py install` even when calling just
before `./setup.py build` as non-root, and this is not acceptable.

Instead implement a simpler version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,6 @@
 .*.swp
 build/
 __pycache__
-src/rdiff_backup.egg-info/
-src/rdiff_backup/Version.py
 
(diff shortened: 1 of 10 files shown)


ottok:
I suggest you re-use some pattern already used by another project to solve #66 in the most reliable way.

ericzolf:
This is exactly what the 2nd commit does.

Issue #115, opened 2019-09-01
Remote backup fails with truncated pickle data


frankcrawford:
Testing rdiff-backup from latest git version with remote backup fails with an error of:

Sending back exception pickle data was truncated of type <class '_pickle.UnpicklingError'>

See traceback attached.
[backup.log](https://github.com/rdiff-backup/rdiff-backup/files/3563105/backup.log)

ericzolf:
Sorry for the problem but can you please: re-run the command with verbosity 9 and send the command used as well as the output of `git describe --long --all`? There isn't enough information to understand the problem. We might have even to ask you to provide the file causing the issue, but let's see first what the more verbose output is saying.

frankcrawford:
As always in computing, both requests are harder than it sounds, in that this is being run under a wrapper program called safekeep, and rdiff-backup is actually installed from an RPM built from a download from git master.
The RPM is rdiff-backup-1.3.4-0.20190829git56cfd72.fc30.x86_64, which was from master 56cfd72.
The command as listed in the attached typescript was (and seems to not have listed the appropriate quotes):
[...]

PR #124, opened 2019-09-18
Make debug output in verbosity 9 easier to sort in client/server mode

ericzolf:
While working on issue #115 it appears that debug messages are unsorted
when working in client/server mode, and hence need a sortable date and
clearer affiliation to one or the other side of the communication.
The development doc has been extended to show how to use the new format.


ericzolf:
Made obsolete by closing #125

Issue #109, opened 2019-08-25
One or more test in tox.ini is unstable especially under Python 3.5


ericzolf:
I haven't yet identified which one exactly but it happens from time to time that I run `tox` and one of the tests fail, generally py35 (but it might be due to the fact that it's the first environment). When I run it again, the issue has disappeared. Because of the length of the output (and because it disappears on re-run) I've never been able to identify the actual problem.

This issue is mostly there to make everybody aware of the issue, please document it here if you find out what the problem is (or at least in which test it appears).

ericzolf:
Need to verify in due time but it might be the killtest, which would make sense as it is time sensitive (it needs to kill the jobs while they run).

ericzolf:
It seems that it's more complicated than expected, as the test seems to fail when rdiff-backup
is killed on time but still exits with a return code unequal 0. It can be an issue of rdiff-backup which doesn't get out properly (meaning here with an error), or just a stupid timing issue like if rdiff-backup is finished but python not, or something similar. No idea for now on how to properly analyze this, especially because it's so difficult to reproduce.

The culprit is most probably the assert statement in the `exec_and_kill` function.

ikus060:
Do you understand why it's failing under 3.5 ?
I'm working on travis build and I've remove py3.5 because of this test.

--
Patrik Dufresne Service Logiciel inc.
http://www.patrikdufresne.com <http://patrikdufresne.com/>/
514-971-6442
130 rue Doris
St-Colomban, QC J5K 1T9

ericzolf:
I can only guess that it's because the timing is slightly different under different versions of Python. It happens more often under Python 3.5 but it happens also under Python 3.7.

PR #125, opened 2019-09-19
fix 115 truncated pickle

ericzolf:
This MR is based on #124 and does only remove the bufsize=0 parameter from Popen.

By removing bufsize=0 to Popen, we get a buffered pipe avoiding short reads which seems to have been the issue encountered.

I'll remove the WIP once the fix has been confirmed by the reporter @frankcrawford


ericzolf:
The failure on one of the Travis CI builds looks like #109 - I'll force a re-build once @frankcrawford has given feedback.

ericzolf:
PR isn't WIP anymore after confirmation that the fix solved the issue. Someone can merge it once the tests are successful. Thanks.

frankcrawford:
Folks,
Can someone do this merge, as it is a significant bug with a simple fix, although it may now need to be refreshed due to formatting changes.

Frank

ericzolf:
I don't like to merge my own branches but here you are.

frankcrawford:
Thnks Eric.  I can understand, but it was dragging on for so long, I felt some need to comment.

PR #131, opened 2019-10-01
Describe how to create a Windows development environment

ericzolf:
I'm a bit stuck, I get a rsync.dll instead of rsync.lib as expected. Has someone a clue what could be the correct approach? @ikus060 or anybody else...

`"cmd": "python.exe setup.py build --librsync-dir=\"C:/Users/IEUser/Develop/librsync.d\" --lflags=\"/NODEFAULTLIB:libcmt.lib msvcrt.lib\""` fails with `"LINK : fatal error LNK1181: cannot open input file 'rsync.lib'"` and indeed the file doesn't exist, only `rsync.dll`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ rdiff-backup homepage at https://rdiff-backup.net/ should
 have a recent version of librsync; otherwise see the librsync homepage
 at http://librsync.sourceforge.net/. On Windows, you must have the
 Python for Windows extensions installed if you are building from source.
-The extensions can be downloaded from: http://pywin32.sourceforge.net/
+The extensions can be downloaded from: https://github.com/mhammond/pywin32
 If you are not building from source on Windows, you do not need Python
 or librsync; they are bundled with the executable.
 
--- a/tools/windows/.gitignore
+++ b/tools/windows/.gitignore
@@ -1,0 +1,1 @@
+.vagrant/
--- a/tools/windows/README.md
+++ b/tools/windows/README.md
@@ -18,6 +18,8 @@ ansible-playbook -i .vagrant/provisioners/ansible/inventory/vagrant_ansible_inve
 ansible-playbook -i .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory playbook-build-rdiff-backup.yml
 ```
 
+> **NOTE:** you can use the variables under `group_vars/all` to steer the build process, check the comments there for more details.
+
 ## Develop and try under Windows
 
 Open a console and type:
--- a/tools/windows/README.md
+++ b/tools/windows/README.md
@@ -1,0 +1,66 @@
+# Windows development environment
+
+## Create the Windows VM
+
+Starting from https://github.com/redhat-cop/automate-windows/tree/master/vagrant-libvirt-image, create a Windows VM
+usable by Ansible (any other alternative approach to a such VM is of course valid).
+
+Then apply the playbook playbook-provision-windows.yml using Ansible to the Windows you've just created to create the necessary development environment (a `vagrant up` resp. `vagrant provision` in the current directory might be sufficient).
+
+> **IMPORTANT:** The current state of the automation isn't very satisfying, the more complex packages need to be installed manually from the command line with something like `choco install <packagename>` and the playbook restarted. A few reboots in-between might be necessary.
+
+## Build librsync and rdiff-backup
+
+It can be as easy as calling twice ansible-playbook:
+
+```
+ansible-playbook -i .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory playbook-build-librsync.yml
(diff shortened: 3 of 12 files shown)


ericzolf:
A hint perhaps: https://stackoverflow.com/questions/7614286/how-do-i-get-cmake-to-create-a-dll-and-its-matching-lib-file and https://cmake.org/cmake/help/latest/command/add_library.html - no clue yet how to apply though...

ericzolf:
Perhaps even more helpful: https://stackoverflow.com/questions/33062728/cmake-link-shared-library-on-windows

ikus060:
Ho ok, I see, you are creating a VM in vagrant and then try to compile.
1. Do you plan to use this as a build system with travis ? To my taste, I find appveyor a more simple solution.
2. Your issue is in librsync.vcproj, but I can't find where yours is comming from.

Take a look here. This is the pipeline I'm using to compile rdiff-backup 1.2.8 on linux with Wine.
https://git.patrikdufresne.com/pdsl/rdiff-backup-build/

Also take a look at the librsync.vcproj line 64
https://git.patrikdufresne.com/pdsl/rdiff-backup-build/blob/master/patches/librsync/librsync.vcproj#L64

ericzolf:
To answer your questions:
1. no, that's not the plan, I just need a way to compile and test interactively under Windows
2. my vcproj comes from the call to cmake - from what I read and what you tell me, it seems that I haven't yet found the magical parameters to generate the correct vcproj for msbuild to do the correct thing. But I return the question, where did you get yours?

And looking at your CICD Pipeline: how did you install MSBuild under Wine?

Side note: it's interesting that py2exe can build Windows executables even under Linux.

ikus060:
> my vcproj comes from the call to cmake
It doesn't make sens for me. I'm no expert of windows build, but usually, the `.vcproj` and `.sln` are created manually. Then you are making use of them to compile. Can you point wich line is generating your vcproj ?
I got mine from a very old post on the mailing list that was describing how to compile rdiff-backup and librsync. 

https://lists.gnu.org/archive/html/rdiff-backup-users/2009-02/msg00110.html

https://git.patrikdufresne.com/pdsl/rdiff-backup-build/blob/8b43af60d13a13489cd1c702f1b762cbb7a193fc/build.py

> And looking at your CICD Pipeline: how did you install MSBuild under Wine?

With `winetricks -q vc2008express`
But that only work for Visual Studio 2008 and python 3.7 need a more recent version of visual studio. I did not manage to make recent version of visual studio working on windows...

Hope it help you

ericzolf:
The command is `"{{ cmake_exe }}" -G "Visual Studio 15 2017" -A x64 -D BUILD_RDIFF=OFF .` called in the librsync source directory.

ericzolf:
Trying to get some help from the librsync project: https://github.com/librsync/librsync/issues/171

ardovm:
I replied on your issue on the librsync project.
Please make sure you are using the _latest_ code of librsync, as they have done some important modifications in the last days, involving static vs dynamic lib building, and compatibility with Visual Studio.

Please check the file `dist_win32.py` from my branch of rdiff-backup; I hope it can be useful for you.

ardovm:
> > And looking at your CICD Pipeline: how did you install MSBuild under Wine?
> 
> With `winetricks -q vc2008express`
> But that only work for Visual Studio 2008 and python 3.7 need a more recent version of visual studio. I did not manage to make recent version of visual studio working on windows...

For what I have seen, the worst problem with Python 3.7 is that it is not compatible any more with py2exe. I am trying to use PyInstaller instead, from file `dist_win32.py` in my rdiff-backup fork.

ericzolf:
Just for me to find it more easily (thanks for all the hints, just need to put the bits & pieces together):

[similar discussion about packaging python as exe](https://github.com/nvaccess/nvda/issues/8375)

direct link to the fork: https://github.com/ardovm/rdiff-backup/blob/master/dist_win32.py

ericzolf:
Thanks for all the help, the current branch state works up to the point where everything is building fine, and `rdiff-backup --version` succeeds. `rdiff-backup source target` fails with "TypeError: Objects of type 'bytes' can not be converted to Unicode.", which doesn't sound like a compilation problem but more like a programmer's one :stuck_out_tongue_winking_eye: 

There are still some rough angles to fix, like documentation and using latest librsync from Git (I'm still using for now 2.0.2 from tarball), but it looks like we're on the right track. Thanks again!

ericzolf:
The playbooks are not perfect but they're good enough for my purpose (debug rdiff-backup under Windows), and hence I don't spend more time on this PR and re-focus on the actual purpose, for which I'd need a merge (@ottok ).

@ikus060 even if you don't use Ansible, I think it would be good if you'd use a similar approach to mine in your appveyor branch, so that we get comparable results. But we can discuss this more in details on your branch.

PR #132, opened 2019-10-05
Move all dist files from dist to tools and ignore dist/

ericzolf:
The dist directory is generally for generated distribution files and is
being used as such by setup.py, similar to the build directory, hence it
belongs outside of the git repo and ignored.

It's getting in my way while compiling under Windows, so a quick merge would be appreciated @ottok  so that I can rebase on master.

@ikus060 please consider the change in your appveyor fork as well (all utilities under tools instead of dist).


ikus060:
Nice move !

frankcrawford:
A minor comment of this PR, but all the scripts in this directory have not been fully upgraded to python3.  I was about to submit a PR to fix them as well as make makedist work correctly, but will leave it until this one is merged into the current master.

ericzolf:
Merged so that everybody can continue to work. Happy hacking!

PR #134, opened 2019-10-06
Updated to correct syntax for python3.

frankcrawford:
This is just a simple clean up of the files, pretty much from a run of 2to3.

### DIFF
--- a/tools/makedist
+++ b/tools/makedist
@@ -129,8 +129,8 @@ def MakeTar(specfiles):
 				  % (tardir,), 1)
 	VersionedCopy(DistDir + "/setup.py", "%s/setup.py" % (tardir,))
 
-	os.chmod(os.path.join(tardir, "setup.py"), 0755)
-	os.chmod(os.path.join(tardir, "rdiff-backup"), 0644)
+	os.chmod(os.path.join(tardir, "setup.py"), 0o755)
+	os.chmod(os.path.join(tardir, "rdiff-backup"), 0o644)
 	CopyMan(os.path.join(tardir, "rdiff-backup.1"), Version)
 	CopyMan(os.path.join(tardir, "rdiff-backup-statistics.1"), Version)
 	if os.name != 'nt':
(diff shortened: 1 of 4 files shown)


frankcrawford:
Not sure where the travis-ci error comes from.  It doesn't really seem to be related to my change.

ericzolf:
Nope, it isn't your change, Debian sid is broken and the docker build fails on installing devscripts. You can call `git commit --amend` and  `git push --force` to re-trigger the build until sid is fixed.

@ottok how can we avoid this in the future? Switch to testing or stable?

ottok:
We need to specifically test against sid, otherwise we won't detect relevant changes in new versions of libraries etc we use until it is too late.

It seems this devscripts step is globally broken right now, has been reported elsewhere as well. I assume it will be fixed soon..

frankcrawford:
So, is there anything I can do at present to get this to pass, or just wait for it to be fixed elsewhere?

ericzolf:
@ottok I have the impression, having been myself a Debian packager, that you are thinking too much like one: it is true that it is recommended to package against Sid, but I have doubts if it is the right strategy to "punish" PR authors because Sid is broken (and it can be broken at any time).

In my opinion, we should check PRs against a stable / repeatable base, and have the ability on demand to check master against unstable before we create a release. Or have the build done against Sid for each master but then have the check optional, and only the one against stable/testing mandatory.

ottok:
The Travis-CI failure is due to perl transition in Debian that started yesterday: https://lists.debian.org/debian-devel-announce/2019/10/msg00000.html

We could perhaps use a monthly updating Docker image of Sid that we roll ourselves instead of real live Sid if issues like this are frequent, but in my experience this is very rare.

Let's hope it is over in a couple of days.

ericzolf:
Well, there are a few transitions with each release of Debian, not that it happens too often and that they must all impact us, but we're stuck for now. That's not really satisfactory.

And how do we know that it's corrected?

ericzolf:
Merging, even though I'm not sure we want to keep those long term (all distribution activities should IMHO happen through setup.py, but other discussion).

Issue #64, opened 2019-08-14
Decide and implement a generic documentation handling method


ericzolf:
The problem is bigger than initially described (see below the line):  we need to streamline the distribution process, also when it comes to documentation: we have currently a mixture of txt, md and html, and we should all align and handle similarly. 

----

During installation:

```
sudo python3 dist/setup.py install --prefix=/usr/local
[...]
running install_data
error: can't copy 'FAQ.html': doesn't exist or not a regular file
```
FAQ-body.html does exist but not FAQ.html, which is actually created in `dist/makedist` with the function `MakeHTML("FAQ", "rdiff-backup FAQ")`.

ericzolf:
My suggestion: transform all documentation in markdown and generate HTML & co using pandoc or similar during the distribution process.

We can also switch to asciidoctor and even generate PDF and more easily :grin:

frankcrawford:
While the idea of transforming all the files into markdown, as a simpler issue, both FAQ-body.html and example-body.html can both be permanently transformed into FAQ.html and example.html.

All that tools/makedist does in this particular case, is to add a header and footer on, for two files, one html and the other wml.

It looks like wml was used in one of the previous websites, but we no longer do, so we don't need the wml file anymore.

I'd suggest at present we just convert these two to html and drop all reference to wml, and then as a longer process covert all the documents to a generic document format.

What do you think about this as a first step?

PR #135, opened 2019-10-06
Transform DEVELOP.adoc into DEVELOP.md

ericzolf:
Doesn't make sense to have two different documentation formats and markdown
seems to be the consense, so here it is.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -1,33 +1,33 @@
-= Development Guide for rdiff-backup
+# Development Guide for rdiff-backup
 
 Some notes for developers and other people willing to help development.
 
-== Testing rdiff-backup
+## Testing rdiff-backup
 
-- Install:
-* tox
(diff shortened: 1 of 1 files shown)


frankcrawford:
Eric,
Similar to the FAQ, do you want this automatically turned into HTML during setup.py, or just left as is for inclusion in the distribution?

Frank

ericzolf:
I'm not sure it belongs in the distribution... And the FAQ generation is currently broken #64 , I had to do an ugly fix. My personal take is that we need to streamline the distribution process, also when it comes to documentation: we have currently a mixture of txt, md and html, and we should all align and handle similarly.

We could transform all in markdown and generate HTML & co using pandoc or similar.

ericzolf:
I've extended accordingly the issue #64 so that we can address the bigger issue at once. Join the discussion over there to give your opinion.

PR #23, opened 2017-10-28
Fix check for proper parent directory inclusion in include/exclude options

ilario:
Fixes #22

PR #137, opened 2019-10-12
Improve check for complete path in include/exclude

### LINKED ISSUE #22: No error raised in certain occasions when excluding files the wrong way
This bug in the input options check was originally reported on the mailing list here:
https://lists.nongnu.org/archive/html/rdiff-backup-users/2017-01/msg00005.html

In this example the software correctly throws an error, because the excluded file is not specified correctly:
```
$ mkdir temp
$ touch temp/file.txt
$ touch temp/picture.jpg
$ rdiff-backup --exclude file.txt temp backup
Fatal Error: Fatal Error: The file specification
    'file.txt'
cannot match any files in the base directory
    'temp'
Useful file specifications begin with the base directory or some
pattern (such as '**') which matches the base directory.
```

But in certain cases this check fails and no error is printed even if the exclusion is not valid, like in this:
```
$ mkdir temp
$ cd temp
$ mkdir .one
$ mkdir .two
$ mkdir three
$ mkdir four
$ rdiff-backup --exclude .one --exclude ./.two --exclude ./three . ../backup
$ ls -a ../backup
.  ..  four  .one  rdiff-backup-data
```

or in this:
```
$ mkdir temp
$ touch temp/file.txt
$ touch temp/temp.txt
$ rdiff-backup --exclude temp.txt temp backup
$ ls backup
file.txt  rdiff-backup-data  temp.txt
```

ilario:
This is a rewrite of #23 and will fix #22 at least for the cases I tested (for the examples, check out #22).
Maybe the same modification should be added also elsewhere inside selection.py?

### DIFF
--- a/src/rdiff_backup/selection.py
+++ b/src/rdiff_backup/selection.py
@@ -630,7 +630,7 @@ probably isn't what you meant.""" % (self.selection_functions[-1].name, ))
 		globbing characters are used.
 
 		"""
-        if not filename.startswith(self.prefix):
+        if not filename.startswith(self.prefix + b"/"):
             raise FilePrefixError(filename)
         index = tuple(
             [x for x in filename[len(self.prefix):].split(b"/") if x])


ericzolf:
Raise the hand if you need help with analyzing the pipeline results or are the opinion that the tests should be adapted.

ilario:
Yes, I suppose that something like the examples I included in #22 should be included in the tests.

I tried to run the tests following the instructions in the DEVELOP.md file.

The fact that the tests have to be unpacked in the parent directory should be underlined, at first I didn't get it.

My user has ID 1000, I am on Arch Linux and I think I have all the required dependencies (and Python 3.7).
I unpacked the tests using sudo, as recommended in the instructions.
Then inside the rdiff-backup I ran tox as user and got an error, which in the log is:

```
cat .tox/log/GLOB-0.log 
action: GLOB, msg: packaging
cwd: /home/ilario/software/rdiff-backup
cmd: /usr/bin/python setup.py sdist --formats=zip --dist-dir /home/ilario/software/rdiff-backup/.tox/dist
running sdist
running egg_info
error: [Errno 13] Permission denied
```

running the same python command manually and adding a "-v" option, I can see some more output but still I have no idea of what is happening:

```
/usr/bin/python -v setup.py sdist --formats=zip --dist-dir /home/ilario/software/rdiff-backup/.tox/dist
[...]
# /usr/lib/python3.7/site-packages/setuptools/__pycache__/glob.cpython-37.pyc matches /usr/lib/python3.7/site-packages/setuptools/glob.py
# code object from '/usr/lib/python3.7/site-packages/setuptools/__pycache__/glob.cpython-37.pyc'
import 'setuptools.glob' # <_frozen_importlib_external.SourceFileLoader object at 0x7f159afa12d0>
import 'setuptools.command.egg_info' # <_frozen_importlib_external.SourceFileLoader object at 0x7f159b0ecdd0>
error: [Errno 13] Permission denied
[...]
```

When I run tox as root, it fails trying to get the ID of the sync user, which on my system does not exist (should it?).

```
py37 run-test: commands[4] | python testing/user_grouptest.py
..E.E
ERROR: test_default_mapping (__main__.UserGroupTest)
Test the default user mapping
  File "testing/user_grouptest.py", line 38, in test_default_mapping
    syncid = pwd.getpwnam('sync')[2]
KeyError: "getpwnam(): name not found: 'sync'"

ERROR: test_user_mapping (__main__.UserGroupTest)
Test the user mapping file through the DefinedMap class
  File "testing/user_grouptest.py", line 56, in test_user_mapping
    syncid = pwd.getpwnam('sync')[2]
KeyError: "getpwnam(): name not found: 'sync'"

Ran 5 tests in 0.005s

FAILED (errors=2)
ERROR: InvocationError for command /home/ilario/software/rdiff-backup/.tox/py37/bin/python testing/user_grouptest.py (exited with code 1)
```

ericzolf:
Sorry for the delay. OK, that's quite a lot of issues there:

1. the Permission denied error is unknown to me. I would suspect that you've run `tox` or `setup.py` as root before and you can't update `.tox` or `dist` or `build` as normal user anymore. Remove those directories and try again. You might want to use `git clean` to make sure you get a clean git repo overall.
2. the `sync` error is most probably due to the user `sync` not existing on your Linux. It's a bit surprising as it seems to exist by default since years, but not impossible. It's not difficult to change the user in the tests but it's a bit more difficult to find one existing on all Linux flavours. Do you mind creating a sync user on your box?
3. what I initially meant and I'm not sure I was clear, is that the continuous-integration pipeline choke on your changes and you need to fix this (adding tests won't help even if it's a very good idea). You can see the issue by clicking on the "Details" button and then on job 162.2; roughly at the end you'll see that the fileselection test has been broken.

In parallel let me look a bit into your PR, I might have a few more hints.

> side note: I checked the [LSB chapter 23.2](https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic.pdf) and sync is only optional, but only root is mandatory and not obsolete, sooo...

ericzolf:
The issue with your fix is that it doesn't work if the last character of prefix is `/` because it then ends with two slashes at the end, which doesn't match a normal absolute path with only one slash.

So you need to differentiate between having a slash at the end of prefix or not and that should make you pass the tests.

src/rdiff_backup/selection.py
```diff
@@ -630,7 +630,7 @@ def glob_get_filename_sf(self, filename, include):
 		globbing characters are used.
 
 		"""
-        if not filename.startswith(self.prefix):
+        if not filename.startswith(self.prefix + b"/"):
```
ericzolf: if you do something like the following, it should work:

```python
self.prefixindex = self.rpath.index  # inserted after line 88 to avoid splitting ourselves

[...]

fileindex = tuple([x for x in filename.split(b"/") if x])

# are the first elements of the path the same?
if fileindex[:len(prefixindex)] != self.prefixindex:
            raise FilePrefixError(filename)
index = fileindex[len(self.prefixindex):]
```

Actually, you should pack these lines in a method that returns the index based on filename, because you can reuse it at lines 438 to 441 of the same file.

ericzolf:
@ilario are you fine with my comments/suggestions to fix the issue? Do you have time and know how to proceed from here?

ericzolf:
@ilario any chance you finalize this PR (which would be useful and just needs to be corrected as I commented)?

ilario:
Sorry for the lag, I hope I'll manage to keep up with this in max 2 weeks.

ericzolf:
All good, I quickly tried a bit more my proposal but it also fails the tests. No clue why...

ericzolf:
`self.prefixindex = self.rpath.index` was a wrong good idea, you need to split the path yourself. And `fileindex[:len(prefixindex)]` should read `fileindex[:len(self.prefixindex)]`.

PR #139, opened 2019-10-16
Add the --no-tar option to tools/makedist

ardovm:
The script `tools/makedist` can only be used to make a compressed tar archive.
It generates a temporary directory, makes the archive and then wipes it.
It could be useful to avoid the last part, i.e. the script could be used to make a ``plain'' directory with all the source files.


ardovm:
Woha! I must have done something very wrong to have all these commits listed here. I am sorry.
But in fact only one file is changed.
I hope this is not a problem.

ericzolf:
Hi, sorry but I won't merge this PR. Create a new one which looks normal in terms of commits carried... I could imagine that you have librsync as submodule in your git repo.

This said, I'm not sure makedist will live long and prosper ever after: IMHO setup.py should be our sole distribution tool, and makexxxx disappear. But if you don't mind, we can review and merge a proper branch.

frankcrawford:
While I expect makedist will undergo some changes in the near future, I think it may stay around, because it is a handy way create a tar distribution, update the version number, etc.

frankcrawford:
Also, I'm planning to submit a PR shortly to dump the --no-html option from makedist, as it won't be needed going forward.

ardovm:
Sorry for the noise. I will issue another PR, but this time following the proper best practices.

PR #140, opened 2019-10-17
Clean up some documentation files

frankcrawford:
This commit makes FAQ.html and examples.html permanent files, and not required to be build by tools/makedist.

It is no longer necessary to build them as we no longer wml files as these only related to the old website and are not used in any of the current distributions or websites.

This also updates the distribution build files (setup.py and RPM spec files) to include all the correct documentation.


ericzolf:
Looks good to me. Just one thing @frankcrawford : would you mind taking the chance to make the HTML code more modern, like HTML5? Just if you agree and find a tool to make it automatically, else I'll be happy to merge as it is.

frankcrawford:
@ericzolf I'll see what I can do about cleaning up the HTML.

frankcrawford:
@ericzolf Okay, I've found an easy way to do this, I run the current HTML through pandoc to turn it into markdown, and then back through pandoc to make it HTML5.

Now the question is, do you want to keep the markdown?  Given you want to move all the documentation to markdown, I think it would be a good idea, but where should it go?

ericzolf:
Nice trick! As they would be the source for our documentation, I would suggest `src/docs`.

frankcrawford:
While I've turned the documentation into markdown, I've currently left the html files in until we have worked out how to automatically generate the html and other documentation from the markdown files.

ericzolf:
That's the way I understood it as will. Fine with me. Merging. Thanks for the effort.

Issue #106, opened 2019-08-25
Recent change to support bytes and python3 is breaking Windows build


ikus060:
Related to #105, the recent pull request to support python3 and make uses of bytes completely broke the Windows build.
All the function os.* on windows platform are only supporting str, not bytes.
All the function from pywin32 are only supporting str, not bytes.

@ericzolf As this point, I'm not sure what to do with the Path. Bytes or Str make it a problem. On Linux Path are clearly manage as bytes, on Windows.

If I was managing this project, I would revert back the py2to3 pull request and work on it again and make sure to make the unittest working on all the platform: Linux, Windows And OSx. Basically, the recent changes broke the build.

ottok:
Can you please test on Windows if the Python 3 built-in Path object works as expected? See https://github.com/rdiff-backup/rdiff-backup/pull/53#issuecomment-522267433

I didn't understand the problem fully and I have no test case to verify anything, so this using Pathlib was just a general suggestion.

I do understand your viewpoint on reverting the merge, but as stated in the last paragraph of https://github.com/rdiff-backup/rdiff-backup/pull/53#issuecomment-523794354 Eric finds it easier to have this on master and then clean up and fix from there, so we will go with that as Eric wrote the Python 3 migration. We could revert though if Eric finds that fixing does not work out.

In the mean time, getting a CI in place is a big priority and we are currently working on both Travis-CI (me) and Appveyor (you). Related issues and PRs (just for reminder): #48, #105, #94, #96.

ikus060:
> Can you please test on Windows if the Python 3 built-in Path object works as expected? See #53 (comment)
> I didn't understand the problem fully and I have no test case to verify anything, so this using Pathlib was just a general suggestion.

I still trying to figure out the extend of this problem my self. Long story short, is bytes are not working well on Windows.

On the same system, different version of python will return different bytes representation of the same files ! The point of using bytes, at least on Linux, was to take the unaltered version of the paths regardless of the encoding defined by the user (LANG environment variable). But on Windows it seams more difficult. NTFS is UTF-16, but python is reporting something different depending of the version used.

**Python 3.7**
```
Python 3.7.4 (tags/v3.7.4:e09359112e, Jul  8 2019, 19:29:22) [MSC v.1916 32 bit
(Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os, sys
>>> sys.getfilesystemencoding()
'utf-8'
[...]

PR #141, opened 2019-10-19
Fix most apparent Windows issues

ericzolf:
- use fsdecode on bytes paths when using win32 functions
- ignore rdiff-backup.spec created by Windows build
- use bytes instead of strings for win_acls
- replace os.path.join and os.getcwdb with own RORPath methods to
  make sure slashes are used under Windows.
- try to make life of a Windows developer on Linux easier:
    - add 7zip to Windows build to be able to compress/uncompress
    - try to add broken cygwin chocolatey package and comment it
    - create helper script for debugging rdiff-backup from build
    - fetch automatically the rdiff-backup.exe binary

rdiff-backup can backup, increment and restore backups with this commit.
More tests should follow but for now it closes issue #106


ericzolf:
The result of the work done on this branch is:
[rdiff-backup.exe.zip](https://github.com/rdiff-backup/rdiff-backup/files/3746833/rdiff-backup.exe.zip)

Everybody is welcome to unpack and use. Please create specific issues if you still find more issues with this version. The tests have (IMHO) never been written to work under Windows so we need to rely on manual testing for now.

PR #144, opened 2019-10-20
Add the --no-tar option to tools/makedist

ardovm:
The script `tools/makedist` can only be used to make a compressed tar archive.
It generates a temporary directory, makes the archive and then wipes it.
It could be useful to avoid the last part, i.e. the script could be used to make a ``plain'' directory with all the source files.

This is an exact replica of PR #139 but the repository, this time, should be fine.


ardovm:
Thank you!

PR #152, opened 2019-10-26
Replace Version.py by scm_version

ikus060:
To fix issue #66

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,11 +11,6 @@ from distutils.debug import DEBUG
 import distutils.command.clean
 from distutils import log
 
-from src.rdiff_backup import Version
-
-version_string = Version.version
-
-
 # --- handling compilation and linking with librsync ---
 
 lflags_arg = []
@@ -59,10 +54,6 @@ class build_templates(Command):
         ('template-files=', None, 'list of tuples of source template and destination files'),
         # TODO we could add the replacement dict as well but not for now
     ]
-    replacement_dict = {
-        "version": version_string,
-        "month_year": time.strftime("%B %Y", time.localtime(time.time()))
-    }
 
     def initialize_options(self):
         """Set default values for options."""
@@ -92,10 +83,14 @@ class build_templates(Command):
     def run(self):
         if DEBUG:
             self.debug_print(self.distribution.dump_option_dicts())
+        replacement_dict = {
+            "version": self.distribution.get_version(),
+            "month_year": time.strftime("%B %Y", time.localtime(time.time()))
+        }
         for template in self.template_files:
             self.make_file(
                 (template[0]), template[1],
-                self.make_template, (template[0], template[1], self.replacement_dict),
+                self.make_template, (template[0], template[1], replacement_dict),
                 exec_msg='templating %s -> %s' % (template[0], template[1])
             )
 
@@ -135,7 +130,7 @@ class clean(distutils.command.clean.clean):
 
 setup(
     name="rdiff-backup",
-    version=version_string,
+    use_scm_version=True,
     description="Local/remote mirroring+incremental backup",
     author="The rdiff-backup project",
     author_email="rdiff-backup-users@nongnu.org",
@@ -158,7 +153,7 @@ setup(
     data_files=[
         ("share/man/man1", ["build/rdiff-backup.1", "build/rdiff-backup-statistics.1"]),
         (
-            "share/doc/rdiff-backup-%s" % (version_string,),
+            "share/doc/rdiff-backup",
             [
                 "CHANGELOG",
                 "COPYING",
@@ -184,4 +179,5 @@ setup(
         'build_py': build_py,
         'clean': clean,
     },
+    setup_requires=['setuptools_scm'],
 )
--- a/src/rdiff_backup/Globals.py
+++ b/src/rdiff_backup/Globals.py
@@ -20,10 +20,15 @@
 
 import re
 import os
-from . import Version, log
+from . import log
 
 # The current version of rdiff-backup
-version = Version.version
+# Get it from package info or fall back to DEV version.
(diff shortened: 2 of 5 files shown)


src/rdiff_backup/Globals.py
```diff
@@ -20,10 +20,14 @@
...
 import re
 import os
-from . import Version
 
 # The current version of rdiff-backup
-version = Version.version
+# Get it from package info or fall back to DEV version.
+try:
+    import pkg_resources
+    version = pkg_resources.get_distribution("rdiffweb").version
```
ericzolf: You probably mean rdiff_backup instead of rdiffweb here.

src/rdiff_backup/Globals.py
```diff
@@ -20,10 +20,14 @@
 
 import re
 import os
-from . import Version
 
 # The current version of rdiff-backup
-version = Version.version
+# Get it from package info or fall back to DEV version.
+try:
+    import pkg_resources
```
ericzolf: That implies a dependency on Python's setuptools at runtime, correct? This should be at least documented in the README.md

src/rdiff_backup/Globals.py
```diff
@@ -20,10 +20,14 @@
...
+    version = pkg_resources.get_distribution("rdiffweb").version
+except:
+    version = "DEV"
```
ericzolf: I get DEV when I run `PYTHONPATH=$PWD/build/lib.linux-x86_64-3.7 PATH=$PWD/build/scripts-3.7:$PATH rdiff-backup --version` (see DEVELOP.md). Would there be a way to get the calculated version without installing?

I'm wondering if we should put a more explicit ("Not.Installed.See.Readme.Md"), or a tracked ("1.3.6") version, or a combination of both ("1.3.6.SeeREADMEmd"), also because this version will appear if someone forgets to install setuptools.

setup.py
```diff
@@ -70,7 +66,7 @@
     data_files=[
         ("share/man/man1", ["docs/rdiff-backup.1", "docs/rdiff-backup-statistics.1"]),
         (
-            "share/doc/rdiff-backup-%s" % (version_string,),
+            "share/doc/rdiff-backup",
```
ericzolf: Rrr, that's pretty much standard to have the docs versioned under Linux. No trick to keep it that way?

src/rdiff_backup/Globals.py
ericzolf: Another thought which might make everybody happy: replace the line 30 with the following, obviously keeping Version.py in this case
```python
except:
    from . import Version
    version = Version.version
```
This means that we'd need to track the version as Git tag _and_ in Version.py, but that would be the price to pay to make everybody happy. I would put the first as v1.3.6 and the 2nd as 1.3.6.0 so that we can differentiate and the generated version will always be higher unless it is exactly the same (Git) version.

src/rdiff_backup/Globals.py
ikus060: Yes, now you know where I copy-paste it from ! :)

src/rdiff_backup/Globals.py
ikus060: Not sure what you want me to add to the README.md. I assume setuptools to be installed in all the possible scenario: 
1. install from pip: pip depends on setuptools
2. install from `python setup.py install`: setup.py already depends on `setuptools`
3. install from a deb or rpm: setuptools should to added as a dependencies in the deb or rpm

src/rdiff_backup/Globals.py
ikus060: If we go with scm_version, the goal is to avoid tracking the version manually. Keeping it in the code doesn't seem a very nice solution.

> I get DEV when I run PYTHONPATH=$PWD/build/lib.linux-x86_64-3.7 PATH=$PWD/build/scripts-3.7:$PATH rdiff-backup --version (see DEVELOP.md). Would there be a way to get the calculated version without installing?

You need to run either "install" or "develop" on the project to get the version. 

We could get the version without installing if I add "setuptools_scm" as a dependencies.

setup.py
ikus060: Before doing the change, I had a look at my `/usr/share/doc/` directory and I don't see any version in the directories name. In Debian, this folder must matches the packages name.

ericzolf: You're right, can't remember where I saw this, Forget about it.

src/rdiff_backup/Globals.py
ericzolf: `./setup.py develop --user` is almost good, it leaves only little traces on my system. And I see that `./setup.py sdist` creates a source package that @ottok could use to do his Debian package and which doesn't forget the version generated when taken out of Git. If Otto is fine with it, he can resolve the conversation.

src/rdiff_backup/Globals.py
ericzolf: Seen like this...

src/rdiff_backup/Globals.py
ericzolf: I would still prefer a more explicit "NotInstalledDevVersion" or similar default version.

ericzolf:
@ikus060 if @ottok doesn't react, just rebase your PR and I'll merge it, I don't want to keep you too much waiting, and if you could consider https://github.com/rdiff-backup/rdiff-backup/pull/152#discussion_r339787311 I'd appreciate it (but wouldn't make it a must).

ikus060:
Before merging to master, we need to create a Tag for our last "release" 1.3.6 or something. Then scm_version will generate the version accordingly.

ericzolf:
We made it! The feature is in.

ikus060:
Yeah !
:D
--
Patrik Dufresne

PR #147, opened 2019-10-26
More documentation cleanups

frankcrawford:
This PR:
- Make distributed documentation markdown where possible,
- Delete files that can be generated from markdown files,
- Move all user documentation to docs directory,
- Update build files for the new files and locations,
- Remove tools/makeweb as it does not match current setup and is not used any more.

PR #154, opened 2019-10-27
Fix typo and indentation of tools/makedist

ardovm:
This is a continuation of PR #144. I was unable to reopen it, so I am proposing this new PR.

Apparently, the merged file contained one typo and also indentation errors (spaces and tabs were mixed).

I saw that Hound accepted the PR and I expected it to have checked PEP 8 compliance.

I am very sorry for this extra work.

Shall we take the occasion to change all indentation to spaces? Please let me know and I will make another commit.


ardovm:
Duh... there are some more problems... please hold on

frankcrawford:
The issue reason you can't reopen it is that it is actually changes from a different PR #147.

What other issues have you found?

ardovm:
> The issue reason you can't reopen it is that it is actually changes from a different PR #147.
> 
> What other issues have you found?

Thank you for the pointer.

One additional problem I found, is that it does not re-make the exact directory layout as the source tree. The `rdiff-backup` and `rdiff-backup-statistics` scripts are not copied into the `src` directory as they should, but rather into the parent directory.

I have been using the script `tools/makedist` as it was _necessary_ for building the latest versions, prior the migration to Python 3 and all the great work you have been doing in the last weeks.

As you commented in PR #139, this script may still be handy, but I still have to understand whether it is still necessary to build rdiff-backup today.

Unfortunately, my free time has just expired; it may take me some days to get my hands back onto this problem. If you would like to close this PR as it _does_ solve a problem for now, then go ahead and I will reopen another one to address the other issues.

Thank you in advance for your understanding, and also your work on this program!

ikus060:
About makedist,

I had a look at it yesterday and all of this seams very outdated. It was
probably required 10year ago, but todays the out of the box tools provided
by setuptools provide all the feature we need. I recommend to leverage them
instead of trying to fix old and deprecated home made script.

> I have been using the script tools/makedist as it was *necessary*

Why do you need makedist ? To generate a tar.gz ? And then what ?

--
Patrik Dufresne Service Logiciel inc.
http://www.patrikdufresne.com <http://patrikdufresne.com/>/
514-971-6442
130 rue Doris
St-Colomban, QC J5K 1T9

ardovm:
> About makedist, I had a look at it yesterday and all of this seams very outdated. It was probably required 10year ago, but todays the out of the box tools provided by setuptools provide all the feature we need. I recommend to leverage them instead of trying to fix old and deprecated home made script.

It is absolutely ok for me to scrap it, if the current source tree is organised to allow running `setup.py` "out of the box" 

>> I have been using the script tools/makedist as it was *necessary*
> Why do you need makedist ? To generate a tar.gz ? And then what ?

For what I could understand of rdiff-backup 1.xx sources, the software could only be compiled out of a .tar.gz archive. And the `makedist` script was the way to build such archive out of the source tree. `setup.py` could not be called directly from the source tree, because it expected some files to be in different places (and possibly for other reasons, that were all addressed by the `makedist` script).

Again: if you and the other developers, who know this software much better than me, believe that `makedist` is outdated, then please just delete it. I am just trying to follow an old path, but I am ready to abandon it in favor of a newer one.

ardovm:
FWIW I am putting my current work in my repository ardovm/rdiff-backup-build here on GitHub.
It _should_ work for Win32. It is currently based on the rdiff-backup branch from this PR.

ericzolf:
@ardovm what are you working on in your rdiff-backup-build repo? Not that we're doing double work... Master branch already builds on Windows.

ericzolf:
I'd vote for getting rid of makedist: `setup.py sdist` creates a tarball, with `-k` it keeps the temp directory, if someone needs it.

frankcrawford:
One other function that makedist does that still hasn't been covered is updating the version in non-python files such as the man pages, spec files, etc.  What is the suggestion to cover this?

Also, we do need to document some of these item somewhere, and while I'm at it make it easy to see what is the current version number somewhere where someone just browsing by can find it.  I'd suggest in the README.md file.

ericzolf:
Hi Frank,

there are 2 different topics:

1. I searched a bit and couldn't find a way to do templating within setup.py. It's surprises me a bit as it should be some rather common requirement. Need to check harder. If anybody has an idea or better searching foo?!
2. Version visible in Git: `./setup.py --version` knows it and we shouldn't burden ourselves with more places to maintain the version.

KR, Eric

ardovm:
> @ardovm what are you working on in your rdiff-backup-build repo? Not that we're doing double work... Master branch already builds on Windows.

My rdiff-backup-build repo wants to be a self-contained build system for librsync and rdiff-backup. With one command, I want to be able to get a binary distribution of rdiff-backup, i.e. an EXE file under windows and... something similar on Linux (still have to make my mind about it). This, starting from the sources of librsync and rdiff-backup.

If the rdiff-backup source tree was already capable (or almost capable) of doing this... then I would be happy! But I could only find Ansible-related documentation... and I have never used Ansible. I need an exe «here, now, on my machine». I hope this makes sense.

Moreover, there seem to be no way to build librsync, but the build documentation for rdiff-backup seem to just require it to be ready. I think that integrating the two steps would make my life easier.

If there were already a possibility of doing the above, then I would be happy to help documenting it.

ericzolf:
@ardovm understood - different discussion then.

Back to makedist: I couldn't find any way to do templating with setuptools (not even a plug-in), so:

1. I created an issue in the project at https://github.com/pypa/setuptools/issues/1895
2. with little effort, we could add (as a workaround) the feature to setup.py, here a proof-of-concept:

```python
replacement_dict = {
        "some_library": "LIBVAL",
        "some_file": "FILEVAL"
    }
with open(sys.argv[1],"r") as infp, open(sys.argv[2],"w") as outfp:
    for line in infp:
        if ("{{" in line):
            for key, value in replacement_dict.items():
                line = line.replace("{{ %s }}" % key, value)
        outfp.write(line)
```

Nobody has by chance experience with writing setuptools plug-in because this could be the correct way to do it? Else doing the templating before calling setup(...) could do the trick as well.

frankcrawford:
Where is this PR going?  Either we need to fix up the couple of bugs in makedist, at least for the short term, while deciding on the long term future,, or drop it now and add the sort of plugin described above.

ericzolf:
The setuptools code is pretty much spaghetti, I have difficulties to find my way... I'm looking at a dirtier/simpler approach.

ericzolf:
It took more time and effort than expected but the PR #173 is now ready for merging and I'd like to close this PR. Anything speaking against it before I do it?

ericzolf:
(sorry, pressed the wrong button, re-opening)

ardovm:
> It took more time and effort than expected but the PR #173 is now ready for merging and I'd like to close this PR. Anything speaking against it before I do it?

Ok for me. Thank you!

PR #157, opened 2019-11-01
Adapt setup-testfiles.sh to handle new location for test files

### LINKED ISSUE #54: Decide where to store test files

ericzolf:
The test files tar ball is now in the rdiff-backup-filesrepo Git repo.
Move also the script to the tools sub-directory.
Closes #54


ericzolf:
Can someone validate my choice?

ericzolf:
@afoster @ikus060 even if you can't merge, can you please review and tell me your opinion in absence of @ottok ? It's a rather important change and I'd like to be sure I don't oversee anything before I merge the PR.

ericzolf:
I allow myself to merge without feedback, we can still change the location if we find a better solution, and at least I'm not anymore the only one able to modify the files, so I think the situation improved even if it's not perfect.

afoster:
Sorry for the late reply @ericzolf , I have no problem with this.

PR #159, opened 2019-11-01
Add info about path where there are too many current_mirror incs

### LINKED ISSUE #37: Found too many current_mirror incs doesn't say where
When displaying this error message the path to the culprit directory should be included.

ericzolf:
Closes #37


ericzolf:
Thanks. Merging!

PR #171, opened 2019-11-09
Get rid of Docker when using Travis

ericzolf:
Travis is already a clean environment and it takes one third to half
of the time to create the Docker image, hence it makes sense to get
rid of it, while still allowing users to use Docker in order to have
a clean testing environment.


ericzolf:
This takes roughly one minute less from 4.5 down to 3.5 minutes, meaning 20-20% less time. I'd say, it worth the effort. What do others think?

Issue #170, opened 2019-11-09
bash completion: do not change user's `extglob` shell option


MestreLion:
Quoted from https://mywiki.wooledge.org/glob :
> If your code must be sourced and needs extglob, ensure it preserves the original setting from your shell:
```shell
# remember whether extglob was originally set, so we know whether to unset it
shopt -q extglob; extglob_set=$?
# set extglob if it wasn't originally set.
((extglob_set)) && shopt -s extglob
# Note, 0 (true) from shopt -q is "false" in a math context.

#... do stuff ...

# unset extglob if it wasn't originally set
((extglob_set)) && shopt -u extglob
```

`/tools/rdiff-backup.bash` should be changed to add this.

ericzolf:
Fine with this. Can you provide a PR to close this issue?

MestreLion:
Sure! It's on the way...

PR #172, opened 2019-11-09
Bash completion fixes

### LINKED ISSUE #169: bash completion: outdated path
Since the bash completion script was recently moved from Debian to upstream repository I'll fill the bug report here:

Currently the script is installed at `/etc/bash_completion.d/rdiff-backup`, but from at least 2012 bash completions default vendor dir is `/usr/share/bash-completion/completions`. This can be checked with `pkg-config --variable=completionsdir bash-completion`. Currently `/etc/bash_completion` and files at `/etc/bash_completion.d` are mere shims/stubs that source from the correct path.

Sources:
https://github.com/scop/bash-completion
https://salsa.debian.org/debian/bash-completion

if needed I can shortly provide a patch to fix this.

MestreLion:
Reorganize and fix bash completion files

Closes #169 and #170 

Note: URL to original `rdiff-backup` bash-completion file in Debian repo was changed to a specific commit SHA and not a direct download from `master`, as it is likely the Debian file will be moved/removed in the future now that we're providing it upstream.


ericzolf:
Looks good to me. Merging. Thanks.

PR #173, opened 2019-11-09
Add templating feature to setup.py and remove makedist

ericzolf:
This is the alternative suggestion to #154 - this is pretty much work in progress, consider it as proof of concept for now. Feedback welcome.


frankcrawford:
@ericzolf  A question on this, does the substitution also get run when you run a `setup.py sdist'?

ericzolf:
That's not the plan because the files are modified (read: similar to compiled) and hence should IMHO not be part of the source distribution, but I haven't yet checked.

frankcrawford:
Ahh, you are probably correct, since the generation of an RPM runs setup.py, so it would be done at that step, so you don't want it done multiple times.

ericzolf:
This said, I'm quite perplex: everything seems to work as I'd expect on my laptop but it doesn't work in Travis CI, as if the file wouldn't exist... I don't see what the relevant difference is, but I'm also surprised that setup.py has a `bdist_wheel` command, which it doesn't have on my laptop...

ericzolf:
A colleague of mine, Nils, gave me some very useful hints:
* https://jichu4n.com/posts/how-to-add-custom-build-steps-and-commands-to-setuppy/
* https://docs.python.org/3/distutils/apiref.html
* https://docs.python.org/3/distutils/extending.html
We're back in business!

ericzolf:
After some discussion on the #pypa IRC channel and some delving into the source code of distutils/setuptools (*sigh*), it looks like I've got it right now.

ardovm:
Works for me! Great job!

PR #156, opened 2019-10-31
Finish cleaning-up the code to have it PEP8 conform

ericzolf:
- do all the necessary changes to make flake8 fully happy
  (add a few `noqa` mainly because imports are not completely clean)
- make the flake8/test-static step mandatory in Travis CI
- closes #60

PR #39, opened 2019-01-02
Change directory mode before deleting them

ardovm:
Apparently, rdiff-backup does not take into account the possibility that folders could not be deletable.
This pull request is about adding a call to os.chmod(), that should render modifiable any directory, before attempting to delete it.

Please note that I only tested this patch under Win32.

A simple use case is with folders with custom icons. When the user sets a custom icon on a folder, Windows also sets the "system" attribute on the folder.

I hope this can be useful.

PR #176, opened 2019-11-12
Change directory mode before deleting them

ardovm:
Apparently, rdiff-backup does not take into account the possibility that folders could not be deletable.
This pull request is about adding a call to os.chmod(), that should render modifiable any directory, before attempting to delete it.

Please note that I only tested this patch under Win32.

A simple use case is with folders with custom icons. When the user sets a custom icon on a folder, Windows also sets the "system" attribute on the folder.

I initially opened PR #39 but then I messed up my repository, so here we are.

The second commit in this PR tries to solve a bug introduced with PR #156, although it is not strictly related to the directory permissions.


ericzolf:
Looks good to me. Merged and thank you!

PR #182, opened 2019-11-17
Add bash-completion file to rpm specs

frankcrawford:
Follow up for adding bash-completion file to distribution.


ericzolf:
Well spotted. Thanks!

PR #185, opened 2019-11-17
Create a script to generate a class and import overview

ericzolf:
It's a simple shell script and it could be done better (and I know there
are some tools to do such analysis), but I wanted to get a first impression
of where we stand. The summary is that the class structure is perhaps not
too bad but the imports are all over the place.

To generate the graphs yourself, you'll need plantuml (see the script comments).

I've attached the two generated graphs for anybody interested to review. I'd like in the future to avoid circular imports/dependencies, but that's post-2.0 outlook, it'll require quite a lot of re-factoring.


ericzolf:
The promised examples of graphs created: imports followed by classes.

rdiff-backup-imports

rdiff-backup-classes

PR #189, opened 2019-11-23
Adding more python version to travis build Close #48

### LINKED ISSUE #48: Create a CI pipeline
We should create a CI pipeline that runs on every commit and every pull request.

## What tests to run?

The new Python 3 branch introduces the Tox testing framework, so we should at least run Tox.

Tox should be extended to test as much as possible. We need to look into if Tox is enough, or if we need some separate tool to do acceptance tests and run in bash some actual rdiff-backup jobs/cycles. I would be interested in having some tests that generate a backup archive with version 1.2.8 and then runs the latest git development version to further process that archive and make new increments etc. This would show us if the code remains backwards compatible or not.

We could also run some static code analysis. Since this is a Python project, running flake8 on all Python files would seem natural. Requiring everybody to follow PEP-8 in their Python code is probably OK.

## What testing system to use?

Github does not have any built in testing system (unlike Gitlab which has Gitlab CI). Personally I have lot's of experience with Travis-CI and it is free for open source projects, so we could adopt it.

ikus060:


ericzolf:
Stupid error (I had already made the same in my thoughts), the Makefile has py37 in hard... either it can be replaced by a variable, or the shortcut `py` works as we need it: "the environment py uses the version of Python used to invoke tox" according to https://tox.readthedocs.io/en/latest/example/basic.html#a-simple-tox-ini-default-environments

ikus060:
Using "py" would make sens to me. Will fix and re-push

tox.ini
```diff
@@ -7,7 +7,7 @@
 # Use tox_slow.ini for longer running tests.
 
 [tox]
-envlist = py35, py36, py37, py38, flake8
+envlist = py, flake8
```
ericzolf: Please revert this change:
1. py is defined by default, no need to explicitly list it
2. it's actually the same for pyXX but it's a documentation of the "support matrix"

tox_root.ini
```diff
@@ -8,7 +8,7 @@
 # Use tox_slow.ini for longer running tests and tox.ini for normal tests.
 
 [tox]
-envlist = py35, py36, py37, py38
+envlist = py
```
ericzolf: ditto!

tox_slow.ini
```diff
@@ -7,7 +7,7 @@
 # Call with `tox -c tox_slow.ini`
 
 [tox]
-envlist = py35, py36, py37, py38
+envlist = py
```
ericzolf: And ditto again!

PR #192, opened 2019-11-24
Updated to include setuptools_scm as a build requirement.

frankcrawford:
This follows up with a few changes for the recent build process changes, especially SCM.

This also means that the tar file needs to include the `PKG_INFO` file, which is generated by `setup.py sdist`

### DIFF
--- a/tools/rdiff-backup.spec.template
+++ b/tools/rdiff-backup.spec.template
@@ -56,8 +56,7 @@ rm -rf $RPM_BUILD_ROOT
 %{_libdir}/python%{PYTHON_VERSION}/site-packages/rdiff_backup/*.so
 %{_datadir}/bash-completion/completions/rdiff-backup
 %ghost %{_libdir}/python%{PYTHON_VERSION}/site-packages/rdiff_backup/*.pyo
-%doc CHANGELOG COPYING README.md
-%doc docs/FAQ.md docs/examples.md docs/DEVELOP.md docs/Windows-README.md
+%doc CHANGELOG COPYING README.md docs/*.md
 
 %changelog
 * Sat Nov 23 2019 Eric Lavarde <ewl+rdiffbackup@lavar.de> 1.4.0b0-1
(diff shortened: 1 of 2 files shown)


tools/rdiff-backup.spec.template
```diff
@@ -53,9 +56,13 @@ rm -rf $RPM_BUILD_ROOT
 %{_libdir}/python%{PYTHON_VERSION}/site-packages/rdiff_backup/*.so
 %{_datadir}/bash-completion/completions/rdiff-backup
 %ghost %{_libdir}/python%{PYTHON_VERSION}/site-packages/rdiff_backup/*.pyo
-%doc CHANGELOG COPYING docs/FAQ.md docs/examples.md docs/README.md docs/DEVELOP.md
+%doc CHANGELOG COPYING README.md
+%doc docs/FAQ.md docs/examples.md docs/DEVELOP.md docs/Windows-README.md
```
ericzolf: stupid question but wouldn't `%doc docs/*.md` work and be easier to maintain?

tools/rdiff-backup.spec.template
```diff
@@ -2,18 +2,22 @@
 %define PYTHON_VERSION %(%{PYTHON_NAME} -c 'import sys; print(sys.version[:3])')
 %define NEXT_PYTHON_VERSION %(%{PYTHON_NAME} -c 'import sys; print("%d.%d" % (sys.version_info[0], sys.version_info[1]+1))')
 
+# Note this in not necessarily the same at the version, so confirm
+%define gittag %{version}
```
ericzolf: Ah, good point, for the first beta version it isn't, but we can and should make sure that it is the same

ericzolf:
> This also means that the tar file needs to include the `PKG_INFO` file, which is generated by `setup.py sdist`

The file you've taken as source is generated by `./setup.py sdist` and it makes sure that the versioning works properly (as you recognized, continuing to use the source files provided automatically by GitHub wouldn't have worked).

tools/rdiff-backup.spec.template
frankcrawford: Yeah, probably.  I'll update them with that.

tools/rdiff-backup.spec.template
frankcrawford: Thanks.

tools/rdiff-backup.spec.template
frankcrawford: Done.

PR #193, opened 2019-11-26
Add tox config file to generate distribution

ericzolf:
For now can be used to generate wheels for all supported Python versions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ MANIFEST
 
 # setup.py installation files
 src/rdiff_backup.egg-info/
-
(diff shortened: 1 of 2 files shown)


ikus060:
Take note, this will generate wheel packages incompatible with pypi. You need to complies to some criteria to make this work and I don't think travis provide the required environment to make it work.

But I'm guessing it's only to automate some of the manual step you are doing right now ? Right ?

ericzolf:
Correct, and to make sure that we can create easily wheel packages for all 4 Python versions we support.

PR #197, opened 2019-12-01
Fix selection of overlapping patterns

### LINKED ISSUE #22: No error raised in certain occasions when excluding files the wrong way

ericzolf:
- closes #137 (thanks to @ilario for the first steps)
- closes #22
- normally selection patterns not relative to the root path lead to an error,
  but 'selection' was accepted for 'select' because they start alike.
- test accordingly added to testing/selection.py

### DIFF
--- a/src/rdiff_backup/selection.py
+++ b/src/rdiff_backup/selection.py
@@ -86,6 +86,7 @@ class Select:
         self.selection_functions = []
         self.rpath = rootrp
         self.prefix = self.rpath.path
+        self.prefixindex = tuple([x for x in self.prefix.split(b"/") if x])
 
     def set_iter(self, sel_func=None):
         """Initialize more variables, get ready to iterate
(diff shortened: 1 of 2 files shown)


ilario:
It works :D Many thanks!

PR #200, opened 2019-12-03
Replace cmodule.c by python implementation

ikus060:


ericzolf:
It's work in progress and I haven't looked at all the details but:

- the quote/unquote functions belong in the metadata module, let's not add cross-module dependencies, that creates more spaghetti-code.
- they should be renamed because they apply not only to ACLs but also to EAs, something like `meta_quote` would be fine.
- have a look if it doesn't have a too negative impact on performance, like `time` the ACLs/EAs test runs before and after the change (e.g. add time to the line in the tox file).
- would something like that be an option: https://stackoverflow.com/questions/555705/character-translation-using-python-like-the-tr-command (`str.translate(str.maketrans)`) ?

ikus060:
@ericzolf 
Thanks for your feedback. Yes it's a WIP. My end goal is to get rid of all the C extension. That would simplify alot how we are distributing the binaries and will increase the number of platform supported. e.g.: RPi

> * the quote/unquote functions belong in the metadata module, let's not add cross-module dependencies, that creates more spaghetti-code.

Will move the code and the test.

> * they should be renamed because they apply not only to ACLs but also to EAs, something like `meta_quote` would be fine.

Will rename the function.

> * have a look if it doesn't have a too negative impact on performance, like `time` the ACLs/EAs test runs before and after the change (e.g. add time to the line in the tox file).

I intend to do some performance testing. I just didn't had a good environment yesterday to do the coding.

> * would something like that be an option: https://stackoverflow.com/questions/555705/character-translation-using-python-like-the-tr-command (`str.translate(str.maketrans)`) ?

I will consider this function and may also measure the various implementation.

ikus060:
@ericzolf 
Had a look at `bytes.translate()` and `bytes.maketrans()`. To make it work I need to convert the byte to str, then call translate, then convert str to bytes. It's working but I don't find it fancy. I've compare the implementation and it doesn't make any differences.

The `quote` function is performing well. The issues is with `unquote`. I've tried various implementation and it's not performing well depending of the input string...

Attached is the best I found. Compare to the C implementation is't not performing well, but in real life it's shoud be good enough.

```
# time python3 -m metadata
# real	0m42.834s
# user	0m42.831s
# sys	0m0.004s

_safe = b'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<>?@[]^_`{|}~'
_quote_map = {}
_unquote_map = {}
for i in bytearray(range(1, 256)):
    c = bytes([i])
    _quote_map[i] = c if c in _safe else '\\{:03o}'.format(i).encode('ascii')
    _unquote_map['{:03o}'.format(i).encode('ascii')] = c


def meta_quote(s):
    """Quote filename for ACL usages."""
    assert isinstance(s, bytes)
    return b''.join(map(_quote_map.__getitem__, s))


def meta_unquote(s):
    """Unquote filename from ACL format."""
    assert isinstance(s, bytes)
    bits = s.split(b'\\')
    res = bits[0]
    for item in bits[1:]:
        res += _unquote_map[item[:3]]
        res += item[3:]
    return res

line = bytes(bytearray(range(1, 256)))
for i in range(1,1000000):
  v = meta_unquote(meta_quote(line))

print('done')
```

Compare to C implementation:
```
# time python3 -m metadata
#real	0m1.103s
#user	0m1.095s
#sys	0m0.008s

from rdiff_backup import C

line = bytes(bytearray(range(1, 256)))
for i in range(1,1000000):
  v = C.acl_unquote(C.acl_quote(line))

print('done')
```
I will integrate all of this into the Pull Request

PR #204, opened 2019-12-03
Add zstd to compressed file extensions.

whataboutpereira:
After some experimenting with more compression formats, I hope you'll add this one too. :)


ericzolf:
If we add too many extensions, there will be soon a performance impact on the regexp... 😉 thanks!

whataboutpereira:
Haha. That's a good problem to have. :D Really impressed by zstd atm. Better compression than gzip on my sql backups plus a few times faster.

Issue #208, opened 2019-12-06
Win32: function _librsync_new_patchmaker() may fail with errno 24 "Too many open files"


ardovm:
There is a return path in src/_librsyncmodule.c:393 that allows the function to return NULL without raising any Python exception.

This happens on some of my Win32 test systems.

ardovm:
Here is the error message:
```
UpdateError: '/path/to/file' fdopen(): Too many open files (24)
```

ardovm:
https://github.com/ardovm/rdiff-backup/blob/issue208/issue208.py is a script that creates enough files to exploit this problem.

Intended usage:
```
./issue208.py temp_base
./rdiff-backup.exe temp_base temp_destination
./issue208.py temp_base
./rdiff-backup.exe temp_base temp_destination # <---- this will fail
```

ikus060:
Ok, I made a couple of test on my side on Linux and Windows:
Linux | rdiff-backup 1.2.8 : OK
Windows | rdiff-backup 1.2.8 : OK
Windows | rdiff-backup 1.4.0b0 : FAIL

But it doesn't fail with same error has you.

```
C:\Users\vmtest\Desktop>C:\Users\vmtest\Downloads\rdiff-backup-1.4.0b0-win64.exe
 test_base test_destination
Exception '[WinError 123] La syntaxe du nom de fichier, de répertoire ou de volu
me est incorrecte: b'test_destination/rdiff-backup-data/current_mirror.2019-12-1
5T08:36:05-05:00.data'' raised of class '<class 'OSError'>':
  File "rdiff_backup\Main.py", line 384, in error_check_Main
  File "rdiff_backup\Main.py", line 406, in Main
  File "rdiff_backup\Main.py", line 342, in take_action
  File "rdiff_backup\Main.py", line 422, in Backup
  File "rdiff_backup\Main.py", line 609, in backup_final_init
  File "rdiff_backup\Main.py", line 1112, in checkdest_if_necessary
  File "rdiff_backup\Main.py", line 1076, in checkdest_need_check
  File "rdiff_backup\restore.py", line 56, in get_inclist
  File "rdiff_backup\rpath.py", line 1375, in append
  File "rdiff_backup\FilenameMapping.py", line 146, in __init__
  File "rdiff_backup\rpath.py", line 1047, in __init__
  File "rdiff_backup\rpath.py", line 1073, in setdata
  File "rdiff_backup\rpath.py", line 352, in make_file_dict

[...]

PR #209, opened 2019-12-06
Raise _librsync.librsyncError if fdopen(3) fails

ardovm:
Address upstream issue #208


ikus060:
Would be nice to have a unit test to reproduce this issue. While I have not clue how to reproduce the problem.

ardovm:
I have encountered this problem once: I had just migrated some systems at work to 1.4.0beta. 

I need to use the "patched" version and see what the error message is. It will take me some (working) days, I believe.

ardovm:
Got it. Please see the issue.

ericzolf:
I think, there was a misunderstanding: @ikus060 asked if you could add a unit test to the tox.ini, which would mean, understand how to reproduce the issue, @ardovm signaled readiness to do it, and I was waiting for @ardovm to add this unit test to the PR, which obviously isn't happening as you seem to consider the job done with the error message :confused: Sooo, we could all have waited for a very long time...

Did I understand the situation correctly? Shall I merge, even without test case? I still would like to understand in which cases the error happens though.

ardovm:
Hello All,

> I think, there was a misunderstanding: @ikus060 asked if you could add a unit test to the tox.ini, which would mean, understand how to reproduce the issue, @ardovm signaled readiness to do it, and I was waiting for @ardovm to add this unit test to the PR, which obviously isn't happening as you seem to consider the job done with the error message Sooo, we could all have waited for a very long time...

Actually, I have not had any time to dedicate to this issue, unfortunately.

There are two problems here:

 1. there is a possibility that a C function returns NULL without setting an appropriate Python exception. This is a bug.
 1. the Win32 build may fail with the "Too many open files" error message.

> Did I understand the situation correctly? Shall I merge, even without test case? I still would like to understand in which cases the error happens though.

I believe that problem 1. has to be fixed.

The next topic is understanding the reason for problem 2.

As we discussed on another issue, I have no experience of tox. And, most important, the current state of the Windows CI integration shows that it will not even start. Therefore, what I can do is providing another means of testing this, such as a Python script.
But I have to find the time to prepare it.

If you want to wait for the solution of both issues 1. and 2., you can wait. It is the same for me, as I will be working on my own branch to troubleshoot this. I will push any proposed fixes to this branch.

ardovm:
I just uploaded the script `issue208.py` that creates _enough_ filescrs and directories to get the "Too many open files" message.
I will describe it in the issue report.

ikus060:
@ericzolf I would merge changes made in `src/_librsyncmodule.c` only
@ardovm Could you remove the issue208.py fikles from this PR. This should not be part of rdiff-backup source code. You wrote this only to help in your debugging.

ardovm:
I found a tentative fix of issue #208 and committed it here. It involves the same file `src/librsyncmodule.c`

At this point, I would delay merging this branch until we are convinced that the reasoning behind the patch is totally right.
Then, I would like to add an intermediate error reporting function based on `errno` (currently, the error reporting code for `fdopen()` and `dup()` is duplicated). But it is not worth writing it until we agree that this is the way to go, IMHO.

I am of course open to any remarks and suggestions.

src/_librsyncmodule.c
```diff
@@ -389,9 +389,21 @@ _librsync_new_patchmaker(PyObject* self, PyObject* args)
...
+  int my_fd = dup(python_fd);
+  if (my_fd < 0) {
+      char error_string[200];
+      snprintf(error_string, 200, "fdup(): %s (%d)", strerror(errno), errno);
+      PyErr_SetString(librsyncError, error_string);
+      return NULL;
+  }
+  pm->patch_file = fdopen(my_fd, "rb"); /* same mode as in the Python code */
+  if (pm->patch_file == NULL) {
+      char error_string[200];
```
ericzolf: where does the 200 come from?

src/_librsyncmodule.c
```diff
@@ -389,9 +389,21 @@ _librsync_new_patchmaker(PyObject* self, PyObject* args)
...
+      snprintf(error_string, 200, "fdopen(): %s (%d)", strerror(errno), errno);
```
ericzolf: I think it would be better to limit the `strerror(errno)` string in size and keep the overall string at a bigger length, making sure that errno in parentheses is output in all cases.
Same applies also to above snprintf.

src/_librsyncmodule.c
ardovm: quick and dirty hack.

ericzolf:
Thanks a lot for the hard work!

PR #213, opened 2019-12-14
Add some info about python and platform to facilitate support

ericzolf:
With verbosity equal or higher to 4, the information given looks now like this:

```
Using rdiff-backup version 1.4.0b1.dev14+g3451b64
	with cpython /usr/bin/python3 version 3.7.5
	on Linux-5.3.14-300.fc31.x86_64-x86_64-with-fedora-31-Thirty_One, fs encoding utf-8
```

Can someone check how it works under Windows, and other platforms?

### DIFF
--- a/src/rdiff_backup/Main.py
+++ b/src/rdiff_backup/Main.py
@@ -25,6 +25,7 @@ import io
 import tempfile
 import time
 import errno
+import platform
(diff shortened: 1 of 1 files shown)


ericzolf:
Thanks, merging!

Issue #239, opened 2020-01-18
Incorrect hard link counts in mirror_metadata file


rknichols:
In rdiff-backup 1.4.0b1.dev4+gd9977cf, when new links are added to an existing group of hard linked files, the files that were originally in the group will not get an updated NumHardLinks count in the mirror_metadata file. Test script:
```
#!/bin/bash
mkdir tdir
echo something >tdir/File-00
for N in 10 20 30; do ln tdir/File-00 tdir/File-$N;done
rdiff-backup -b tdir tdir-bkup
sleep 2
for N in 15 25 26 ;do ln tdir/File-00 tdir/File-$N;done
ls -l tdir
rdiff-backup -b tdir tdir-bkup
cd tdir-bkup/rdiff-backup-data
New=$(ls mirror_metadata* | tail -1)
zgrep -E '^File|NumHardLinks|Inode' $New
```

Result:
```
# bash  ~/rblinktest 
total 28
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-00
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-10
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-15
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-20
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-25
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-26
-rw-r--r--. 7 root root 10 Jan 17 22:05 File-30
File .
File File-00
  NumHardLinks 4
  Inode 917506
File File-10
  NumHardLinks 4
  Inode 917506
File File-15
  NumHardLinks 7
  Inode 917506
File File-20
  NumHardLinks 4
  Inode 917506
File File-25
  NumHardLinks 7
  Inode 917506
File File-26
  NumHardLinks 7
  Inode 917506
File File-30
  NumHardLinks 4
  Inode 917506
#
```

Only the newly added files have the correct NumHardLinks count. The files in the mirror are correctly linked.

zjw:
I've submitted pull request #240 to resolve this issue.

Issue #78, opened 2019-08-14
Backup doesn't contain SHA1 hash for hard linked files


ericzolf:
<a href="https://github.com/ericzolf"><img src="https://avatars3.githubusercontent.com/u/3105685?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ericzolf](https://github.com/ericzolf)**
_Sunday May 26, 2019 at 07:01 GMT_
_Originally opened as https://github.com/ericzolf/rdiff-backup/issues/23_

----

The first one contains a hash but not the following ones, either it should be stored as usual or the reading should be smart enough to get it from the first.

Else we get something like:

```
PATH=$PWD/build/scripts-3.7:$PATH PYTHONPATH=./build/lib.linux-x86_64-3.7/ python3 build/scripts-3.7/rdiff-backup --verify build/testfiles/output
Warning: Cannot find SHA1 digest for file 'file1_linked',
perhaps because this feature was added in v1.1.1
Warning: Cannot find SHA1 digest for file 'file2_linked',
perhaps because this feature was added in v1.1.1
Every file verified successfully.
```

and the hash test fails with:

```
======================================================================
FAIL: test_session (__main__.HashTest)
Run actual sessions and make sure proper hashes recorded
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./testing/hashtest.py", line 95, in test_session
    assert hashlist == hashlist1, (hashlist1, hashlist)
AssertionError: ([None, '943a702d06f34599aee1f8da8ef9f7296031d699', '943a702d06f34599aee1f8da8ef9f7296031d699', 'eab21fb1a18b408909bae552b847f6b13f370f62', 'eab21fb1a18b408909bae552b847f6b13f370f62'], [None, '943a702d06f34599aee1f8da8ef9f7296031d699', None, 'eab21fb1a18b408909bae552b847f6b13f370f62', None])
```

ericzolf:
<a href="https://github.com/ericzolf"><img src="https://avatars3.githubusercontent.com/u/3105685?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [ericzolf](https://github.com/ericzolf)**
_Tuesday May 28, 2019 at 19:32 GMT_

----

The issue already existed with version 1.2.8/Python 2, hence will be fixed at a later stage. For now, the test will be skipped.

PR #240, opened 2020-01-19
Incorporate Hardlink patch (fixes issue #239)

zjw:
About 10 years ago, I wrote a patch to fix certain errors related to hardlinks.  My original submission is here:

https://savannah.nongnu.org/bugs/?26848

Since that time, some distributions have incorporated the patch into their own builds.  Debian, for example:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=558035

https://salsa.debian.org/python-team/applications/rdiff-backup/blob/7cd963df/debian/patches/03_fix_hardlinks.diff

The patch resolves issue #239.

I've also added some changes to the tests to go along with the patch.


ericzolf:
We stated that we wanted to have no regression since 1.2.8 and would address long standing bugs after 2.0.0 but we didn't think about bugs fixed by distros... Let me think about this and look in more details into the patch. We might release the next beta before applying the patch.

As a side remark, don't bother about the error in the pipeline, it seems due to a timeout in the Windows build. It'll most probably go away next time.

src/rdiff_backup/compare.py
```diff
@@ -103,13 +104,27 @@ def Verify(mirror_rp, inc_rp, verify_time):
...
     RepoSide.close_rf_cache()
     if not bad_files:
         log.Log("Every file verified successfully.", 3)
     return bad_files
 
 
+def get_hash(repo_rorp):
+    """ Try to get a sha1 digest from the repository.  If hardlinks
+    are saved in the metadata, get the sha1 from the first hardlink """
+    Hardlink.add_rorp(repo_rorp)
```
ericzolf: I see a potential issue with the fact that (and I know it's not your code, and I don't understand everything yet):

- add_rorp adds a hardlink if it doesn't exist but does NOT increment the "remaining numlinks" if it already exists
- whereas del_rorp does decrement it in all cases (and potentially remove it)

(sorry, still struggling with this part of the code)

zjw: Every hardlinked RORP is processed with `add_rorp` and then later with `del_rorp`.  When a hardlinked RORP is first added with `add_rorp`, its inode is added to `_inode_index` along with the total number of hardlinks for the inode.  Additional calls to `add_rorp` for other RORPs linked to that inode do not change the hardlink count.  Calls to `del_rorp` will decrement the hardlink count, and once zero, the inode is removed from `_inode_index`.  At that point, all of the RORPs that were hardlinked to the inode will have been seen and processed, and we can be sure that we will not encounter any other RORPs linked to that inode.

In theory, if a backup does not include _all_ of the files that are hardlinked to an inode, then the hardlink count will never decrement completely to zero.  That is not a serious issue.  It only means the inode will never be removed from the `_inode_index`, and so the memory that it uses can never be freed before the the backup is finished.

(In practice and when possible, users would do well to include all the hardlinks for a given inode in their backups.  Doing otherwise would mean that, when restored, those files will no longer be hardlinked the same way as when they were backed up.)

You would not want the code to simply increment and decrement the inode hardlink count as calls are made with the same inode to `add_rorp` and `del_rorp`, respectively.  You must keep the information about a hardlinked inode in memory for as long as there is a possibly of encountering that same inode again later in the backup process.

I'll also mention that this PR eliminates all of these warnings mentioned in issue #78:
```
Warning: Cannot find SHA1 digest for file 'file1_linked',
perhaps because this feature was added in v1.1.1
```
(I expect that I will eventually be submitting a PR with a suggested fix for the failed unit test mentioned in #78.)

ericzolf: Thanks for the explanations. Makes sense. Let me continue the review, it's quite a lot I need to swallow.

testing/hardlinktest.py
```diff
@@ -156,6 +156,121 @@ def testInnerRestore(self):
         assert hlout2.getinode() == hlout3.getinode()
         assert hlout1.getinode() != hlout2.getinode()
 
+    def testHardlinkBug1(self):
```
ericzolf: Please give  a more descriptive function name and a short description of the bug you're trying to reproduce and how. In X years, nobody will care about the initial bug report but the test might still be relevant.

I would also have expected a test that:
1. there is no warning about missing hash when checking the repo.
2. no discrepancy in number of links in the metadata

And I would suggest to backup with an explicit `current_time` and restore with the same time to make it clearer what you're trying to do. Other tests use 10000, 20000, etc... You can then also remove the 1s wait time.

testing/hardlinktest.py
```diff
@@ -156,6 +156,121 @@ def testInnerRestore(self):
...
+        if out3.lstat():
+            out3.delete()
+        InternalRestore(1, 1, sub_dir, out3_dir, int(time.time()))
+        out3.setdata()
+        for rp in [hlout1, hlout2, hlout3]:
+            rp.setdata()
+        assert hlout1.getinode() == hlout2.getinode()
+        assert hlout1.getinode() == hlout3.getinode()
+
+    def testHardlinkBug2(self):
```
ericzolf: Basically the same comments as for the 1st function... I'm also asking myself if it wouldn't be simpler to just add one more step to the previous function and cover the same aspects (but I'll leave it to you to decide).

ericzolf:
This said, I can confirm that the patch fixes the two issues with missing hash and wrong hardlink count in the metadata. Well done.

ericzolf:
Note to myself:
* Closes #78 
* Closes #239

testing/hardlinktest.py
zjw: Thanks.  I plan to work on this today.

zjw:
I have now updated the hardlink tests so that the mirror metadata is double-checked to make sure that it is not missing sha1 hashes and that it correctly records the number of hardlinks for files.  Those tests, along with the testing to assure that the restored files are properly hardlinked, should be sufficient to cover the issues related to the buggy hardlink code that is being fixed.

Offhand, I don't see an easy way of testing that there is no warning about missing hashes when checking the repo.  But such warnings are just symptomatic of the underlying problems which are covered by the tests.

ericzolf:
Thanks! Merged.

Issue #86, opened 2019-08-14
Decide on backup of symbolic link where linked file is excluded


ericzolf:
<a href="https://github.com/ericzolf"><img src="https://avatars3.githubusercontent.com/u/3105685?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ericzolf](https://github.com/ericzolf)**
_Thursday Jul 25, 2019 at 19:26 GMT_
_Originally opened as https://github.com/ericzolf/rdiff-backup/issues/36_

----

The test class FinalSelection(PathSetter):def testSelFilesRemote(self) contains a line which causes an error (now as seen below commented out):

```
# FIXME: should symbolic links to excluded files be included?
#assert os.lstat(Local.rpout1.append('symbolic_link').path)
```

Which means that even if the link should actually be included (and the file it links excluded), also the soft link isn't backed up.

The same behaviour can be reproduced with rdiff-backup 1.2.8 hence it's not a regression of the migration to Python 3 and will be fixed later.

zjw:
The assertion error in this issue is due to a bug in the unit test, not if rdiff-backup proper.  Pull request #242 fixes it.

The title of this issue implies a decision should be made:  "Decide on backup of symbolic link where linked file is excluded."  

Personally, I think symbolic links should still be backed up even when the linked file is excluded.  I believe that is the expected behavior, and I think it is what `tar` does.

I think this issue can be closed.

ericzolf:
It'll be closed with #242 - thanks.

PR #242, opened 2020-01-20
Fix bug in testing/cmdlinetest.py

zjw:
The bug is not in rdiff-backup, but in the unit test.

The bug was pointed out in issue #86.


ericzolf:
Well spotted, I'll merge once the tests have passed.

ericzolf:
Closes #86

PR #246, opened 2020-01-23
metadata: handle unknown group ids from file system

seife:
an unknown group id did cause all files to be backed up every time
due to an invalid comparison.
When saving metadata 'None' was converted to ":", but when reading
":" was not converted back to 'None'.


ericzolf:
Well spotted, thanks!

PR #247, opened 2020-01-23
fix invalid type in ea.read_from_rp()

seife:
TypeError: startswith first arg must be str or a tuple of str, not bytes
Not sure if thisis dependent of the used xattr package, so just convert
it to bytes if it is a string.

### DIFF
--- a/src/rdiff_backup/eas_acls.py
+++ b/src/rdiff_backup/eas_acls.py
@@ -77,6 +77,8 @@ class ExtendedAttributes:
                 return
             raise
         for attr in attr_list:
+            if isinstance(attr, str):
+                attr = bytes(attr, 'utf-8')
             if attr.startswith(b'system.'):
                 # Do not preserve system extended attributes
                 continue


ericzolf:
Can you please explain when you get the issue you're addressing, which version, OS, condition? I was the opinion that everything is bytes at this stage.

seife:
This happened on openSUSE Leap 15.1, python 3.6.10, python3-xattr version 0.9.2

I also looked into the python-xattr source to find out what's happening, but could not really understand what was going on there.

The file that tripped this is...

```
server:~ # xattr -l ~seife/tmp/radioclk-1.0.tar.gz
user.xdg.origin.url: http://www.buzzard.me.uk/jonathan/downloads/radioclk-1.0.tar.gz
```

seife:
I tried again with openSUSE Factory, which has python3 3.7.3 and python3-xattr 0.9.6
I amended the code like that for debugging:
``` diff
--- a/src/rdiff_backup/eas_acls.py
+++ b/src/rdiff_backup/eas_acls.py
@@ -76,7 +76,10 @@ class ExtendedAttributes:
                         4)
                 return
             raise
+        if attr_list:
+            print(type(attr_list), attr_list)
         for attr in attr_list:
+            print(type(attr), attr)
             if attr.startswith(b'system.'):
                 # Do not preserve system extended attributes
                 continue
```
and got this before the exception:
```
<class 'tuple'> ('user.xdg.origin.url',)
<class 'str'> user.xdg.origin.url
Exception 'startswith first arg must be str or a tuple of str, not bytes' raised of class '<class 'TypeError'>':
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/robust.py", line 35, in check_common_error
    return function(*args)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 1411, in append
    return self.__class__(self.conn, self.base, self.index + ext)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 1083, in __init__
    self.setdata()
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 1111, in setdata
    self.conn.rpath.setdata_local(self)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 1820, in setdata_local
    rpath.data['ea'] = ea_get(rpath)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/eas_acls.py", line 707, in rpath_ea_get
    ea.read_from_rp(rp)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/eas_acls.py", line 82, in read_from_rp
    if attr.startswith(b'system.'):
```

could this be a python-xattr vs pyxattr issue?

ericzolf:
Thanks for checking. I didn't even know that there are two python libraries for xattr. Anyway, we support pyxattr, not xattr. I've clarified this in MR #249 and close this MR.

seife:
Actually with python-xattr, one needs to do:
```
import xattr.pyxattr_compat as xattr
```
Then the compat layer is used and the types are compatible.
I'll do this (with a try/except...) in the openSUSE pacakge.

The problem is, that if both pyxattr and python-xattr are installed, then python-xattr is used for plain "import xattr"
Now in a linux distribution, there are certainly some programs that require python-xattr and others that require pyxattr, so it would be bad for the user experience if I added "Conflicts: python-xattr" to the package description to force only pyxattr to be installed.
I understand that from your developer perspective, this is not an issue and you prefer to keep it as simple as possible, but as a distribution packager, I need to take the whole thing into consideration :-)

ericzolf:
OK, thanks for the background information, you've convinced me. I can't re-open your MR so I'll introduce it in my MR which you're welcome to review once it's been updated.

seife:
I have it in branch python-xattr_compat, I can send a PR if you want.

https://github.com/b1-systems/rdiff-backup/tree/python-xattr_compat

ericzolf:
Sorry, I already pushed my changes to MR #249, let's continue the discussion over there.

PR #248, opened 2020-01-23
handle sparse files efficiently

seife:
This patch is originally from Erich Wheeler, posted on the mailing list
http://lists.gnu.org/archive/html/rdiff-backup-users/2011-01/msg00000.html
I ported it forward and amended it to work with GzipFile output.

This patch has been shipping in the openSUSE rdiff-backup package since 2014, so it is quite field-tested ;-)

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -67,11 +67,32 @@ class RPathException(Exception):
 def copyfileobj(inputfp, outputfp):
     """Copies file inputfp to outputfp in blocksize intervals"""
     blocksize = Globals.blocksize
+
+    sparse = False
+    """Negative seeks are not supported by GzipFile"""
+    compressed = False
+    if isinstance(outputfp, gzip.GzipFile):
+        compressed = True
+
     while 1:
         inbuf = inputfp.read(blocksize)
         if not inbuf:
             break
-        outputfp.write(inbuf)
+
+        buflen = len(inbuf)
+        if not compressed and inbuf == b"\x00" * buflen:
+            outputfp.seek(buflen, os.SEEK_CUR)
+            # flag sparse=True, that we seek()ed, but have not written yet
+            # The filesize is wrong until we write
+            sparse = True
+        else:
+            outputfp.write(inbuf)
+            # We wrote, so clear sparse.
+            sparse = False
+
+    if sparse:
+        outputfp.seek(-1, os.SEEK_CUR)
+        outputfp.write(b"\x00")
 
 
 def cmpfileobj(fp1, fp2):


ericzolf:
As much as I like the sparse file handling, I don't really like the "short reads" optimization: it avoids perhaps a few writes but it's IMHO compensated by copying the code between inbuf and buf, and it makes the code a lot more complex.

I think that without this short read handling, you can get rid of the buf variable and avoid half of the additional code. Can you have a take at it?

As a side note, I'm very hesitant at adding new features to our current beta, but no issues once 2.0.0 is out (hopefully soon).

seife:
I actually took offense on that buf handling, too, but I'm just a casual python hacker, so I forwarded more or less the original patch.

The only thing I could imagine is, that this should avoid creating sparse "holes" smaller than blocksize. But then, the whole thing is not blocksized aligned anyway, and in the end the file system will have to handle such things. I'll get rid of the buf and short-reads handling and test this in the openSUSE package, then resubmit.

ikus060:
Take note recent filesystem like BTRFS and ZFS are making this patches completely useless.
> As a side note, I'm very hesitant at adding new features to our current beta, but no issues once 2.0.0 is out (hopefully soon).
Same here. This patches was not part of Debian build. Let keep it aside

seife:
I'd rather use anything else than BTRFS and won't switch to Solaris or BSD any time soon ;-)

But out of interest: do BTRFS and ZFS not handle sparse files anymore? I know NTFS doesn't, but I don't care so much about that.

ikus060:
My point is more, many Filesystem are supporting sparse file out of the box. Without the user telling anything. I know ZFS does it very well when compression is enabled.

seife:
with compression it probably just compresses the zeroes, and the file is not really sparse afterwards.

Anyway, there are enough linux file systems that can benefit from this ;-)

ericzolf:
XFS and ext4 are still all over the place indeed. I'll look at the new patch later, have to work for real now.

seife:
We can postpone this until after the release of 2.0,
I'll expose it to some testing in the openSUSE package until then.

ericzolf:
Can you please merge master back into your branch and push it? Somehow I can't do it myself from the WebUI. I'll merge then. Thanks.

seife:
I rebased on master instead of merging, that saves us one merge commit ;-)

ericzolf:
Well, it doesn't matter too much as we squash and we did it a lot like you in the past, but in PR #255 I proposed to change our approach for some reasons, hopefully good ones. Have a look and let me know there what you think.

Anyway, merging! Thanks.

seife:
To be honest, I did not know about the possibility to squash PRs and (of course ;-) ) I did no tread the documentation before submitting my patches. Will do so in the future.

PR #249, opened 2020-01-24
Add compatible support for xattr on top of pyxattr

ericzolf:
Clarification based on the MR #247

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ You need to make sure that the following requirements are met:
 * Python 3.5 or higher
 * librsync 1.0.0 or higher
 * pylibacl (optional, to support ACLs)
-* pyxattr (optional, to support extended attributes)
+* pyxattr (optional, to support extended attributes) - **not** the xattr library!
 * SSH for remote operations
 
 Then you can install one of the following packages:
--- a/README.md
+++ b/README.md
@@ -16,8 +16,7 @@ and fastest to restore the most recent backups, combining the space
 advantages of incremental backups while keeping the speed advantages of full
 backups (at least for recent ones).
 
-If the optional dependencies pylibacl and pyxattr are installed,
-rdiff-backup will support
+If the optional dependencies pylibacl and pyxattr are installed, rdiff-backup will support
 [Access Control Lists](https://en.wikipedia.org/wiki/Access-control_list#Filesystem_ACLs)
 and [Extended Attributes](https://en.wikipedia.org/wiki/Extended_file_attributes)
 provided the file system(s) also support these features.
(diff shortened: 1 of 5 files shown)


ericzolf:
Pushed different changes based on discussion in #247

ericzolf:
If I compare my patch with seife 's one:
1. order of import (we still prefer pyxattr over xattr ;-))
2. my version check is for both libraries, not only pyxattr (@seife - can you please check that it isn't an issue for xattr?)
3. I use os.fsencode instead of bytes(..., utf8) - this is on purpose, I let Python do what it thinks is right.

seife:
This got longer than expected, so the executive summary first:

_If both pyxattr and python-xattr are installed, there is no way to import pyxattr (AFAIK)._
---
The problem is, that both python-xattr and pyxattr can be imported with "import xattr", but then behave differently.

This is why I first try to `import xattr.pyxattr_compat as xattr` and if this fails `import xattr`.
I understand that you want to prefer pyxattr over xattr.pyxattr_compat, but I'm not sure how to achieve that. In my tests (virtualenv; pip install pyxattr; pip install xattr, then use something like the following), xattr always "won" if both were present
``` python
import xattr
xattr.setxattr(__file__, b"user.test", b"test val")
xattr.setxattr(__file__, b"user.foo", b"foo val")
la = xattr.listxattr(__file__, False)
print(type(la), la)
for xa in la:
    print("  ", type(xa), xa)
```
When only pyxattr is installed, I get:
```
<class 'list'> [b'user.test', b'user.foo']
   <class 'bytes'> b'user.test'
   <class 'bytes'> b'user.foo'
```
now I `pip install xattr`, then the same code gets
```
<class 'tuple'> ('user.test', 'user.foo')
   <class 'str'> user.test
   <class 'str'> user.foo
```
So I think xattr.pyxattr_compat needs to be tried first.
Or we need to find out if the xattr instance is from pyxattr or from xattr, and if it is from xattr. something like this works for me, but I'm not sure if it is considered ugly:
``` python
import xattr
try:
    foo =  xattr._listxattr # pyxattr does not contain _listxattr but python-xattr does
    import xattr.pyxattr_compat as xattr
except:
    pass
```
The `os.encode` part is then not needed, because AFAICT pyxattr_compat is really compatible with pyxattr and there is no difference in the type. And the version check also needs to be skipped, because `xattr.pyxattr_compat` does not contain `__version__`, so the check always fails and the warning is always printed.

(The pyxattr version check could be removed completely IMHO, because the oldest listed version on pypi is 0.5.1 from 2012. Given we now requires python3.5 or newer, it might be safe to assume that pyxattr is new enough).

ericzolf:
OK, got it, I'll fix it once I find time.

ericzolf:
OK, hopefully fixed now. Can you please check that it works under SuSE before I merge?

seife:
Yes, this works with python3-xattr and python3-pyxattr packages installed.

ericzolf:
Thanks!

PR #253, opened 2020-01-31
Use the git log to populate a minimal changelog

ericzolf:
This is a preparation for the release of v1.9.0b0


ericzolf:
Note to myself: adding `[skip travis]` to a commit doesn't help because it blocks the pipeline instead of skipping it.

PR #254, opened 2020-02-01
Fix some more ugly bytes output in strings using _safe_str

### LINKED ISSUE #238: More binary strings printed.
I've come upon a couple more strings printed with b'

```
ssh: Could not resolve hostname backup: Name or service not known
Fatal Error: Truncated header string (problem probably originated remotely)

Couldn't start up the remote connection by executing

    b'ssh backup rdiff-backup --server'
```

```
ListError: 'run/systemd/journal/streams/9:344377075' [Errno 2]
No such file or directory: b'/run/systemd/journal/streams/9:344377075'
```

btw, am I the only one bothered by the fact that the leading slash is cut off from file paths most of the time? :)

'run/systemd/journal/streams/9:344377075'
vs
'/run/systemd/journal/streams/9:344377075'

ericzolf:
FIX: remove some more ugly bytes output in strings using _safe_str, closes #238

The function _safe_str is for now a local workaround, which I'd like to move to the Log object at a later stage and avoid having this function all over the place.

### DIFF
--- a/src/rdiff-backup-statistics
+++ b/src/rdiff-backup-statistics
@@ -78,13 +78,13 @@ def parse_args():
         os.path.join(os.fsencode(args[0]), b"rdiff-backup-data"),
     )
     if not Globals.rbdir.isdir():
-        sys.exit("Directory %a not found" % (Globals.rbdir.path,))
+        sys.exit("Directory %s not found" % (Globals.rbdir.get_safepath(),))
 
 
 def system(cmd):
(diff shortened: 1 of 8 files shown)


ericzolf:
I need this fix into master to work on #260 - can someone please review/merge it?

ikus060:
I don't like having those `_safe_str` function everywhere, but I don't see a good alternative.
+1

PR #255, opened 2020-02-01
Describe the release process and extend the PR explanations

ericzolf:
As the subject says.

Regarding the release process, the description is pretty straight forward I hope. Regarding the PR process and especially the 3 letters acronym prefix, this allows the following, which makes creating the changelog out of it very easy (which is a pain right now):

```
$ ./tools/get_changelog_since.sh v1.9.0b0
=== Changes ===
DOC: new release rules and procedure added to docs/DEVELOP.md
DOC: modified recommended practices and clarified roles for Pull Requests.
NEW: add a new tool to help generate the changelog (description in DEVELOP.md)
=== Authors ===
Eric L
```

It comes at the price of little constraints in the PR process, and changing from rebasing/push--force to merge/push/squash, which I hope is acceptable for everybody. Else tell me an alternative.


ottok:
I am strongly against using merges to keep up with upstream and amending changes by adding more commits. The merges will make the history ugly and difficult to review. Also adding more and more commits makes the PR difficult to review.

For example this PR has now 10 commits, and there is no way I am going to look at those 10 commits and read their commit messages when clearly their commit messages are not intended to describe the contents.

Instead of:
```
git checkout master
git pull
git checkout PR-BRANCH
git pull
git merge master
# fix potential conflicts
git push
```

rather recommend:
```
git checkout PR-BRANCH
git remote update
git rebase origin/master
git push
```

Making one change and pushing that to a PR makes sense only for explicitly showing the changes since last review to the reviewer if it is the same person. If somebody else reviews it is hard to track anything if the PR consists of 10 commits with random commit messages. Much better that the PR at all times represent the final thing with correct commit messages.

Also, when rebasing and force pushing an updated branch Github is indeed capable of telling that this happened and showing some changes. I don't think it is a good trade-off to try to make once particular re-review scenario between 2 participants optimized with the cost of making all other scenarios (initial review, follow-up reviews a week later when the whole change anyway needs to be read completely to be refreshed to reviewer memory, reviews with multiple persons commenting etc).

ericzolf:
From past experiences, how to handle MR/PR is subject to holy wars if we don't pay attention, so we need to be very precise on what we say or we won't come to an agreement. At the end of the day, we all want to reach a smooth release process. I also want to state that I used the rebase/push --force for a long time before I recently changed my opinion.

Before I start with my arguments, one question: I don't know/find `git remote update` - what is this for a command?

This said, few statements from my side:

- as long as we "squash and merge" PRs, the ugly story doesn't remain in the master branch and this is (IMHO) the most important.
- as documented, getting a clean commit description is the responsibility of the merger helped by the PR author, GitHub helps here by offering a summary of all commits as squash commit message, it's generally only a matter of removing the useless cruft.
- there is no difference between re-basing on master and merging master into one's branch when it comes to conflicts, they need to be resolved in both cases before pushing, and before you can re-base one's branch onto master, you need also to `pull master` i.e. switch branches, etc.
- given that I'm developing in parallel on multiple PCs/laptops, I need to be able to easily exchange branches between them; this doesn't work well (or I missed a trick) if the branches are forced pushed as I need to delete my local branch before I re-checkout it (pulling a forced-push branch onto its local tracking branch is a mess I never solved). _This_ is really the aspect that made me change my preferred approach.
- the same issue applies if I checkout a branch to test/review it locally and want to pull the new version after changes have been made by the author.
- as for the dirtyness of PR history, I don't really see the issue as under "Files changed", GitHub always shows the full change, and you only need to go into the individual commits if you really need to see the history (e.g. during a review). If you want to see the global delta locally, `git diff master..` shows it to you.

But, as said, it really depends on how one works, so let's have the discussion.

ikus060:
At the end, all we want is to improve rdiff-backup and have help as mush as possible from outside. If we start adding more rules for contributions and care about how we get stuff merged into the project, it's not helping. 

The more constraint you add, more reluctant people will be to submit changes and work with. Could we embrace a more liberal way of working ?

ericzolf:
I didn't express it as a constraint, more as a recommendation (my purpose was also to give contributors with less knowledge of Git, a few hints how to go about it). For now, I can suggest instead of recommending, suggesting. Would it be better/acceptable for you? I can also move the whole chapter to a new file DEVELOP_FAQ.md, which would contain such learned lessons in the form of an FAQ.

I actually like my last suggestion, it would make DEVELOP.md leaner and easier to read (I would also move away my notes about debugging). What do you think?

ikus060:
As long as we don'T start refusing PR or contributions only because the branches are not properly merged. It's like we are all working in the same company with one single policy. We all have different working methodologies.

My self, I usually, reorganise my commits using rebase interactive to squash commit and renamed my commit properly. To make it clean and everything. When merging a PR, we always use a merge commit. That's our way of working and it suit our need.

ottok:
* Command `git remote update` is quite handy, it updates your remote branches (like running `git fetch` on everything)
* If the plan is to always squash all commits in every PR, then essentially all PRs will be single commit PRs. Using rebase to refresh PR's (instead of merges) keeps the PR commits clean and allows to merge them nicely to that there can be multiple commits in a PR (when it makes sense and the change is big enough that several commits are justified)
* If the plan is to always squash all commits at merge time, then you put the burden of writing the final commit message on the code reviewer. I'd rather but that burden on the submitter.
* If you locally test something and later re-test, you simply need to discard the local copy git `git branch -d` first so that you get the new rebased version. You will automatically notice there is a need to do so if you pull a branch and git starts to suggest there are conflicts and merges to be done. Alternatively using `git remote update` this problem does not exist that much since you will see your local (old) branch and new remote branch have diverged.
* In the rebase model you can easily compare git diff feature/x...origin/feature/x what are the new changes in the PR since last review.
* Remember though that the local git commands apply only to contributors who have write access to the repository. For those making PR's from their own Github forks it is much easier to review on Github that to run `git remote add` for new remotes and looking at the commits in git.

ottok:
If https://github.com/rdiff-backup/rdiff-backup/pull/255/files#diff-1abac1411cbf3059f2c8f28710daba7fL45-R54 was changed so that nothing is recommended, and contributors can either use rebase or merges? (and most likely those who use rebases will get a quicker review since that is easier to review)

Also, I don't think we should go into details on what git commands to use. Teaching basic git usage is not in the scope of the project development docs.

ericzolf:
OK, I removed the controversial part, we can see how it evolves. Final review please, and merge if acceptable!

ericzolf:
Is it OK to merge now? (as said, I don't like merging my own PRs)

PR #258, opened 2020-02-06
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.
Also use `gmtime` to be independent of timezone.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,9 +88,10 @@ class build_templates(Command):
     def run(self):
         if DEBUG:
             self.debug_print(self.distribution.dump_option_dicts())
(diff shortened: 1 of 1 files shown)


ericzolf:
The PR looks OK but the build pipeline had a hick-up. Can you please push a dummy commit to trigger it again? `git commit --amend && git push --force` on the current branch should be sufficient.

ericzolf:
Thanks. Merged.

PR #262, opened 2020-02-09
Add workaround because Popen doesn't accept byte commands on Windows

### LINKED ISSUE #260: Running latest version on windows errors when specifying a remote directory
image

Expected results:
Connects via supplied SSH command to remote server and performs a backup

What happens:
Errors with an iterable error on Popen. 

Really not sure what I'm doing wrong. Any help here would be great.

Remote server is an RPI running the latest version 1.9.0b as well.
Local backups work fine but soon as I add the :: it fails.

This happens when using --test-server as well

ericzolf:
FIX: avoid int is not iterable error when calling remote command on Windows
Closes #260

I didn't have time to test, if someone finds the time...

### DIFF
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -169,11 +169,19 @@ def init_connection(remote_cmd):
     try:
         # we need buffered read on SSH communications, hence using
         # default value for bufsize parameter
-        process = subprocess.Popen(
-            remote_cmd,
-            shell=True,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE)
+        if os.name == 'nt':
+            # FIXME workaround because python 3.7 doesn't yet accept bytes
+            process = subprocess.Popen(
+                os.fsdecode(remote_cmd),
+                shell=True,
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE)
+        else:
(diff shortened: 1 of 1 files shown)


src/rdiff_backup/SetConnections.py
```diff
@@ -169,11 +169,19 @@ def init_connection(remote_cmd):
     try:
         # we need buffered read on SSH communications, hence using
         # default value for bufsize parameter
-        process = subprocess.Popen(
-            remote_cmd,
-            shell=True,
-            stdin=subprocess.PIPE,
-            stdout=subprocess.PIPE)
+        if os.name == 'nt':
```
ikus060: Just to avoid repetition of code:
```
remote_cmd = os.fsdecode(remote_cmd) if os.name == 'nt' else remote_cmd
process = subprocess.Popen(
...
```
Or something similar.

ericzolf: the reason for the duplication is that the string is used in further commands down the line and hence I didn't want to simply overwrite its byte version with its string version. I could have worked with a temporary variable but that wouldn't have made the code nicer IMHO.

ericzolf:
@ikus060 after my explanations and geekbozu's confirmation in #260 that it solves his issue, can you approve the change?

ikus060:
done

Issue #252, opened 2020-01-30
Debian package version in 1.9.0b0 still shows version 1.4.0b1


ericzolf:
This was discovered in PR #234 and I have a suspicion that it is due to a mix of pip and apt installed packages, but no time for now to fix it properly. There is a workaround anyway.

ericzolf:
@ottok please have a look at the latest beta release: the version of the Debian packages is wrong and still 1.4.0b1.

ottok:
Indeed the source package version is wrong as is the binary package versions inside:
```
Format: 1.8
Date: Fri, 02 Aug 2019 21:01:40 +0100
Source: rdiff-backup
Binary: rdiff-backup
Architecture: source amd64
Version: 1.4.0b1
Distribution: unstable
Urgency: medium
Maintainer: Otto Kekäläinen <otto@debian.org>
Changed-By: Otto Kekäläinen <otto@debian.org>
Description:
 rdiff-backup - remote incremental backup
Changes:
 rdiff-backup (1.4.0b1) unstable; urgency=medium
 .
   * Initial changelog entry for native rdiff-backup packaging.
Checksums-Sha1:
 ff419f824ba663a1a1b7c627102287068dfbbe64 844 rdiff-backup_1.4.0b1.dsc
 9d98db41c9c2224dda4ab8f09cf9001efae03f96 243028 rdiff-backup_1.4.0b1.tar.xz
 a1cceb459d7486c6754d5118df50451e5f829aaf 28496 rdiff-backup-dbgsym_1.4.0b1_amd64.ddeb
 e260100c7d25cfa9e37d804c6908403484a0b203 7096 rdiff-backup_1.4.0b1_amd64.buildinfo
 a98f5ea08f1f4379c312103a18ac2e32f2228206 191432 rdiff-backup_1.4.0b1_amd64.deb
Checksums-Sha256:
 2863a6f7cb7841615d26ac644658283b5cb5abb58f439870a432ace7e63f9e24 844 rdiff-backup_1.4.0b1.dsc
 d4febc1cb55a2f27eeb4e4acbd85a7f7e9f2fc112b8353984e4f79337df31cf6 243028 rdiff-backup_1.4.0b1.tar.xz
 a8179cd3cfeebf102ac34edf732fadaad77ee7e3751e0cc8edb2cda528efe596 28496 rdiff-backup-dbgsym_1.4.0b1_amd64.ddeb
[...]

PR #265, opened 2020-02-11
Control Debian version via with helper script (Closes: #252)

ottok:
Turned out the dynamic version control in debian/rules did not work always,
and even when it worked it affected only binary packages, not the main
source package.

With this wrapper the debian/changelog is updated before build and both
source and binary packages end up with correct version string inherited
from python_scm.


ottok:
Working build in https://travis-ci.org/rdiff-backup/rdiff-backup/jobs/648810963

image

ericzolf:
I don't really understand why it works better this way, but I trust you that it does. Merging...

Issue #266, opened 2020-02-11
Fail to restore file: '<' not supported between instances


ikus060:
I'm starting to test the latest version of rdiff-backup with rdiffweb. I have a series of test that play hard with encoding and special character to make sure everything is well supported.

As a start, I have problem to restore file containing '<' in their name. rdiff-backup is failing with an exception.

Here the step to reproduce.
```
$ wget https://git.patrikdufresne.com/pdsl/rdiffweb/raw/master/rdiffweb/tests/testcases.tar.gz
$ tar -zxvf testcases.tar.gz
$ rdiff-backup --version
rdiff-backup 1.9.0b0
$ rdiff-backup --restore-as-of=1414921853 '/testcases/Fichier @ <root>'
Fatal Error: Wrong number of arguments given.
See the rdiff-backup manual page for more information.
root@b94a5e7b3814:/# rdiff-backup --restore-as-of=1414921853 '/testcases/Fichier @ <root>' /tmp
Warning: special_escapes file not found,
will assume need to escape DOS devices and trailing spaces based on file systems.
Exception ''<' not supported between instances of 'QuotedRPath' and 'QuotedRPath'' raised of class '<class 'TypeError'>':
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/Main.py", line 390, in error_check_Main
    Main(arglist)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/Main.py", line 412, in Main
    take_action(rps)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/Main.py", line 368, in take_action
    Restore(rps[0], rps[1], 1)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/Main.py", line 713, in Restore
    restore_root.new_index(restore_index), inc_rpath, dest_rp, time)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/restore.py", line 39, in Restore
    TargetS.patch(target, diff_iter)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/restore.py", line 331, in patch
    for diff in rorpiter.FillInIter(diff_iter, target):
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/rorpiter.py", line 191, in FillInIter
    first_rp = next(rpiter)
  File "/usr/local/lib/python3.7/dist-packages/rdiff_backup/restore.py", line 279, in get_diffs_from_collated
[...]

PR #273, opened 2020-02-16
fix failed restore quoting

ericzolf:
The issue described in #266 doesn't appear anymore (@ikus060 you might want to try it), but the  surrounding code looks suspicious to me and I'd like to take the chance to fix/verify a few more things, and possibly add a few more tests to our suite.

### DIFF
--- a/src/rdiff_backup/FilenameMapping.py
+++ b/src/rdiff_backup/FilenameMapping.py
@@ -191,8 +191,6 @@ class QuotedRPath(rpath.RPath):
                 self.inc_timestr = unquote(temp_rp.inc_timestr)
         else:
             result = rpath.RPath.isincfile(self)
-            if result:
-                self.inc_basestr = unquote(self.inc_basestr)
         return result
 
     def get_path(self):
--- a/src/rdiff_backup/restore.py
+++ b/src/rdiff_backup/restore.py
@@ -28,7 +28,10 @@ class RestoreError(Exception):
 
 
 def Restore(mirror_rp, inc_rpath, target, restore_to_time):
-    """Recursively restore mirror and inc_rpath to target at rest_time"""
+    """Recursively restore mirror and inc_rpath to target at restore_to_time
+    in epoch format"""
+
+    # Store references to classes over the connection
     MirrorS = mirror_rp.conn.restore.MirrorStruct
     TargetS = target.conn.restore.TargetStruct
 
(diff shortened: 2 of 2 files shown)


ericzolf:
I wouldn't say that I like very much the code but it looks OK so it's ready for merging.

src/rdiff_backup/restore.py
```diff
@@ -441,6 +444,11 @@ def __init__(self, mirror_rp, inc_rp, inc_list):
         self.inc_rp, self.inc_list = inc_rp, inc_list
         self.set_relevant_incs()
 
+    def __str__(self):
```
ikus060: That was used for debugging ?

src/rdiff_backup/FilenameMapping.py
```diff
@@ -191,8 +191,6 @@ def isincfile(self):
                 self.inc_timestr = unquote(temp_rp.inc_timestr)
         else:
             result = rpath.RPath.isincfile(self)
-            if result:
-                self.inc_basestr = unquote(self.inc_basestr)
```
ikus060: I feel suspicious about this change. Why are you skipping the unquote here ? Is inc_basestr already quoted ?
Cause I took a look in git blame and those lines was not changed for a while.

ikus060:
Is it possible to write a unit test to cover this bug ?

src/rdiff_backup/restore.py
ericzolf: Yes, it's quite handy in the debugger, you just call `p str(rf_object)` and you get an overview of the object's content. It's nowhere used in the code.

src/rdiff_backup/FilenameMapping.py
ericzolf: So was I, that's why I did put first the PR in WiP, so I can check. Everything seems legit, I guess that I changed enough of the code to create a side effect. Basically this branch uses `self.index` as basis (which is unquoted and doesn't need to be unquoted again, hence the existence of `self.quoted_index`), whereas the other branch of the same function (if index is empty) uses the `self.path` variable, which is quoted.

This is this inconsistency that I don't like in the code, and why I'd like to get rid of QuotedRPath altogether; IMHO it should be transparent. But I don't want to change it right now.

ericzolf: Oh, and to be clear: `inc_basestr` is set as a side effect of `isincfile`, and because it is created on the basis of `index`, it is already unquoted and shouldn't be unquoted a 2nd time.

ericzolf:
> Is it possible to write a unit test to cover this bug ?

Yes, it is possible but I ran out of time, I'm travelling right now and will be then in holiday, so I didn't want to block anything. If it's OK for you, I created #275 to keep track of this, but I don't know if I'll find time to address it for the up-coming release.

PR #276, opened 2020-02-22
Remove Debian packages from deploy to GitHub to avoid confusion

### LINKED ISSUE #263: Debian package cannot be installed on debian:sid
Either we should document more or get our debian pacakge fixed.
I'm trying to use the debian package as part of an automation and it's failing as follow.

```
$ docker run --rm -it debian:sid
# apt update
...
# apt install wget
...
# apt install ./rdiff-backup_1.4.0b1_amd64.deb 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'rdiff-backup' instead of './rdiff-backup_1.4.0b1_amd64.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 rdiff-backup : Depends: python3 (< 3.7) but 3.7.5-3 is to be installed
                Depends: librsync1 (>= 0.9.6) but it is not installable
                Recommends: python3-pylibacl but it is not going to be installed
                Recommends: python3-pyxattr
E: Unable to correct problems, you have held broken packages.
```
@ottok Which version of Debian is supported ?

ericzolf:
FIX: remove too specific Debian packages from GitHub deployment, closes #263
The Debian target has been renamed so that we can keep the generic target
pattern for future distributable targets ({one letter}dist{whatever}).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -170,10 +170,10 @@ jobs:
         - docker run --rm -e PLAT=$PLAT -v `pwd`:/io $DOCKER_IMAGE $PRE_CMD /io/tools/build_wheels.sh
         - ls dist/
 
-    #--- Build and deploy Debian packages ---
+    #--- Build Debian packages ---
     - os: linux
       language: shell
-      env: MAKE_STEP=bdist_deb RUN_COMMAND=
+      env: MAKE_STEP=dist_deb RUN_COMMAND=
       addons:
         apt:
           packages:  # make sure these match debian/control contents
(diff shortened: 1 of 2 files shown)


ericzolf:
And please squash and merge the PR once you've reviewed and approved it, and make sure the Squash commit message looks according to guidelines.

If it looks ok to you but you would like someone else to specifically review it, then comment so.

.travis.yml
```diff
@@ -173,7 +173,7 @@ jobs:
     #--- Build and deploy Debian packages ---
```
ikus060: We should review this comments: Build and deploy Debian packages, since we are not deploying anymore. ;)

ericzolf: Fixed :-)

PR #284, opened 2020-02-24
FIX: do not fail when starting with uid/gid equal to maximum

ericzolf:
Closes https://savannah.nongnu.org/bugs/?31624
A "random" uid/gid was generated by adding 1 to the current uid/gid
which would fail if one of them is equal to the maximum possible
(might apparently happen under MacOS), now half of the id is taken
and one added (to avoid 0).


ikus060:
Do you want to merge this for 2.0.0 ?
This change is ok for me.

ericzolf:
Yes, it's just a few characters and one issue less.

Issue #295, opened 2020-03-08
particular long cyrillic dirname breaks backup


legolegs:
I caught an issue on my data. A directory with particular long but correct name being backed up breaks backup.

I have tried both `rdiff-backup 1.9.2rc0` and 1.9.1b0.
System: Fedora 29 x64, filesystem - both ext4 and tmpfs tested, locale ru_RU.UTF-8.

How to reproduce:
```bash
mkdir source backup
touch source/file
rdiff-backup source/ backup/ #fist
mkdir source/'ггггггггг1ггггггггг2ггггггггг3ггггггггг4ггггггггг5ггггггггг6ггггггггг7ггггггггг8ггггггггг9.ггггггггг0._'
rdiff-backup source/ backup/ #second
rdiff-backup --list-increment-sizes backup/
```
Got this:
```
Exception ''ascii' codec can't decode byte 0xd0 in position 1: ordinal not in range(128)' raised of class '<class 'UnicodeDecodeError'>':
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/Main.py", line 390, in error_check_Main
    Main(arglist)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/Main.py", line 412, in Main
    take_action(rps)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/Main.py", line 362, in take_action
    ListIncrementSizes(rps[0])
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/Main.py", line 917, in ListIncrementSizes
    print(manage.ListIncrementSizes(restore_root, restore_index))
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/manage.py", line 220, in ListIncrementSizes
    time_dict = get_time_dict(get_inc_select())
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/manage.py", line 167, in get_time_dict
    if not inc.isincfile():
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 1494, in isincfile
    inc_info = get_incfile_info(basename)
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/rpath.py", line 474, in get_incfile_info
    if Time.bytestotime(timestring) is None:
  File "/usr/lib64/python3.7/site-packages/rdiff_backup/Time.py", line 123, in bytestotime
    return stringtotime(timebytes.decode('ascii'))

Traceback (most recent call last):
[...]

PR #297, opened 2020-03-10
bytestotime() should return None on decode failure (issue #295)

zjw:
The function may be called with arbitrary non-ascii byte strings.  In such
cases it should return None without throwing a UnicodeDecodeError exception.

### DIFF
--- a/src/rdiff_backup/Time.py
+++ b/src/rdiff_backup/Time.py
@@ -120,7 +120,10 @@ def stringtotime(timestring):
 
 
 def bytestotime(timebytes):
-    return stringtotime(timebytes.decode('ascii'))
+    try:
+        return stringtotime(timebytes.decode('ascii'))
+    except UnicodeDecodeError:
+        return None
 
 
 def timetopretty(timeinseconds):
--- a/testing/timetest.py
+++ b/testing/timetest.py
@@ -35,6 +35,15 @@ class TimeTest(unittest.TestCase):
         assert cmp("2001-09-01T12:00:00-08:00",
                    "2001-09-01T12:00:00-07:00") == 1
 
+    def testBytestotime(self):
+        """Test converting byte string to time"""
+        timesec = int(time.time())
+        assert timesec == int(Time.bytestotime(Time.timetostring(timesec).encode('ascii')))
+
+        # assure that non-ascii byte strings return None and that they don't
+        # throw an exception (issue #295)
+        assert Time.bytestotime(b'\xff') is None
+
     def testStringtotime(self):
         """Test converting string to time"""
         timesec = int(time.time())


zjw:
I wouldn't say that it is swallowing the error.  The error is not being ignored.  The error is generating a specific return value.

So `bytestotime()` calls `stringtotime(timebytes.decode('ascii'))` (the method that immediately precedes it in the file).  The docstring for `stringtotime()` states:
```
    """Return time in seconds from w3 timestring

    If there is an error parsing the string, or it doesn't look
    like a w3 datetime string, return None.
    """
```
Then `stringtotime()` proceeds with:
```
try:
    ...
    return int(utc_in_secs) + tzdtoseconds(timestring[19:])
except (TypeError, ValueError, AssertionError):
    return None
```
So `stringtotime()` is also guilty of swallowing the error (as is the case for numerous other methods within the codebase).  In my opinion, the exception handling done in `stringtotime()` is acceptable.  I personally think that it is acceptable in `bytestotime()` as well, although I admit that it might be good to add a similar docstring which explains the exception handling.

Would better commenting suffice?  Otherwise, elaborate further on what you are looking for.

By the way, `bytestotime()` is new since V1.2.8, so issue #295 is a regression.

ikus060:
@zjw Had a deep look into the code and must agree. You changes it's probably the best to follow the current logic in place. I might be accustom to different coding practice in regards to error handling.

ikus060:
The only comments I have left is would be nice to have a unittest to reproduce this scenarios to avoid any regression in the future. If not now, we should create a task to remember the scenario to be added as a test.

zjw:
I have added a unit test for bytestotime().

ericzolf:
I'd like to understand how a non-ASCII string comes into the function, and which one? I have the impression there is another underlying issue.

zjw:
Well, the traceback log provided in issue #295 is a pretty good roadmap of how it occurs.  That log was provided for this command:  `rdiff-backup --list-increment-sizes backup/`.  The program proceeds to recurse through every file and directory found under backup/rdiff-backup-data/increments*, including subdirectories, tallying up increment sizes and and increment times, then printing a list giving the total size for each different increment time found.

The tallying occurs in [`get_time_dict(inc_iter)`](https://github.com/rdiff-backup/rdiff-backup/blob/2b6941e7febae497a8eb40f5b665528b51aebd7b/src/rdiff_backup/manage.py#L163).  For each file name, the name is tested to see if it is an increment (`isincfile()`), and if it is, the increment time is extracted from the file name and the increment size is tallied.

Not all file names recursed through are regular files; instead it is a mix of increment file names and subdirectory names.  The `isincfile()` function tests whether a given name is an increment by looking for the usual naming pattern used for increment files.  That pattern, of course, includes a time/date string for the increment.  So every name is broken apart at the periods, and the text that is found between the 2 periods where the date would be expected to be found is then passed through `bytestotime()` to see if it actually is a valid time string.  That test, along with other tests, determines whether the file is truly an increment file.  So text from all names that have the requisite periods in them is passed through `bytestotime()` as a test to see if it is an increment file.  Therefore there is no guarantee that `bytestotime()` will only be called with ASCII text.

There may be better ways for the program to do what I described above --  I won't argue that.  But that's they way it has been done, and it has worked up until now.  Changing it would take some work and some testing.

ericzolf:
Thanks for the explanations, you don't need to apologize for the quality of the code, it's what it is and it grew over time (but that's why I think it's important to clean it up step after step).

Issue #296, opened 2020-03-09
Overlapping destination and source directory broken


heikkiorsila:

heikkiorsila:
I'm testing "ppa:rdiff-backup/rdiff-backup-development" repository to backup a following setup:

/home/foo/tmp: All the files (source directory)
/home/foo/tmp/files: contains files to backup
/home/foo/tmp/backups: contains the backups (destination directory)

Note: the destination directory overlaps with the source directory.

I'm running:  rdiff-backup --exclude /home/foo/tmp/backups /home/foo/tmp /home/foo/tmp/backups

I'm using --exclude /home/foo/tmp/backups to avoid infinite regress.

The error I get is:

$ rdiff-backup --exclude /home/foo/tmp/backups /home/foo/tmp /home/foo/tmp/backups
Exception ''generator' object has no attribute 'Select'' raised of class '<class 'AttributeError'>':
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 390, in error_check_Main
    Main(arglist)
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 412, in Main
    take_action(rps)
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 348, in take_action
    Backup(rps[0], rps[1])
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 430, in Backup
    backup_warn_if_infinite_regress(rpin, rpout)
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 588, in backup_warn_if_infinite_regress
    if not Globals.select_mirror.Select(relative_rpout):

Traceback (most recent call last):
  File "/usr/bin/rdiff-backup", line 32, in <module>
    rdiff_backup.Main.error_check_Main(sys.argv[1:])
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 390, in error_check_Main
    Main(arglist)
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 412, in Main
    take_action(rps)
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 348, in take_action
    Backup(rps[0], rps[1])
  File "/usr/lib/python3/dist-packages/rdiff_backup/Main.py", line 430, in Backup
    backup_warn_if_infinite_regress(rpin, rpout)
[...]

PR #299, opened 2020-03-12
Fix overlap source target 296

ericzolf:
So shortly before release, I only dared to come with a minimal invasive workaround, I'll fix it more correctly for the next release.

The underlying issue is that select_mirror can't contain a selection object because it isn't pickable, but the function Select can't be called on a selection iterable, so it's a bit of a catch 22.

I also added a test case to the selection tests.

FIX: add workaround to avoid error when backup directory is under the source directory (see issue #296), there is a warning but the backup can succeed.
DEV: add test to backup to a repo within the source directory

### DIFF
--- a/src/rdiff_backup/Main.py
+++ b/src/rdiff_backup/Main.py
@@ -584,15 +584,18 @@ def backup_warn_if_infinite_regress(rpin, rpout):
         return
 
     relative_rpout_comps = tuple(rpout.path[len(rpin.path) + 1:].split(b'/'))
-    relative_rpout = rpin.new_index(relative_rpout_comps)
-    if not Globals.select_mirror.Select(relative_rpout):
-        return
+    relative_rpout = rpin.new_index(relative_rpout_comps)  # noqa: F841
+    # FIXME: this fails currently because the selection object isn't stored
+    #        but an iterable, the object not being pickable.
+    #        Related to issue #296
+    #if not Globals.select_mirror.Select(relative_rpout):  # noqa: E265
+    #    return
 
     Log(
         """Warning: The destination directory '%s' may be contained in the
 source directory '%s'.  This could cause an infinite regress.  You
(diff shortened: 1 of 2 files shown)


ericzolf:
Could someone review and squash merge? After that I could release...

ottok:
Do you want to review this @heikkiorsila?

I skimmed through the code but I didn't quite understand it. I guess @heikkiorsila could at least test it?

src/rdiff_backup/Main.py
```diff
@@ -584,15 +584,18 @@ def backup_warn_if_infinite_regress(rpin, rpout):
...
 
     relative_rpout_comps = tuple(rpout.path[len(rpin.path) + 1:].split(b'/'))
-    relative_rpout = rpin.new_index(relative_rpout_comps)
-    if not Globals.select_mirror.Select(relative_rpout):
-        return
+    relative_rpout = rpin.new_index(relative_rpout_comps)  # noqa: F841
+    # FIXME: this fails currently because the selection object isn't stored
+    #        but an iterable, the object not being pickable.
+    #        Related to issue #296
+    #if not Globals.select_mirror.Select(relative_rpout):  # noqa: E265
```
ikus060: Tell me if I'm wrong: 
This was checking if the `relative_rpout` is part of the selection. If it was, it doesn't print the error messages.

ikus060:
For me the changes it good.

heikkiorsila:
I tested the change in my local setup. It seems to run correctly for our use case :) Thanks!

src/rdiff_backup/Main.py
ericzolf: More precisely, it was checking if Select was excluding it from the selection.

ericzolf:
I'd like to merge and release, but it can wait until tomorrow.

ericzolf:
Uuups. Forget it, sometimes the page doesn't refresh properly in GitHub.

PR #315, opened 2020-04-04
NEW: add support for Python 3.9 (alpha release).

ericzolf:
Even if Python 3.9 is still an alpha version, we should start testing it to avoid issues


ericzolf:
OK, it seems too early for Travis, so I close it for now and we can re-open it once it gets better. Just for the record: on my Fedora 31 it works already without issue, so adding support shouldn't be a big deal once it's released.

Issue #316, opened 2020-04-04
Unable to install on Raspbian (librsync.h/librsync.c)


Tahutipai:
Manual install on Ubuntu via pip3: no problem.  But I am unable to do the same on a raspberry pi.

``python-librsync`` is installed, but it appears to be looking for the C source code for librsync, and I am not sure how to install that.

Truncated log:
```
>sudo pip3 install rdiff-backup
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting rdiff-backup
  Using cached rdiff-backup-2.0.0.tar.gz (408 kB)
Building wheels for collected packages: rdiff-backup
  ERROR: Command errored out with exit status 1:
   arm-linux-gnueabihf-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.7m -c src/_librsyncmodule.c -o build/temp.linux-armv7l-3.7/src/_librsyncmodule.o
  src/_librsyncmodule.c:25:10: fatal error: librsync.h: No such file or directory
   #include <librsync.h>
            ^~~~~~~~~~~~
```
How does one install the required C librsync library on Raspbian?

--

Also, just to avoid traps, for beginners, under installation the line:
``curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py sudo python3 get-pip.py sudo pip3 install rdiff-backup``
Should perhaps read:
``curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && sudo python3 get-pip.py && sudo pip3 install rdiff-backup``

Tahutipai:
Ah figured it out, one just needs to run ``sudo apt update; sudo apt install librsync-dev``

Perhaps the installation document could be updated to reflect this? As it appears having the C source for librsync is a requirement.

ericzolf:
It depends, only because it's ARM and we don't provide (pre-compiled) wheels for ARM but only for x86 and x86_64. And the formatting is also wrong, as you noticed yourself. We'll keep this issue opened until we've fixed the documentation. Thanks for reporting.

Tahutipai:
Thanks Eric.  

Would it be useful if I format a clean RaPi, and write up the procedure to install rdiff-backup on vanilla Raspbian? (for beginners)

[...]

PR #318, opened 2020-04-05
Update installations steps

ikus060:
Following #316, I've update the installations step to fix the formatting and to add a notice about non-x86 platform.


ericzolf:
Would you mind using 3 backticks for each code snippet, so that it remains consistent?

ikus060:
> code snippet, so that it remains consistent?

check again

PR #329, opened 2020-04-21
Adding missing `software-properties-common`

ikus060:
Add command line to install `software-properties-common` for ubuntu backports installation


ottok:
Indeed, this might be needed in some installs, e.g. in a container (but on a desktop system it that package is pre-installed). It makes the instructions a bit longer but I guess it does not hurt, and running the command in "vain" does no harm.

PR #331, opened 2020-04-25
Clarify how tagged commits are released to GitHub and PyPI

### LINKED ISSUE #326: Deploy pre-release versions on PyPi
To make it easier for a casual bug reporter to check out if their bug is fixed on git master and otherwise to install/test/debug latest development version and thus contribute and save our time in testing, it would be good to publish pre-release versions on PyPi.org.

We [did that](https://github.com/rdiff-backup/rdiff-backup/issues/58#issuecomment-576016726) prior to the 2.0.0 release, but not after. I think we should.

I did it manually now, but it could be in Travis-CI.

image

Steps to do it manually:
```
git checkout feature/pypi-integration
git rebase -i master
make container
make bdist_wheel
python3 -m twine upload --verbose dist/rdiff-backup-2.0.1.dev9+g1e4f236.tar.gz
```

Unfortunately this fails right now due to the `+g1e4f236` local component in the version number.

Related:
- https://packaging.python.org/guides/distributing-packages-using-setuptools/#id68
- https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions

ericzolf:
DEV: clarify version tag pattern and their influence on releases, closes #326

I've taken also the chance to yank a few releases on PyPI, don't be surprised.

### DIFF
--- a/docs/DEVELOP.md
+++ b/docs/DEVELOP.md
@@ -99,6 +99,26 @@ In versioning we utilize git tags as understood by
 Version strings follow the [PEP-440
 standard](https://www.python.org/dev/peps/pep-0440/).
 
+The rules are currently as follows (check the `.travis.yml` file for details):
+
+- all commits tagged with an underscore at the end or with a tag looking like a
+  version number (i.e. as in next two bullets) are released to
+  [GitHub](https://github.com/rdiff-backup/rdiff-backup/releases/).
+- all commits tagged with alpha, beta, rc or final format are released to
+  [PyPI](https://pypi.org/project/rdiff-backup/#history), i.e. the ones looking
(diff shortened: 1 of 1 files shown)


ottok:
Using https://test.pypi.org/project/rdiff-backup/#history is just for testing PyPi, it is not the correct place to publish pre-releases. Correct location to release pre-releases is the real PyPi, from where people can install pre-release using normaly PyPi with `--dev` flag as documented in our README.md.

ericzolf:
You are surely rights, so .dev version are for testing the deployment, and alpha versions are meant if you want to release something.

Getting the Travis thing working and stable was hard work and I'm not going to change it unless absolutely needed.

Issue #327, opened 2020-04-20
[ENH] "--check-destination-dir" should check if multiple metadata with same date exists


ikus060:
In reference to issue #322, we should try to provide a better analysis of the command "--check-destination-dir" to do more validation on the repository and ensure the integrity of the repo really valid.

We should validate if repository contains multiple metadata with the same date.

@ericzolf  I'm reading your discussion on issue #322. You are free to close this issue if it doesn't make any sens. ;) I'm just raising the idea

PR #332, opened 2020-04-26
Improve the PR #328 based on feedback from zjw

ericzolf:
- re-write the tests to take into consideration that --remove-older-than doesn't need --allow-duplicate-timestamps
- improve the error message and the man-page also accordingly
- make the tests also slightly more robust, sorting entry and using date instead of number of backups

### DIFF
--- a/docs/rdiff-backup.1
+++ b/docs/rdiff-backup.1
@@ -64,9 +64,13 @@ out the examples.html file included in the rdiff-backup distribution.
 .TP
 .B \-\-allow-duplicate-timestamps
 This option is only to be used if you encounter the issue of metadata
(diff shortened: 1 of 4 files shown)


ericzolf:
@zjw the promised philosophical discussion: if a user sees the error message, his last backup will have failed and hence he should regress it, which will fail without the `--allow-duplicate-timestamps` flag. It technically works to first delete the old duplicate increments and then to regress, but that doesn't sound right and isn't what we want our users to get used to.

And as @ikus060 suggested in #327 the future should be that `--check-destination-dir` detects (and fixes!) such issues. For now, the user needs to do it themself by calling first `--check-destination-dir --allow-duplicate-timestamps` and then `--remove-older-than`.

Issue #337, opened 2020-04-29
[BUG] Since v2.0.0 cannot backup from Windows network drive


xfournet:
## Bug summary

It's no more possible to backup from a Windows shared network drive (e.g. \\NAS\Work) with v2.0.0. It was working with v1.3.3

## Version, Python, Operating System

Using Windows build available on GitHub
```
2020-04-30 00:38:03.526578 +0200  <CLIENT-5532>  Using rdiff-backup version 2.0.0
2020-04-30 00:38:03.526578 +0200  <CLIENT-5532>         with cpython D:\scoop\apps\rdiff-backup\current\rdiff-backup.exe version 3.7.5
2020-04-30 00:38:03.527582 +0200  <CLIENT-5532>         on Windows-10-10.0.18362, fs encoding utf-8
2020-04-30 00:38:03.527582 +0200  <CLIENT-5532>  Fatal Error: No arguments given
See the rdiff-backup manual page for more information.
```

## rdiff-backup call

How did you call rdiff-backup, with which parameter:

```
rdiff-backup \\NAS\Work backup
```

## What happened and what did you expect?

It output an error message
````
Fatal Error: Source directory /NAS/Work does not exist
````
however the directory exists on the network share and the same command is working with the previous rdiff-backup version (1.3.3)

Tried with //NAS/Work and various argument quoting, this doesn't help and it give the same error

## More information

````
2020-04-30 00:47:36.758150 +0200  <CLIENT-6244>  Using rdiff-backup version 2.0.0
2020-04-30 00:47:36.758150 +0200  <CLIENT-6244>         with cpython D:\scoop\apps\rdiff-backup\current\rdiff-backup.exe version 3.7.5
2020-04-30 00:47:36.760123 +0200  <CLIENT-6244>         on Windows-10-10.0.18362, fs encoding utf-8
2020-04-30 00:47:36.771155 +0200  <CLIENT-6244>  Fatal Error: Source directory /NAS/Work does not exist
````

ericzolf:
Sorry, this is one is on me, I know that I fiddled quite a lot with the parsing of paths until I got it working under Windows in all combinations I could think of, but I definitely didn't think about direct use of Windows shares (i.e. without assigning a drive letter).

As I don't have a CIFS NAS to test from, I'm right a bit unsure how to test this... Perhaps I can simulate somehow but I don't know.

[...]

PR #339, opened 2020-05-01
Fix and improve automated installation of Windows dev VM

ericzolf:
DEV: much better automated installation of Windows development VM via Vagrant/Ansible
- use standard VM image from a known Ansible developer
- increase required memory to 8GB to avoid issues with VS installation
- add more robust Vagrant parameters to use Windows with Ansible
- development directory is dependent on the user's name through variable
- sort better and enable previously broken Chocolatey packages
- install certifi to avoid SSL errors and setuptools-scm
- shortly explain how to use the Ansible playbooks to use a non-Vagrant Windows

Allows me to address #337 much more easily, so a quick review and merge would be much appreciated.

### DIFF
--- a/tools/windows/README.md
+++ b/tools/windows/README.md
@@ -2,12 +2,41 @@
 
 ## Create the Windows VM
 
-Starting from https://github.com/redhat-cop/automate-windows/tree/master/vagrant-libvirt-image, create a Windows VM
-usable by Ansible (any other alternative approach to a such VM is of course valid).
+A simple `vagrant up` should do, using a default Windows image, and you'll get
+in the best case a fully workable rdiff-backup development environment on Windows.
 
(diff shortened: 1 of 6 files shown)


ericzolf:
You are more than welcome to do so but it can't break anything because it was broken before :roll_eyes: 

It seems to work much better but I have the impression that there is still one issue during the provisioning step, but restarting the VM and restarting the `vagrant provision` step helps it. I don't need to tweak any more directly in the VM.

ericzolf:
@ikus060 if you could be lenient and fast on this patch, I would appreciate: it doesn't influence the code of rdiff-backup, only my development environment, and currently I need to toggle branches to be able to develop in good conditions under Windows.

ikus060:
Yes, sure. I approve

PR #354, opened 2020-05-10
Add specific instruction how to install rdiff-backup on CentOS6

ikus060:


README.md
```diff
@@ -90,6 +90,22 @@ sudo pip3 install rdiff-backup
...
 
+### CentOS and RHEL 6 (from Pypi)
+
+```
+yum install centos-release-scl
+yum install rh-python36 gcc libacl-devel
+ln -s /opt/rh/rh-python36/root/usr/bin/python /usr/bin/python3
+python3 --version
+
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
ericzolf, line 101: According to https://www.softwarecollections.org/en/scls/rhscl/rh-python36/ I am the opinion that pip is installed once you've installed the collection, and that get-pip shouldn't be required. Do you confirm the need?

README.md
```diff
@@ -90,6 +90,22 @@ sudo pip3 install rdiff-backup
 > **NOTE:** If your platform is not i386 or amd64, e.g. ARM/MIPS/etc,
   you may need the build dependencies `build-essentials`, `librsync-dev`.
 
+### CentOS and RHEL 6 (from Pypi)
+
+```
+yum install centos-release-scl
+yum install rh-python36 gcc libacl-devel
+ln -s /opt/rh/rh-python36/root/usr/bin/python /usr/bin/python3
```
ericzolf, line 98: Shouldn't it be `scl enable rh-python36 bash`?

README.md
```diff
@@ -90,6 +90,22 @@ sudo pip3 install rdiff-backup
...
+python3 --version
+
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+python3 get-pip.py
+ln -s /opt/rh/rh-python36/root/usr/bin/pip /usr/bin/pip3
+
+pip install rdiff-backup pyxattr pylibacl
+ln -sf /opt/rh/rh-python36/root/usr/bin/rdiff-backup /usr/bin/rdiff-backup2
```
ericzolf, line 106: Why this? To install in parallel to rdiff-backup 1.x?

README.md
ericzolf: Things seem a bit broken right now, but http://mirror.centos.org/centos/6/sclo/x86_64/rh/Packages/r/ conains pip for Python 3.5 and 3.6.

ericzolf:
OK, I couldn't resist and this works for me, as root:

```
# yum install centos-release-scl
# yum install rh-python36 gcc libacl-devel
# python --version
Python 2.6.6
# scl enable rh-python36 bash
> python --version
Python 3.6.9
> pip install rdiff-backup pyxattr pylibacl
> which rdiff-backup
/opt/rh/rh-python36/root/usr/bin/rdiff-backup
> rdiff-backup --version
rdiff-backup 2.0.0
> exit
# echo 'exec scl enable rh-python36 -- rdiff-backup "$@"' > /usr/bin/rdiff-backup  # or rdiff-backup-2 !
# chmod +x /usr/bin/rdiff-backup
# which rdiff-backup
/usr/bin/rdiff-backup
# rdiff-backup --version
rdiff-backup 2.0.0
```

This was in a fresh CentOS 6 container, so no other rdiff-backup installed. Auto-completion and man-pages don't work, but who cares? The advantage of the approach is that it's supportable by Red Hat, whereas the soft-linking all over the place isn't (in case someone cares about SCL Python support). A quick test with a normal user showed that it works as well, and it should work also over SSH.

ikus060:
@ericzolf 
I really recommend against "scl enable rh-python36 bash" as it replace the default "python" command by python v3.6. The impact on the system might be greater than what we expect. This is why I recommended to create a symbolic link instead.

frankcrawford:
@ikus060 I disagree, at least in the way @ericzolf is using it here.  It is only to setup rdiff-backup and the related libs.
The shell script then created looks to be a way to limit the use of scl to just the running of rdiff-backup.

ikus060:
My mistake I've read further about "scl enabled" and it's not behaving the way I expected. Will fix the instruction.

ikus060:
@ericzolf I've made some changes. Anything else required ?

ericzolf:
I just saw that PyPI has the wrong typography but else I approved.

ikus060:
I've update the PyPi typographie ;)

ericzolf:
The pipeline is failing. You need to review/merge #360 and then rebase your PR upon master.

PR #359, opened 2020-05-14
Handle selection files with Windows/DOS endings

### LINKED ISSUE #357: No longer parsing exclude-globbing-filelist
## Directories and files are being included on Windows.

Today I replaced v1.2.8 with v2.0.0 and I discover that the following command line, which has always worked, is now trying to include directories listed in exclude-list.

`rdiff-backup -v9 --no-acls --exclude-globbing-filelist exclude-list "c:/Users/Jon Hunt/" "g:/Backup"`

The content of the file are as follows.

`c:/Users/Jon Hunt/NTUSER.DAT
c:/Users/Jon Hunt/.*
c:/Users/Jon Hunt/Dropbox
c:/Users/Jon Hunt/AppData
c:/Users/Jon Hunt/MicrosoftEdgeBackups`

However, straight away the program attempts to backup the "AppData" directory.

Call `rdiff-backup -v9` and paste the output between the backquotes below, repeat for each environment impacted:

```
2020-05-12 14:37:00.990779 +0100  <CLIENT-11916>  Using rdiff-backup version 2.0.0
2020-05-12 14:37:00.991774 +0100  <CLIENT-11916>        with cpython C:\WINDOWS\rdiff-backup.exe version 3.7.5
2020-05-12 14:37:00.993768 +0100  <CLIENT-11916>        on Windows-10-10.0.18362, fs encoding utf-8
2020-05-12 14:37:00.993768 +0100  <CLIENT-11916>  Fatal Error: No arguments given
```

ericzolf:
FIX: handle properly include/exclude files with Windows/DOS endings, closes #357
Also added a test to check for this case.

### DIFF
--- a/src/rdiff_backup/selection.py
+++ b/src/rdiff_backup/selection.py
@@ -402,6 +402,7 @@ probably isn't what you meant.""" % (self.selection_functions[-1].name, ))
         something_excluded, tuple_list = None, []
         separator = Globals.null_separator and b"\0" or b"\n"
         for line in filelist_fp.read().split(separator):
+            line = line.rstrip(b'\r')  # for Windows/DOS endings
             if not line:
                 continue  # skip blanks
             try:
@@ -478,6 +479,7 @@ probably isn't what you meant.""" % (self.selection_functions[-1].name, ))
         log.Log("Reading globbing filelist %s" % list_name, 4)
         separator = Globals.null_separator and b"\0" or b"\n"
         for line in filelist_fp.read().split(separator):
+            line = line.rstrip(b'\r')  # for Windows/DOS endings
             if not line:
                 continue  # skip blanks
             if line[:2] == b"+ ":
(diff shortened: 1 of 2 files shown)


src/rdiff_backup/selection.py
```diff
@@ -402,6 +402,7 @@ def incr_warnings(exc):
         something_excluded, tuple_list = None, []
         separator = Globals.null_separator and b"\0" or b"\n"
         for line in filelist_fp.read().split(separator):
```
ikus060, line 404: Quickly, I'm wondering why we are not using `readlines()` or even `for line in f` for that mather.

ikus060: I see it to support the `null_separator`. Was not aware of this feature. Since when can we have carriage return in a filename !? If that tru, we should probably enabled this feature by default.

Any how, This is good for me.

ericzolf: The problem is that null characters are not that easy to enter and filenames with carriage return aren't that often.

PR #361, opened 2020-05-16
Add mkv and img files to default no compression list

heisath:
Adds mkv (video/audio/movie) and img (drive images) to no compression list.

These are either compressed themself or probably big / hard to compress.


ericzolf:
I'm ok with mkv, I'm not OK with img,it can be anything, even an uncompressed file image, which can profit _a lot_ from compression.

heisath:
Yeah you're right. Shall I push a revert commit to this PR or make a new one?

Issue #282, opened 2020-02-24
add more media filetypes to no_compression_regexp_string


ericzolf:
From https://savannah.nongnu.org/bugs/index.php?27841

There are a few relatively common multimedia file extensions that are missing from no_compression_regexp_string in `src/rdiff_backup/Globals.py`:

ogv
oga
ogm
mkv

Having these in there would make the same kind of sense as "avi" does.

PR #362, opened 2020-05-16
Update no_compression_regexp to include more files

heisath:
This adds the file extensions for multimedia files mentioned in #282 to Globals.py.
Specifically ogv, oga, ogm and mkv.


ericzolf:
Looks good to me, but long term, we need to find a better way, we can't add all possible extensions to the code.

Issue #377, opened 2020-05-25
[BUG] Invalid Python download link in the Windows-DEVELOP.md file.


JirkaV:
The verbal description suggests downloading Python 3.7.5 (32-bit) while the download link points to Python 3.7.3 (64-bit).

I'm pointing this out mostly because the documentation specifically says "Python 3.7.3 is not supported".

ericzolf:
Thanks. Would you mind providing a pull request for this fix?

JirkaV:
No problem - wasn't sure on the versions but used the docs text as a guidance.

PR #383, opened 2020-05-28
Fix for #377 - incorrect Python download link in docs.

JirkaV:
Fix for link in documentation - kept the target Python ad 3.7.5 32-bit to be consistent with the text


ericzolf:
Looks good to me, than you!

Issue #384, opened 2020-05-28
[BUG] Crash on network drive due to handling too specific exceptions


JirkaV:
## Bug summary

I have a home NAS (WD MyCloud) and Windows 10 PC connected to it via regular network sharing. When I try to backup to that destination, rdiff-backup crashes. I'll submit a pull request shortly.

## Version, Python, Operating System

Call `rdiff-backup -v9` and paste the output between the backquotes below, repeat for each environment impacted:

```
```

## rdiff-backup call

How did you call rdiff-backup, with which parameter:

```
D:\Fotky\Export\pro_Pavla>python "C:\Program Files (x86)\Python37-32\Scripts\rdiff-backup" --force -v5 --no-acls --exclude "C:/Lightroom Catalog/Lightroom Catalog Previews.lrdata" --exclude "*Smart*" --exclude "**.lrprev" --exclude "**.lrfprev" --exclude "**.dng" "C:/Lightroom Catalog/" "X:/Lightroom Catalog Backup Test/"
Using rdiff-backup version 2.0.3
        with cpython C:\Program Files (x86)\Python37-32\python.exe version 3.7.7
        on Windows-10-10.0.18362-SP0, fs encoding utf-8
Hardlinks disabled by default on Windows
Unable to import module (py)xattr.
Extended attributes not supported on filesystem at C:/Lightroom Catalog
POSIX ACLs test skipped. rdiff-backup run with --no-acls option.
Windows ACLs test skipped. rdiff-backup run with --no-acls option.
-----------------------------------------------------------------
Detected abilities for source (read only) file system:
  Access control lists                         Off
  Extended attributes                          Off
  Windows access control lists                 Off
  Case sensitivity                             Off
  Escape DOS devices                           On
  Escape trailing spaces                       Off
  Mac OS X style resource forks                Off
  Mac OS X Finder information                  Off
-----------------------------------------------------------------
Directories on file system at X:/Lightroom Catalog Backup Test/rdiff-backup-data/rdiff-backup.tmp.0 are not fsyncable.
Assuming it's unnecessary.
Unable to import module (py)xattr.
Extended attributes not supported on filesystem at X:/Lightroom Catalog Backup Test/rdiff-backup-data/rdiff-backup.tmp.0
POSIX ACLs test skipped. rdiff-backup run with --no-acls option.
Windows ACLs test skipped. rdiff-backup run with --no-acls option.
[...]

PR #385, opened 2020-05-28
Fix for #384

JirkaV:


ericzolf:
Can you please re-create a new PR based on a clean master? Else it's quite confusing, and the risk for merging errors grows.

Regarding the fix itself, it's very coarse, I'm not convinced that we can accept it. The failure in the tests seems genuine to me, meaning that your change caused it.

ottok:
@JirkaV Please rebase on latest master so we can review and merge this. Thanks!

ericzolf:
As I already wrote above, the patch isn't acceptable in its current state, we're still working on identifying the root cause and find a better patch. No need to refresh now.

ottok:
Submitter never followed up on this, so closing now.

Thanks for your steps towards contributing. Feel free to take a new shot at solving this at a later time in a new clean PR.

Issue #374, opened 2020-05-23
[BUG] Get rid of Python Deprecation Warning PY_SSIZE_T_CLEAN will be required for '#' formats


ericzolf:
## Bug summary

With Python 3.8 new deprecation warnings regarding need of `PY_SSIZE_T_CLEAN` for '#' formats appeared.

## Version, Python, Operating System

```
$ ./setup.py --version
2.0.4.dev3+ga2c73cf.d20200523
```

Important is Python 3.8+

## rdiff-backup call

```
$ PATH=$PWD/build/scripts-3.8:$PATH PYTHONPATH=$PWD/build/lib.linux-x86_64-3.8 python testing/librsynctest.py 
/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:153: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  self.maker = _librsync.new_deltamaker(sig_string)
/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:90: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  self.eof, len_inbuf_read, cycle_out = self.maker.cycle(self.inbuf)
../home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:90: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  self.eof, len_inbuf_read, cycle_out = self.maker.cycle(self.inbuf)
/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:90: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  self.eof, len_inbuf_read, cycle_out = self.maker.cycle(self.inbuf)
/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:90: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  self.eof, len_inbuf_read, cycle_out = self.maker.cycle(self.inbuf)
/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.8/rdiff_backup/librsync.py:90: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
[...]

PR #390, opened 2020-06-04
FIX PY_SSIZE_T Depreciation Warning

neha-shah99:
**Addresses issue #374**

Fixed the depreciation warnings requiring PY_SSIZE_T_CLEAN for '#' formats, given while using Python 3.8

Now, librsynctest.py gives no depreciation warnings when used with Python 3.8.

### DIFF
--- a/src/_librsyncmodule.c
+++ b/src/_librsyncmodule.c
@@ -20,7 +20,7 @@
  *   02110-1301, USA
  *
  * ----------------------------------------------------------------------- */
-
+#define PY_SSIZE_T_CLEAN
 #include <Python.h>
 #include <librsync.h>
 #define RS_JOB_BLOCKSIZE 65536
@@ -50,7 +50,7 @@ static PyObject*
 _librsync_new_sigmaker(PyObject* self, PyObject* args)
 {
   _librsync_SigMakerObject* sm;
-  long blocklen;
+  Py_ssize_t blocklen;
 
   if (!PyArg_ParseTuple(args, "l:new_sigmaker", &blocklen))
 	return NULL;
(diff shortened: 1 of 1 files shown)


ericzolf:
Thanks, it looks good to me but I'm no C specialist. Just one question: is it OK to cast a signed type `Py_ssize_t` to an unsigned type `size_t`? With my limited knowledge, I would fear that it can go wrong.
https://www.python.org/dev/peps/pep-0353/

neha-shah99:
https://www.python.org/dev/peps/pep-0353/ does say that using `size_t` could give errors in a language where negative indices are used. 

However, I took a look at https://librsync.github.io/structrs__buffers__s.html#details, and `avail_in` uses type `size_t` to store the length of the input. Length isn't a signed value and in C it is illegal to access objects outside of the array bounds, so I did not alter the code casting it to `size_t.` I'll look into the matter more since I've mostly been taught to prefer unsigned types except in certain cases.

If it feels risky to cast `Py_ssize_t` to unsigned type `size_t,` please do let me know, I will change it immediately.

dbaarda:
FTR, in C ssize_t exists as a version of size_t that functions can return
using negative values to indicate errors.

It is only risky to cast an ssize_t value to size_t if there is a
possibility the value is negative (usually because of an error). Casting a
small negative value will turn it into a very large positive value, which
will likely cause a crash when something thinks it's got nearly 16 exabytes
of data to handle.

So, if something returns an ssize_t value, it can possibly return a
negative value when there's an error, so you should check for errors before
casting it to an size_t.

ericzolf:
Thanks for the explanation. Then we are fine because those signed length are only used in the function [PyArg_ParseTuple](https://docs.python.org/3.8/c-api/arg.html#c.PyArg_ParseTuple), which does _not_ use the length parameter to declare errors, but return code, which we check, and exception.

One last thing to @neha-shah99 which I'll fix myself now, but for the next time: your commit message needs to contain one line with 3 capital letters prefix, check the developer/contributor documentation for details. Else, thanks for jumping in, good job!

neha-shah99:
Thanks for letting me work on this and for being patient and helpful with me! It's the first time I've worked on fixing code in Open Source. I'll make sure to include the prefix in the commit message in the future.

Thanks a lot!

PR #404, opened 2020-06-16
improve handling of file description

### LINKED ISSUE #395: [BUG] Backslash in end of path on Windows
## Bug summary
rdiff-backup crashes if there is a backslash at the end of a folder name on Windows. This issue is an inconvenience since when you autocomplete a folder name in Windows Command Prompt it will include the backslash.

```
2020-06-07 15:01:17.983568 +0200  <CLIENT-13984>  Using rdiff-backup version 2.0.3
2020-06-07 15:01:17.983568 +0200  <CLIENT-13984>        with cpython C:\rdiff-backup.exe version 3.7.5
2020-06-07 15:01:17.984569 +0200  <CLIENT-13984>        on Windows-10-10.0.18362, fs encoding utf-8
2020-06-07 15:01:17.984569 +0200  <CLIENT-13984>  Fatal Error: No arguments given
```

## rdiff-backup call
```
>rdiff-backup.exe a\ b\
Exception 'Unexpected end to file description b'a\\'' raised of class '<class 'rdiff_backup.SetConnections.SetConnectionsException'>':
  File "rdiff_backup\Main.py", line 393, in error_check_Main
  File "rdiff_backup\Main.py", line 410, in Main
  File "rdiff_backup\SetConnections.py", line 66, in get_cmd_pairs
  File "rdiff_backup\SetConnections.py", line 128, in parse_file_desc
  File "rdiff_backup\SetConnections.py", line 113, in check_len

  File "rdiff-backup", line 32, in <module>
  File "rdiff_backup\Main.py", line 393, in error_check_Main
  File "rdiff_backup\Main.py", line 410, in Main
  File "rdiff_backup\SetConnections.py", line 66, in get_cmd_pairs
  File "rdiff_backup\SetConnections.py", line 128, in parse_file_desc
  File "rdiff_backup\SetConnections.py", line 113, in check_len
rdiff_backup.SetConnections.SetConnectionsException: Unexpected end to file description b'a\\'
[14104] Failed to execute script rdiff-backup
```

It should backup as normal as it's a valid path on Windows.

ericzolf:
FIX: avoid issue with backslash at the end of file path under Windows, closes #395
The function parse_file_desc was simplified by using a proper regexp instead
of parsing the line manually. The setconnection tests were also improved to
use assertEqual function.

### DIFF
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -25,6 +25,7 @@ the related connections.
 """
 
 import os
+import re
 import sys
 import subprocess
 from .log import Log
@@ -107,43 +108,34 @@ def parse_file_desc(file_desc):
 
     """
 
-    def check_len(i):
-        if i >= len(file_desc):
-            raise SetConnectionsException(
-                "Unexpected end to file description %s" % file_desc)
+    # paths and similar objects must always be bytes
+    file_desc = os.fsencode(file_desc)
+    # match double colon not preceded by an odd number of backslashes
+    file_match = re.fullmatch(rb"^(?P<host>.*[^\\](?:\\\\)*)::(?P<path>.*)$",
+                              file_desc)
+    if file_match:
+        file_host = file_match.group("host")
+        # According to description, the backslashes must be unquoted, i.e.
+        # double backslashes replaced by single ones, and single ones removed.
+        # Hence we split along double ones, remove single ones in each element,
+        # and join back with a single backslash.
+        file_host = b'\\'.join(
+            [x.replace(b'\\',b'') for x in re.split(rb'\\\\', file_host) if x])
+        file_path = file_match.group("path")
+    else:
+        if re.match(rb"^::", file_desc):
+            raise SetConnectionsException("No file host in '%s'" % file_desc)
+        file_host = None
+        file_path = file_desc
 
-    host_info_list, i, last_was_quoted = [], 0, None
-    file_desc = os.fsencode(
-        file_desc)  # paths and similar must always be bytes
-    while 1:
-        if i == len(file_desc):
-            # make sure paths under Windows use / instead of \
-            if os.path.altsep:  # only Windows has an alternative separator for paths
-                file_desc = file_desc.replace(os.fsencode(os.path.sep), b'/')
-            return (None, file_desc)
-
-        if file_desc[i] == ord(
-                '\\'):  # byte[n] is the numerical value hence ord
-            i = i + 1
-            check_len(i)
-            last_was_quoted = 1
-        elif (file_desc[i] == ord(":") and i > 0
-              and file_desc[i - 1] == ord(":") and not last_was_quoted):
-            host_info_list.pop()  # Remove last colon from name
-            break
-        else:
-            last_was_quoted = None
-        host_info_list.append(file_desc[i:i + 1])
-        i = i + 1
-
-    check_len(i + 1)
-
-    filename = file_desc[i + 1:]
     # make sure paths under Windows use / instead of \
     if os.path.altsep:  # only Windows has an alternative separator for paths
-        filename = filename.replace(os.fsencode(os.path.sep), b'/')
+        file_path = file_path.replace(os.fsencode(os.path.sep), b'/')
 
-    return (b"".join(host_info_list), filename)
+    if not file_path:
+        raise SetConnectionsException("No file path in '%s'" % file_desc)
+
+    return (file_host, file_path)
 
 
 def fill_schema(host_info):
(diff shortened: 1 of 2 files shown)


src/rdiff_backup/SetConnections.py
```diff
@@ -107,43 +108,34 @@ def parse_file_desc(file_desc):
...
-        else:
-            last_was_quoted = None
-        host_info_list.append(file_desc[i:i + 1])
-        i = i + 1
-
-    check_len(i + 1)
+    # paths and similar objects must always be bytes
+    file_desc = os.fsencode(file_desc)
+    # match double colon not preceded by an odd number of backslashes
+    file_match = re.fullmatch(rb"^(?P<host>.*[^\\](?:\\\\)*)::(?P<path>.*)$",
```
ikus060, line 114: For a first time, I'm not following you on this. Regex is not well suited to do this kind of parsing. It add complexity to the code without resolving the actual problem.

ericzolf: It resolved the actual problem because trailing backslashes are now possible under Windows. It was never the purpose of this function to remove trailing backslashes, it happens later in the code (rpath.normalize), but the previous code in the current function was choking on such trailing backslashes, which it isn't any more.

Regarding usage of regex:
1. did you have a look at the previous code and did you understand it (I didn't really, that's why I rewrote it)? I think that regexp makes the code much clearer than it used to be.
2. even the python documentation says it: https://docs.python.org/3/library/re.html#simulating-scanf - pattern matching is the way to go, unless you want to write your own parser (which is what the previous code did).
3. the code/regex could be _a lot_ cleaner/simpler but the man page in the REMOTE OPERATION section, in the 3rd paragraph, starting with "Concerning quoting", states quite a lot of complex quoting rules, which makes the regex more complex than it could be.

OK, let me ask on the mailing list if someone is using the extended quoting capabilities, then the code could become much simpler.

ericzolf: As per the discussion on the mailing list, we need the current complexity to potentially support IPv6 and its colons. I'm happy to address them if you have further concerns, but I need more concrete questions and feedback.

ikus060: Let keep the regex then.

HRio:
This fix broke safekeep local tests. The test does the following
```
nice -n10 ionice -c3 -t rdiff-backup --exclude **/*.nobackup --include /tmp/tmpsdsr755l/client --exclude-regexp /.* ::/ /tmp/tmpsdsr755l/server
```

HRio:
@frankcrawford heads-up

ericzolf:
Looks to me actually like a correct error message. What was `::/` supposed to express?

HRio:
> Looks to me actually like a correct error message. What was `::/` supposed to express?

Not sure but this is a local test case, I think it comes from this safekeep code
```python
    userhost = ''
    if cfg['host']:
        userhost = '%s@%s' % (cfg['user'], cfg['host'])
    args.extend([userhost + '::' + bdir, cfg['dir']])
```

and that there is no `cfg['host']` simply defined, in this test case.

@frankcrawford what do you say?

frankcrawford:
I'll have to have a look and see what is happening.

frankcrawford:
@HRio What actual test are you running?  Are you running "make check"?
@ericzolf I think it is for a run on the localhost, if it is the test I think, and it is correct syntax according to the man page:

`rdiff-backup      [options]      [[[user@]host1.foo]::source_directory] [[[user@]host2.foo]::destination_directory]`

and that does imply that you can miss `user@host1.foo`.

HRio:
> @HRio What actual test are you running? Are you running "make check"?

I am running:
```
./safekeep-test --local
```

ericzolf:
I would say that the man page is wrong and not the code: nothing in the rest of the man page hints at the host being optional (and neither in the code to be clear), and the double columns isn't mandatory, as the man page would hint. Furthermore, what is actually the meaning of `::/some/dir`, localhost via SSH, or simply the same as `/some/dir`? Long story short, I would fix the man page to `rdiff-backup [options] [[[user@]host1.foo::]source_directory] [[[user@]host2.foo::]destination_directory]` (shifting the double-columns into the optional part). Fair enough?

frankcrawford:
@ericzolf I wouldn't disagree with you, although it is actually a change in how it is processed, as it used to work, up to 5.0.3 and is only generating an error now.

The fix for safekeep is pretty simple, but will it affect others any other users?

ericzolf:
It's already released as part of v2.0.5 so it's anyway too late to correct, and reverting would rather be worse (if it made at all any sense).

frankcrawford:
@ericzolf yep, not disagreeing, there isn't much that can be done at this point.

PR #407, opened 2020-06-23
Fix RHEL CentOS install docs

ericzolf:
DOC: added clearer instructions for installing weak dependencies to support ACLs and EAs under CentOS and RHEL


ericzolf:
@frankcrawford probably something for you to review

frankcrawford:
Will do.

ikus060:
I've tested the new instruction for CentOS 8 in docker and it's working like a charm. I see the libacl and pyxattr getting installted to.,
image

PR #411, opened 2020-07-02
Add sample script to create file system as user

ericzolf:
As written in the header of the script, it's more a proof of concept than a real script, but it's still helpful to do some tests.


ikus060:
Merge as you see fit since it's a "misc" script

PR #414, opened 2020-07-03
Replace deprecated xattr functions

### LINKED ISSUE #177: Replace deprecated xattr function and expect pyxattr >= 0.4
With pyxattr version 0.4 the functions `xattr.XXXxattr` (list, get, set...) were replaced by `xattr.XXX`. As there is a bug in pyxattr 0.2, it makes sense to replace the functions and make 0.4 (from 2008!) the minimum acceptable version in the code, dependencies and documentation (searching for xattr in the repo gives all relevant places).

ericzolf:
DEV: replace deprecated xattr.<verb>xattr with xattr.<verb> function, closes #177

I didn't even document that we need at least version 0.4 as this version is already 12 years old (it's not even any more available from PyPI).

### DIFF
--- a/src/rdiff_backup/eas_acls.py
+++ b/src/rdiff_backup/eas_acls.py
@@ -67,7 +67,7 @@ class ExtendedAttributes:
     def read_from_rp(self, rp):
         """Set the extended attributes from an rpath"""
         try:
-            attr_list = rp.conn.xattr.listxattr(rp.path, rp.issym())
+            attr_list = rp.conn.xattr.list(rp.path, rp.issym())
         except IOError as exc:
             if exc.errno in (errno.EOPNOTSUPP, errno.EPERM, errno.ETXTBSY):
                 return  # if not supported, consider empty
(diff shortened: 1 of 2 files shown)


src/rdiff_backup/eas_acls.py
```diff
@@ -101,9 +101,9 @@ def read_from_rp(self, rp):
     def clear_rp(self, rp):
         """Delete all the extended attributes in rpath"""
         try:
-            for name in rp.conn.xattr.listxattr(rp.path, rp.issym()):
+            for name in rp.conn.xattr.list(rp.path, rp.issym()):
                 try:
-                    rp.conn.xattr.removexattr(rp.path, name, rp.issym())
```
ikus060, line 106: hum, if we are doing this. We are breaking inter operability of the network protocole again ?

ericzolf: Very valid remark, that's why I just created #416 and tested this MR against it, and didn't notice any issue.

This said, as xattr isn't an rdiff-backup module but an external one, calling it, even remotely, shouldn't make any difference from a network protocol perspective (only if e.g. `xattr.get` wouldn't exist on the other side, but with 12 years to go, we are on the very safe side).

PR #417, opened 2020-07-05
Add python3-setuptools as a run time dependency so --version works

ottok:
Closes: #305.


ericzolf:
Looks good to me. One of the jobs failed but it was only because Travis CI had a network hick-up, a re-start helped.

ericzolf:
Please don't forget to add FIX/DEV/CHG/NEW/DOC prefix in your commit message!

PR #421, opened 2020-07-08
Add setuptools as runtime dependency to tox and setup.py

### LINKED ISSUE #305: rdiff-backup --version from Ubuntu PPA show DEV
Want to report a problem. I didn't check what wrong with the build.
```
docker run -it --rm ubuntu
apt update && apt install software-properties-common
add-apt-repository ppa:rdiff-backup/rdiff-backup-development
apt update
apt install rdiff-backup
rdiff-backup --version
```
Shows `rdiff-backup DEV`

ericzolf:
Also add a note in the developer documentation.
CHG: setuptools is a runtime dependency for installation and tests so that version appears correctly instead of DEV, closes #305

This is obviously a follow-up of the corresponding discussion regarding Debian packages, it applies also to any kind of packaging, even wheels.

### DIFF
--- a/tox_dist.ini
+++ b/tox_dist.ini
@@ -15,6 +15,7 @@ envlist = py35, py36, py37, py38
 passenv = RDIFF_TEST_* RDIFF_BACKUP_*
 skip_install = true
 deps =
+    setuptools-scm
     pyxattr
     pylibacl
     wheel
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ python:
 deploy:
   - provider: releases
     file_glob: true
-    file: dist/*.*
+    file: dist/rdiff*.*
     skip_cleanup: true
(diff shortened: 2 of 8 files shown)


tox_dist.ini
```diff
@@ -15,6 +15,7 @@ envlist = py35, py36, py37, py38
 passenv = RDIFF_TEST_* RDIFF_BACKUP_*
 skip_install = true
 deps =
+    setuptools-scm
```
ikus060, line 18: Any reason why `setuptools-scm` vs `setuptools` ?

ericzolf: Not sure it was required, but it's because the runtime of this fox isn't rdiff-backup but the wheel packaging of it.

ericzolf:
I allow myself to merge so that I can publish a new version during the week-end.

PR #423, opened 2020-07-08
Fix semi-broken nongnu.org links

andreaso:
The problem with these https://www.nongnu.org/rdiff-backup/... links
is that they redirect to the top level https://rdiff-backup.net/ page,
rather than to the corresponding subpage.


ericzolf:
Thanks for the pull request, well spotted.

PR #430, opened 2020-07-18
Add few development notes about profiling rdiff-backup

ericzolf:
It can be merged before or after the bug-fix release, it doesn't really matter.


ericzolf:
I wanted to avoid doing changes to the code, but, you're right, it might come handy in time. I've added your hints as well and split profiling with from without code changes.

PR #431, opened 2020-07-24
Make pipeline faster by joining small jobs together

ericzolf:
Because each job means a new VM to be started, it takes much more time
than putting the smaller jobs into one single VM. And because build and
static tests don't influence the slow tox job, it's also OK.


ericzolf:
The pipeline took 12min instead of 14 to 16min (mostly 15 to 16min), so I would consider the improvement worth the change.

ikus060:
Good for me

PR #435, opened 2020-07-26
Add faq regarding regression and make local copy of tool

ericzolf:
Kudos to Dominic for https://www.timedicer.co.uk/programs/help/rdiff-backup-regress.sh.php


ikus060:
+1

PR #437, opened 2020-07-26
Add requirements.txt to facilitate security

### LINKED ISSUE #434: [ENH] add requirements.txt to facilitate discovery by GitHub dependency graph

The dependency graph of Github doesn't show us all or dependencies, which means we wouldn't be warned if one got compromised.

Document all our dependencies as described under https://docs.github.com/en/github/visualizing-repository-data-with-graphs/about-the-dependency-graph in a `requirements.txt` so that it's discovered by the dependency graph. Point the developer docs to it in order to avoid duplicate information.

Use `pipfile.lock` but it looks too early, too complex, without clear advantage.

## Additional context

A way to define a dependency to librsync would be nice but I don't see any option.

ericzolf:
DEV: add requirements.txt to help GitHub detect our dependencies and warn about security flaws, closes #434

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,25 @@
+# This file describes the (development) dependencies so that we can be warned
+# by GitHub dependency graph if something is compromised, see:
+# https://docs.github.com/en/github/visualizing-repository-data-with-graphs/about-the-dependency-graph
+
+# We don't pin versions unless we see a reason for it. Having developers with
+# slightly different versions of the dependencies increases our chance to
+# detect non working combinations.
+
+# You can also use the file to install your environment with the following command:
+#	pip install -r requirements.txt
+
+# mandatory
+setuptools
+setuptools-scm
+importlib-metadata ~= 1.0 ; python_version < "3.8"
+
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+setuptools-scm
+importlib-metadata ~= 1.0 ; python_version < "3.8"
+
+# optional
+pylibacl
+pyxattr
+# or xattr under SuSE Linux & Co
+
+# purely for development and testing purposes
+tox
```
ikus060, line 23: Developement dependencies should not be listed here.

ericzolf: Why not? The file isn't of any use beside telling us if there is some security flaw in our dependencies, and I think it's good to know also about flaws in our development tooling.

ericzolf:
Last forced push was to resolve the merge conflict.

requirements.txt
ericzolf: I consider the point closed by lack of response, so that I can merge myself.

Issue #347, opened 2020-05-03
[ENH] make as many tox tests work under windows as possible


ericzolf:
## Is your feature request related to a problem? Please describe.

We don't have a test suite under Windows and this has led to regressions in the move from 1.2/1.3 to 2.0 and might lead again to regressions. Not all the tests can work, but most should work.

## Describe the solution you'd like

I've started a branch https://github.com/rdiff-backup/rdiff-backup/tree/ericzolf-make-tox-work-under-win but it requires more work and I fear _much_ more work, so if someone wants to take over, be my guest!

ericzolf:
The approach is rather simple, even if perhaps tedious:

- create your development environment under Windows is the first step (obviously). There is help under tools/windows even if you just read through the Ansible playbooks to understand what's being done.
- then copy tox.ini to tox_win.ini and adapt to the Windows needs (already started in the branch)
- then adapt step by step each test in the "commands_pre" and in the "command" section until the whole file works.

My suggestion would be to first comment out all tests and then uncomment them step by step, making a Pull Request for each few steps in order to avoid having one big PR at the end with a major review, while still having value to each PR. Feel free to create a "draft/WIP" PR if you need help, it makes review and discussion around the code rather easy.

t9t:
So far I'm stumbling into some issues where some bytes APIs don't work on Windows and/or some APIs don't accept bytes, etc. 

For example in `commontest.py` `os.getenvb` is used, which is not available on Windows, see: 
- https://docs.python.org/3/library/os.html?highlight=os#os.getenvb
- https://www.geeksforgeeks.org/python-os-getenvb-method/

So now I'm working my way through just changing some things one by one to try and get more and more working.

But in parallel (as not a very experienced Python developer) I would like to enquire why bytes types are used rather than strings, so that I won't go changing them all to strings in places where it's not appropriate?

t9t:
I have addressed a few baby steps in #444. I can continue to address more and more, but as a next step I will try to fix up the `tox_win.ini` from your branch in a way that all tests are commented-out by default except the ones that run fine, and then the others can be addressed on by one.

Unfortunately I was unable to set up vagrant properly on my Linux machine (Ubuntu 18.04 LTS) and I don't want to go through all the effort of setting up all the Ansible stuff on my Windows environment (on which Docker won't run anyway, because it's in a VirtualBox VM). So I cannot verify anything using the automated tests on Windows so will continue to keep doing it manually using `tox` in my Python env.

[...]

Issue #443, opened 2020-08-03
[BUG] Windows build error: _librsyncmodule.obj : error LNK2001: unresolved external symbol rs_...


t9t:
## Bug summary

When building on Windows following [these instructions](https://github.com/rdiff-backup/rdiff-backup/blob/master/docs/Windows-DEVELOP.md), I get the error `_librsyncmodule.obj : error LNK2001: unresolved external symbol rs_...` (for multiple symbols).

## Version, Python, Operating System

rdiff-backup `master` as of right now (`f858f01c917854e3bce1741297c50a69e4d958e0`), Python 3.7.8, Windows 10 (2004).

## More information

As promised, I want to try and see if I can help out with getting some of the test working on/for Windows (#347). So to start I'm trying to build the app op Windows.

I'm trying to get it working on my Windows machine using [the instructions on this page](https://github.com/rdiff-backup/rdiff-backup/blob/master/docs/Windows-DEVELOP.md) and after having installed all the dependencies and building librsync, I'm unfortunately stuck on a build issue when running `python setup.py build`. I have some Python experience (mainly webapps), but no experience integrating with C libraries, so I'm kind of stuck here unsure how to continue.

I have Python and the VC++ build tools installed and built librsync (version 2.2.1 as indicated on the aforementioned page). The files in `c:\librsync` are:
```
include/librsync.h
include/librsync_export.h
lib/rsync.lib
share/man/man1/rdiff.1
share/man/man3/librsync.3
```

I would expect a DLL file there somewhere, but then again I'm not sure.

When I try to build rdiff-backup, I get the following error:

```
> python setup.py build
running build
running build_py
running build_templates
running build_ext
building 'rdiff_backup._librsync' extension
[...]

PR #444, opened 2020-08-03
Better cross-platform compatibility in test files

t9t:
This is a (small) step towards #347, to ensure tests run on Windows. I have found a pretty good way to run the tests both on Linux and Windows (manually, yet) which allows me to identify areas which need some more work, address them, then check if tests still run on both.

- `os.getenvb` is not supported on Windows
- `rm -rf` obviously also isn't, unfortunately `shutil.rmtree` only removes directories so an additional check to delete files was necessary
- Again, obviously no `/dev/urandom` on Windows

### DIFF
--- a/testing/commontest.py
+++ b/testing/commontest.py
@@ -40,7 +40,11 @@ def Myrm(dirstring):
     for rp in selection.Select(root_rp).set_iter():
         if rp.isdir():
             rp.chmod(0o700)  # otherwise may not be able to remove
-    assert not os.system(b"rm -rf %s" % (root_rp.path, ))
+    path = root_rp.path
+    if os.path.isdir(path):
+        shutil.rmtree(path)
+    elif os.path.isfile(path):
+        os.remove(path)
 
 
 def re_init_rpath_dir(rp, uid=-1, gid=-1):
--- a/testing/commontest.py
+++ b/testing/commontest.py
@@ -13,9 +13,9 @@ from rdiff_backup import Globals, Hardlink, SetConnections, Main, \
 RBBin = os.fsencode(shutil.which("rdiff-backup"))
 
 # Working directory is defined by Tox, venv or the current build directory
-abs_work_dir = os.getenvb(
-    b'TOX_ENV_DIR',
-    os.getenvb(b'VIRTUAL_ENV', os.path.join(os.getcwdb(), b'build')))
+abs_work_dir = os.getenv(
+    'TOX_ENV_DIR',
+    os.getenv('VIRTUAL_ENV', os.path.join(os.getcwdb(), b'build'))).encode()
 abs_test_dir = os.path.join(abs_work_dir, b'testfiles')
 abs_output_dir = os.path.join(abs_test_dir, b'output')
 abs_restore_dir = os.path.join(abs_test_dir, b'restore')
--- a/testing/commontest.py
+++ b/testing/commontest.py
@@ -13,9 +13,9 @@ from rdiff_backup import Globals, Hardlink, SetConnections, Main, \
 RBBin = os.fsencode(shutil.which("rdiff-backup"))
 
 # Working directory is defined by Tox, venv or the current build directory
-abs_work_dir = os.getenv(
+abs_work_dir = os.fsencode(os.getenv(
     'TOX_ENV_DIR',
-    os.getenv('VIRTUAL_ENV', os.path.join(os.getcwdb(), b'build'))).encode()
+    os.getenv('VIRTUAL_ENV', os.path.join(os.getcwd(), 'build'))))
 abs_test_dir = os.path.join(abs_work_dir, b'testfiles')
 abs_output_dir = os.path.join(abs_test_dir, b'output')
 abs_restore_dir = os.path.join(abs_test_dir, b'restore')
--- a/testing/librsynctest.py
+++ b/testing/librsynctest.py
@@ -10,11 +10,8 @@ def MakeRandomFile(path, length=None):
     """Writes a random file of given length, or random len if unspecified"""
     if not length:
         length = random.randrange(5000, 100000)
-    fp = open(path, "wb")
-    try:
+    with open(path, "wb") as fp:
         fp.write(os.urandom(length))
-    finally:
-        fp.close()
 
 
 class LibrsyncTest(unittest.TestCase):
--- a/testing/librsynctest.py
+++ b/testing/librsynctest.py
@@ -11,12 +11,10 @@ def MakeRandomFile(path, length=None):
     if not length:
         length = random.randrange(5000, 100000)
     fp = open(path, "wb")
-    fp_random = open('/dev/urandom', 'rb')
-
-    fp.write(fp_random.read(length))
-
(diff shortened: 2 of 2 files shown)


testing/librsynctest.py
```diff
@@ -11,12 +11,10 @@ def MakeRandomFile(path, length=None):
     if not length:
         length = random.randrange(5000, 100000)
     fp = open(path, "wb")
```
ericzolf, line 13: This is already a lot cleaner, to make it even cleaner, you could write something like:
```
with open(...) as fp:
    fp.write(...)
```
then you don't even need to explicitly close the file handler `fp`.

ericzolf: Reference: https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

testing/commontest.py
```diff
@@ -13,9 +13,9 @@
 RBBin = os.fsencode(shutil.which("rdiff-backup"))
 
 # Working directory is defined by Tox, venv or the current build directory
-abs_work_dir = os.getenvb(
-    b'TOX_ENV_DIR',
-    os.getenvb(b'VIRTUAL_ENV', os.path.join(os.getcwdb(), b'build')))
+abs_work_dir = os.getenv(
+    'TOX_ENV_DIR',
+    os.getenv('VIRTUAL_ENV', os.path.join(os.getcwdb(), b'build'))).encode()
```
ericzolf, line 18: Two things here:
1. I think you need to use `getcwd` and `'build`' (both without b) or you might get a mix of strings and bytes (it happens only if you use the test outside of tox, but I do this a lot while developing).
2. instead of str.encode for paths and similar, please use [os.fsencode](https://docs.python.org/3/library/os.html#os.fsencode) to be consistent with the rest of the code.

ericzolf:
I didn't realise but what you described in #443 also happens in the Travis CI, meaning we can't merge even if you fix the few review comments. We need first to fix the Travis pipeline. Something has changed in the environment, it used to work for months. Let me know if you want to fix it yourself (even in this same PR) or if I should do it.

t9t:
Thanks for the pointers, I really appreciate it! I will address them later today. I can also fix the issue described in #443 in this PR, no problem.

When you have some time, I am interested to hear how to run individual tests outside of tox. Being still largely unfamiliar with this ecosystem of tools, I've now resorted to running tox from PyCharm, which is less than ideal.

t9t:
Looks like "later today" is now. :) I have addressed the pointers. I also changed the Travis Windows build command but it seems like it didn't pick it up for the pull request build..., that is a bit awkward. :) https://travis-ci.com/github/rdiff-backup/rdiff-backup/jobs/368143319

ericzolf:
Run individual tests outside of tox: check the `docs/DEVELOP.md` where I explain how to debug, and just remove the dbg/-pydebug things and then it works also with tests. Shortly summarized:

```bash
./setup.py clean --all  # only if you've built wheels before and the shell complains about missing python
./setup.py build
PATH=$PWD/build/scripts-3.7:$PATH PYTHONPATH=$PWD/build/lib.linux-x86_64-3.7 python testing/somethingtest.py
```

The nice thing is that you can also use the unittest parameters e.g. `--failfast`. It works similarly under Windows. If you peek again at my ansible playbook `tools/windows/playbook-build-rdiff-backup.yml`, the 2 last tasks, you'll see that I copy around the rsync.dll and create a bash file to export the necessary variables.

ericzolf:
Regarding the Travis error, it seems to take the new version (I see the added `-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE` option in the job logs) but to not be enough to succeed. No clue what changed, I need to check in my own environment if it still works and what there are for other differences.

Do you have enough rights to see the Travis job logs?

t9t:
@ericzolf Thanks for the testing instructions, I will check it out!

As for the Travis, yeah I didn't look closely enough and was looking only at the abbreviation. Upon closer inspection indeed it seems to have taken the changed command line.

I did some permutations on my machine (because honestly I have no knowledge about cmake at all...) and turns it out it's rather simple. There was `-A Win32` but the build runs on an AMD64 machine in Travis. So now I changed it to `-A x64` (and reverted the other settings) to see if that will work.

ericzolf:
Everything looks good so nothing to do with your changes _but_ I'm still hesitating:
1. it used to work with Win32, why doesn't it any more?
2. as we setup the automated build, we explicitly set Win32 because user voices stated that they need the Windows binary to be 32 bits and not only 64...
3. I now saw that under tools/windows, I already use x64 but that's just my development environment, not what we deliver.

t9t:
Yeah I totally understand. What I observed is that using `-A Win32` with cmake didn't work on my 64 bit computer, nor in the Travis build pipeline for this pull request, but that it does work with `-A x64`. Unfortunately I have no knowledge of cmake so I can't make any judgment about whether it should work or not. Perhaps they did an update which broke this (seems unlikely to me, though).

At the same time I'm fairly certain that my code changes did not break the pipeline either, as it already didn't work on my computer before I started tinkering with the code (hence why I raised #443).

Is there any way to start a Travis CI build for `master` and see if that works just fine?

I would propose that I strip the commits that are trying to fix the Travis build from this pull request and limit the scope to only the changes in the Python files. Then the Travis build issue can be addressed separately and doesn't have to interfere with these other changes.

t9t:
This is the last build on master: https://travis-ci.com/github/rdiff-backup/rdiff-backup/jobs/366370770
This is the first build on my branch: https://travis-ci.com/github/rdiff-backup/rdiff-backup/jobs/368011823

When quickly scanning some of the output, I notice the following:
- master: `creating build\lib.win32-3.7`
- my branch: `creating build\lib.win-amd64-3.7`

So it looks like somewhere either between 7 days ago and now, or between master and my branch, it has changed between using `win32` to `win-amd64`. I can imagine this is also the culprit why it can't find the symbols, as the librsync was built for Win32, but now it's trying to find an x64 library or something. ~I believe this is something in the Python Setuptools which is building the extension, but then again my knowledge of this ecosystem is so limited that this is just a wild guess.~

Doing some experimentation locally again, it looks like it takes this value from the currently used Python installation. When I run `python setup.py build` using a 64-bit Python, it will use `win-amd64` and it will fail when librync was built using `-A Win32`. On the other hand, when I run `python setup.py build` explicitly using a 32-bit Python, it will work fine.

Checking now how Python is installed in Travis, it seems like this is the issue: https://github.com/pyenv-win/pyenv-win/pull/128
`pyenv-win` now by default installs a 64-bit Python rather than 32-bit. And since rdiff-backup just takes the master rather than any specific tag or release, any newly started build will get such changes for free.

I will try and see how difficult it is to pin a `pyenv-win` version in the Travis configuration so that until further changes it will just take the version that was running fine, before that change.

ericzolf:
I've merged back master into this branch so there shouldn't be anything against merging if the tests don't fail again.

ericzolf:
It was a bumpy ride for a first PR (which became a 2nd one)! Thanks for your tenacity!

PR #447, opened 2020-08-07
tox_win.ini: commented failing tests and included deps

t9t:
I took your `tox_win.ini` from https://github.com/rdiff-backup/rdiff-backup/tree/ericzolf-make-tox-work-under-win, then proceeded to run every test and see which one succeeded and failed.

Personally I'm not fond of big-bang pull requests with a million and one fixes in it, so I would not be comfortable making a long-living branch in which I fix all the tests at once. So my suggestion is to merge this basic set up to `master`, after which I can look at, fix, uncomment, and submit each test individually.

I had to add the `deps` section to include the Windows specific dependencies (otherwise there were errors that it could find. eg `pwintypes` (`NameError: name 'pywintypes' is not defined`). Again, this is my lack of knowledge about the tooling, I'm not sure if this is the proper way to deal with this. I have all these dependencies installed globally on my system...

In a previous pull request you noted:
> Just one small thing for the future: check the prefixes to use for changelog-relevant messages in commits, in the developer documentation.

I am a little unsure when this applies and when it doesn't...

### DIFF
--- a/testing/commontest.py
+++ b/testing/commontest.py
@@ -10,7 +10,7 @@ from rdiff_backup.log import Log
 from rdiff_backup import Globals, Hardlink, SetConnections, Main, \
     selection, rpath, eas_acls, rorpiter, Security, hash
 
-RBBin = os.fsencode(shutil.which("rdiff-backup"))
+RBBin = os.fsencode(shutil.which("rdiff-backup") or "rdiff-backup")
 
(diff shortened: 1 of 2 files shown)


ericzolf:
I'm with you, small PRs are better, as long as they don't break master.
Regarding the prefixes, we should add a DEV prefix once we add the windows tests to Travis (whenever you want actually).

PR #457, opened 2020-08-22
Just slightly reformat win_acls for analysis reasons

ericzolf:
It allows one to call something like the following command and get
a meaningful list of calls through the connection (hence kind of
describes the API between client and server):
`grep -ro conn\.[a-zA-Z0-9._]* src testing | awk -F: '{print }' | sort -u`


ericzolf:
Sorry, I pushed the API description to the wrong branch, but it is now rolled back and can be reviewed.

ikus060:
+1

PR #458, opened 2020-08-26
Add description of the API versioning approach

ericzolf:
DEV: add docs/api folder with API description to be followed and API v2.0.

Also create a tool script `prepare_api_doc.sh` to help with documenting the API.

A thorough review is important as it will drive future changes of the code for the foreseeable future.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -7,10 +7,14 @@ installations.
 
 ## API version files
 
-Each version file named vX.Y.md lists the objects and functions or methods
-defining the API interfaces. A basis version can be created using the script
-`tools/prepare_api_doc.sh` but it needs to be improved respecting the
-following structure:
+Each version file named vXYY.md lists the objects and functions or methods
+defining the API interfaces for the major version X, minor version YY.
+We use an integer instead of a traditional version scheme to make comparaisons
+easier internally, and avoid confusion with the application's version.
+
+A basis version of the file can be created using the script
+`tools/prepare_api_doc.sh` but it needs to be improved manually respecting
+the following structure:
 
 * _Format_ contains changes at the communication format as a bulleted
   list of free text describing the changes.
@@ -75,8 +79,8 @@ Each interface is described on a separate line of the following formats,
 * API and application don't need to be aligned, else we wouldn't need to
   consider separate versioning schemes.
 
-> **NOTE:** we consider v1.0 the implicit API version of the 1.y application
-	version and v2.0 the de-facto API version of the 2.0.z application
+> **NOTE:** we consider v100 the implicit API version of the 1.y application
+	version and v200 the de-facto API version of the 2.0.z application
 	versions. The interfaces of both API versions don't differ, they
 	are still incompatible due to changes between Python 2 and 3.
 
@@ -84,12 +88,12 @@ Each interface is described on a separate line of the following formats,
 
 To make the above rules more concrete:
 
-Version 15.0.0 of rdiff-backup uses version 5.0 of the API and has 2 interfaces:
+Version 15.0.0 of rdiff-backup uses version 500 of the API and has 2 interfaces:
 
 * `deepen`
 * `unchain`
 
-A version 5.1 of the API is created with a new function `neglect` and making
+A version 501 of the API is created with a new function `neglect` and making
 `deepen` deprecated:
 
 * `deepen` **deprecated**
@@ -103,20 +107,20 @@ to change API version in a bug-fix version). Version 15.1.0 defines the
 its use, it could be simply a new command line option).
 Version 15.1.0 works hence by default with version 15.0.0.
 
-A version 5.2 of the API could be created, with other changes, but the
+A version 502 of the API could be created, with other changes, but the
 new resp. deprecated states wouldn't change:
 
 * `deepen` **deprecated**
 * `neglect` **new**
 * `unchain`
 
-A version 6.0 of the API can then be created, which removes the deprecated
+A version 600 of the API can then be created, which removes the deprecated
 interface and makes the new interface usable by default:
 
 * `neglect`
 * `unchain`
 
-This version 6.0 could then be used by a new version of rdiff-backup 15.2.0,
+This version 600 could then be used by a new version of rdiff-backup 15.2.0,
 which would work with version 15.1.0 but not with version 15.0.0.
 
 > **NOTE:** we could declare a version 16.0.0 but that shouldn't be necessary
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -1,0 +1,127 @@
+# client/server API documentation
+
+It is important that the API between two versions of rdiff-backup run in
+client/server mode remains stable, so that an upgrade doesn't need to be done
+on both sides at the same time, as some of our users have sizable
+installations.
+
+## API version files
+
+Each version file named vX.Y.md lists the objects and functions or methods
+defining the API interfaces. A basis version can be created using the script
+`tools/prepare_api_doc.sh` but it needs to be improved respecting the
+following structure:
+
+* _Format_ contains changes at the communication format as a bulleted
+  list of free text describing the changes.
+* _Sources_ contains the analysis of the actual source files under `src`
+* _Testing_ contains the interfaces used by the test files. These don't require
+  to be stable and are present mostly for reference.
+
+We also make the difference between `internal` and `external` interfaces:
+
+* _Internal_ are defined in the rdiff-backup modules and are under control of
+  the programmer, and must respect the rules edicted later.
+
+* _External_ are defined in Python and 3rd party libraries, they are
+  rather independent of the programmer of rdiff-backup but more of the
+  system on each side of the connection, python version installed. This can
+  help to assess which version of Python can be supported. It must follow
+  similar rules to deprecated and new interfaces e.g. so that users can upgrade
+  their python version timely.
+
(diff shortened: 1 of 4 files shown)


ikus060:
Nice job.
I'm not so convince about the way you plan the backward compatibility. On paper it sound OK, but I'm a bit worried about the timeline. I'm very conservative in that regards and would keep a rdiff-backup backward compatible with any version older then 1 year.

Meaning, if we release a new version of rdiff-backup, I would expect it to be working with a version that was release 1 year ago. Kinda

After that, the release number doesn't really matter if we respect this principle. And we should probably move this dicussion into the mailinglist.

ericzolf:
Before I start with your main point: I've pushed a new commit changing the API versions to integers. It'll make it less likely that someone mixes it with the application version and makes it easier to compare.

More important, I also had in mind that we'll need to make sure that we leave enough time for distros to package one version and for users to deploy it, but one year seems a very long time. As I don't have yet a feeling how it will be practically, I might be wrong... I'm also thinking how we'll keep track of this properly, because it's the time between the deprecation/introduction and release of _any_ interface, and it's final removal/usage in a next release.

What I mean:

1. we have version 2.0 (API 200)
2. we deprecate one interface in version 2.1.0 released in Jan 2021 (API 201)
3. we add a new interface in version 2.3.0 released in June 2021 (API 202)
4. etc...
5. deprecate another interface in version 2.7.0 released in December 2021 (API 220)
5. assuming API 300 is ready in Jan 2022, we could only remove the one interface of API 201 and nothing else from the whole year...

I'm slightly concerned about the speed of evolution, but I'm more concerned about how to keep track of this properly and simply. I need to think about it... Perhaps we should after all align major versions of application and API and declare that changes introduced (deprecate/new) in API 2YY can only be applied (removed/used) in API 400, released together with v4.0.0 of rdiff-backup, so that v3.x can run with v2.x and v4.x of rdiff-backup, but version 2.x can't run with version 4.x. Then we only need to make sure that we don't release major versions of rdiff-backup too often (once per year, but I'm still struggling with this idea).

This approach would also have the advantage that users don't need to care at all about API, it would only be relevant to the developers, and they would have a clear rule regarding compatibility between client and server.

derekatkins:
In my opinion, breaking client-server communication even over a 2 year period is too quickly.

As a case in point, I can use a modern SSH client to talk to an SSH server from **10** years ago.. And it just works!  Sure, it will use older crypto, and the older version might not support modern (ecc) keys, but the underlying client-server protocol is stable.

I consider backup software as critical as SSH.

Please repeat this mantra:  One does not have control over all assets being backed up to ensure that they are always running the most up-to-date systems and software.

In this case, the Postel Rule should apply:  be conservative in what you send, but be liberal in what you accept.  In other words, versions 3.0, 4.0, 5.0 etc should be able to talk to version 2.0.  Even version 10.0 should be able to talk to version 2.0.

ericzolf:
I hear you but you need to consider that SSH is a professionally developed product with zillions of developers behind, and without interruption of 9 years or so in its development history. Hopefully/possibly they had foreseen the evolution and have a modular setup for the different protocols and algorithms they have added/removed out there.

All what you suggest is understandable, but it means additional effort to maintain the code and never getting rid of old code, which can't be acceptable to me. This one year rule is already a big stretch, meaning I can get rid  of old code in average only 1.5 years after I've cleaned it up. It wasn't foreseen before Patrik put the requirement on the table. I've reluctantly said that I'll try to adhere to the rule, without promise yet (still munching on this one, your intervention makes me think that I'll have to adhere to it as a middle way; give me time to think it over though), but I can definitely say that your ask isn't going to happen, or only by chance.

On the other hand, who knows, perhaps it's less a mess than I think and all will be fine, but I don't think so and I prefer a rule that we can follow, and our users can rely on, than no rule or a rule which breaks the development and can't be followed. Once the code is clearer, perhaps we can revisit this decision and come to another conclusion.

On a personal note, as an Automation Architect in Consulting, I see many customers who tell me the same story about not having their environment under control, there are ways to this issue and it helps in terms of security, efficiency and agility.

This said, if you have concrete arguments on why it is difficult to upgrade rdiff-backup, we can look at ways to make it simpler.

Sorry I can't be more positive, and thanks for your feedback, I truly appreciate that you participate in the discussion,

ikus060:
Another solution to mitigate this constraints is to change the way we packages and deliver rdiff-backup. A right full approach is to continue supporting and delivering rdiff-backup2 for long term support distribution. Meanwhile we start working on rdiff-backup3 that introduce backward incompatible changes. And we need a way to make our packages installable side by side. In other words, we need to allows the installation of rdiff-backup2 and rdiff-backup3 on the same system.

And here you have a solution to clean-up rdiff-backup at the speed you want, while keeping everything backward compatible. We only need some effort to maintain v2.x.x branch.

Details should be discuss about how long we would maintain each version, etc.

derekatkins:
While there was a company behind the initial implementation of sshv1, the v2 protocol was standardized by a relatively small group of people.  I was there -- there were maybe a dozen (or less!) people actively working on the SSH protocol standard, and I really doubt there are zillions of people working on OpenSSH (but you are right that it is probably a larger group than we have here).  The point is that they made a concerted effort to ensure the client-server portion of the protocol remained stable, while allowing the endpoints grow and mature.

For example, I don't care if the on-disk format for the backup metadata changes over time.  I don't care if local data is compressed by gzip or bzip or lzh or whatever.  I don't care if you use a flat file, directory tree, or sqlite.  All I care about is that the communication between rdiff-backup and rdiff-backup --server remain consistent for a long time (perhaps with the some level of internal protocol negotiation, as required).

Having been in protocol standardization for 30 almost years now, it SHOULD be easy, if you make a concerted effort, to build a network protocol that allows you to continue to communicate even while you make changes to the program over time without requiring both ends to be upgraded.

It DOES mean there will be a flag day.  With SSH, there was a flag day between v1 and v2.  Servers supported both versions for a long time, but then eventually v1 went away (and servers finally ripped that code out).  I believe that transition lasted a decade.  Maybe that flag day was rdiff-backup-2?  Or maybe we still need to have that flag day.  But I would really like to see (sooner rather than later) a stable network protocol that we can live with that will allow us to maintain client-server compatibility for long periods of time.

ikus060:
@ericzolf Another way to mitigate the transition and make stuff backward compatible would be to play arround `eval()` calls made in `connection.py`: eval(expression,globals, locals). If we properly pass a mock dict with globals, we can really do what ever we want.

e.g.:
```
# Patch api_v1 to be backward compatible
api_v1 = {
  'regress': lamba x: rpath.regress
}

# Here regress is removed.
api_v2 = {
   
}
...
eval(request.function_string,locals=api_v1)(*argument_list)
```

ericzolf:
Sorry for the delay, but I was also in holiday. Anyway, as Derek wrote, it requires a concerted effort, which hasn't yet happened, before we can stabilize the interface. So, I fear, however we proceed, there will be a period of relative instability before it can get better, even if I remain convinced that we can do the switch over 2 or more versions to simplify upgrades.

I'll update the text in the sense that we want to keep the compatibility as long as possible, at least over one year and one major version. We can then discuss how to reach this objective.

ericzolf:
I've pushed few changes to the document, making it clearer that we strive to keep backward compatibility for a longer period of time. Check the [last commit](https://github.com/rdiff-backup/rdiff-backup/pull/458/commits/e93ee62e856c776dc5a240c97a085d3856b4426c) so that you don't need to read the full document.

ikus060:
Let me try to explain my point regarding backward compatibility.

The goals is simple: If a major distribution like Debian keep rdiff-backup v2.0.0 we need to support that version and make it available on a wide range of distribution to keep it backward compatible. During the migration from v1.2.8 and v2.0.0 we really lost the opportunities to make it easy for the user. If we had kept the package named rdiff-backup for v1.2.8 and provide a new packages rdiff-backup2 for the v2.x.x series if would have solve a lot of trouble to the users. I'm not making this solution, Debian is using this technique quite often for libraries that brake the ABI.

So for me it goes as follow.
1. Solution 1: We provide and maintain multiple major version v1, v2, v3. Each major is compatible only with the same major version. For each major version we provide new packages: rdiff-backup1, rdiff-backup2, rdiff-backup3
  * Pros:  We can do breaking changes without caring about inter-version compatibility.
  * Cons: Will need to maintain multiple branches and possible back-port fixes.
2. Solution 2: We provide and maintain a single major version and we write the code with compatibility shim to make it backward compatible with v2, v3, v4, etc.
  * Pros: We provide a single binary and we support only the latest one.
  * Cons: I'm not sure it's event possible to write the compatibility shim for each version. As we saw for v1 and v2 It was technically too difficult.

My vision would be to work toward a v3.0.0 in a seperate branch that would introduce yet again a new network protocol. But we can work on this new protocol and define it on paper. Similar to SSH or RSYNC protocol that are well defined. Meanwhile we continue supporting v2 until v3 is stable.

ericzolf:
Final review?

ikus060:
I'm aligned with your latest comments. And I think we are aligned with the ultimate goal to keep stuff backward compatible as mush as possible.

I really want to avoid the incompatibilities we faced with migration from v1.2.8 to v2.0.0. As an example for Minarca, I need to deploy rdiff-backup v1 and rdiff-backup v2 on the server in order to keep everything working for my users. Otherwise, I would have to upgrade minarca-client every where it's deployed.

ericzolf:
Thanks for your inputs. Merging now.

PR #459, opened 2020-09-02
Add no-compression defaults for .webm and .tzst

dgasaway:
.tzst is the abbreviated form of .tar.zst, see https://github.com/facebook/zstd/blob/dev/programs/fileio.h
.webm is the video container created by Google, see https://www.webmproject.org/docs/container/


ericzolf:
Merged. Thanks for the help!

PR #464, opened 2020-09-23
Draft: Fix ACL for non-UTF-8 folder

ikus060:
Make use of folder file descriptor to get the ACL to workarround
limitation of pylibacl interface not supporting bytes nor
surrogateescape.


ikus060:
Nevermind, "fd" paramter doesn't fetch the default ACL.
Looking at the pylibacl code we don't have any workarround: 
https://github.com/iustin/pylibacl/issues/16

PR #467, opened 2020-09-26
Add api version handling

ericzolf:
Based on the documentation created previously, I've added the concept of API version with min, max, default and actual one. The actual one can be "agreed" between client and server based on their respective default/min/max values.
I've also added an option `--api-version` to change the default version used (currently 200, which is equivalent to version v2.0.x), and an option `--versions` to output this API version and more information, in YAML format for easier parsing by potential scripts.
Last but not least, I've added few tests to cover the new functionality.

### DIFF
--- a/docs/rdiff-backup.1
+++ b/docs/rdiff-backup.1
@@ -75,9 +75,11 @@ parameter.
 .BI "\-\-api-version " api
 Sets the API version to the given integer between minimum and maximum versions
 as given by the
-.B \-\-versions
-option. It is the responsibility of the user to make sure that this version is
-also supported by any server started by this client.
+.B \-\-version
+option, with API above 200, i.e. by calling something like
+.B rdiff-backup \-\-version \-\-api-version 201.
+It is the responsibility of the user to make sure
+that this version is also supported by any server started by this client.
 .TP
 .B \-b, \-\-backup-mode
 Force backup mode even if first argument appears to be an increment or
@@ -536,13 +538,10 @@ Check all the data in the repository at the given time by computing
 the SHA1 hash of all the regular files and comparing them with the
 hashes stored in the metadata file.
 .TP
-.B "-V, \-\-version, \-\-versions"
+.B "-V, \-\-version"
 Print the current version and exit.
-.B \-\-versions
-(plural) is only known
-by rdiff-backup starting with version 2.1 and outputs also information about
-API, call, python and operating system. This is also the default behavior
-of the traditional options if the API version is strictly more than 200.
+Starting with version 201 of the API, it outputs also information about
+API, call, python and operating system.
 
 .SH ENVIRONMENT
 .TP
--- a/docs/rdiff-backup.1
+++ b/docs/rdiff-backup.1
@@ -72,6 +72,13 @@ using the
 .B \-\-remove-older-than
 parameter.
 .TP
+.BI "\-\-api-version " api
+Sets the API version to the given integer between minimum and maximum versions
+as given by the
+.B \-\-versions
+option. It is the responsibility of the user to make sure that this version is
+also supported by any server started by this client.
+.TP
 .B \-b, \-\-backup-mode
 Force backup mode even if first argument appears to be an increment or
 mirror file.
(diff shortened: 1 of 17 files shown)


ericzolf:
Shoot, formatted strings literal e.g. `f"some {placeholder} string"` exists only since 3.6... Good reason to get rid of 3.5, I'll do it in a separate merge request. This one is until then WIP.

ericzolf:
Formatted string literals are very concise and look cool but I think that they're not good for future developments like translations as they're interpreted right away and can't be delayed like str.format can (in other words, you can't prepare the string with place holders, e.g. read them from a file, and replace the place holders explicitly later).

Anyway, the PR is ready for review.

docs/rdiff-backup.1
```diff
@@ -72,6 +72,13 @@ using the
 .B \-\-remove-older-than
 parameter.
 .TP
+.BI "\-\-api-version " api
+Sets the API version to the given integer between minimum and maximum versions
```
ikus060, line 76: Would rephrase this to : "Force an older API version to be used instead of the default."

docs/rdiff-backup.1
```diff
@@ -529,8 +536,13 @@ Check all the data in the repository at the given time by computing
 the SHA1 hash of all the regular files and comparing them with the
 hashes stored in the metadata file.
 .TP
-.B "-V, \-\-version"
-Print the current version and exit
+.B "-V, \-\-version, \-\-versions"
+Print the current version and exit.
+.B \-\-versions
```
ikus060, line 541: I'm not a big fan of this --versions option. Could we stick with `--version` and use `--verbose` to display more details ? Or something similar.

src/rdiff_backup/Main.py
```diff
@@ -416,6 +440,13 @@ def _Main(arglist):
     cmdpairs = SetConnections.get_cmd_pairs(_args, _remote_schema, _remote_cmd)
     Security.initialize(_action or "mirror", cmdpairs)
     rps = list(map(SetConnections.cmdpair2rp, cmdpairs))
+
+    # if any of the remote paths is None, we have an error.
+    # We continue to test-server so that all connections can be tested at once.
+    if any(map(lambda x: x is None, rps)) and _action != "test-server":
```
ikus060, line 446: How is this related to the API version ?

src/rdiff_backup/SetConnections.py
```diff
@@ -180,19 +183,26 @@ def _init_connection(remote_cmd):
     conn_number = len(Globals.connections)
     conn = connection.PipeConnection(stdout, stdin, conn_number)
 
-    _check_connection_version(conn, remote_cmd)
+    if not _validate_connection_version(conn, remote_cmd):
+        return None
```
ikus060, line 187: Why returning `None` ? Shouldn't we stop the process with a fatal error ? At least I would raise an exception instead and let the caller handle the exception.

docs/rdiff-backup.1
ericzolf: No, the default is currently the oldest one, mainly based on our discussion that you want stability. This said, it might change in the future.

docs/rdiff-backup.1
ericzolf: Version is output before the logging mechanism is setup, so difficult to do it like this. I also wanted to make sure that we get all information we need without having to explain too many cases. I could use solely the API version to decide about the output given... Again, my only driver was to be backward compatible. Let me think about it.

src/rdiff_backup/Main.py
ericzolf: It's related to the "None" discussion after: basically, by sending None, I can test _all_ servers and don't need to stop each time one fails, forcing the user to call again `rdiff-backup --test-server` to make sure they identified all failing servers. I had to touch this part of the code because I added the handling of the API version between client and server.

src/rdiff_backup/SetConnections.py
ericzolf: First, I've always learned that exceptions take more time than normal if/then/else clauses, so it plays against performance and should be used sparingly. Second, see my comment previously, this is how I can test more than one server at once. And, third, I also wanted to cleanly quit the code and not somewhere in the middle.

ericzolf:
OK, I removed `--versions`.

docs/rdiff-backup.1
ericzolf: I removed --versions, `--version` outputs the same thing as long as the actual `api-version` is above 200.

ikus060:
I might not be align with the changes to `--version` output. Obscure script might use `--version` to do something special. Might be better to keep `--version` output untouched and add a new `--api-version` maybe to output the list of compatible API ?

Othersie all your change are ok to me.

ericzolf:
Good point, this can be considered like an API change, actually it already was set as such, but not explicitly. It's now properly documented (and the default was anyway API v200 and the old format).

PR #468, opened 2020-09-27
Introduce Python 3.9 and remove Python 3.5

ericzolf:
CHG: rdiff-backup now supports the newly released Python 3.9 and stops supporting the obsolete Python 3.5.
It's not completely true now but it'll be true once the changelog is written on this basis.
The order of versions in Travis-CI has also been changed to have the supported Windows version be first;
I think this will allow to simplify the setup in an upcoming step.

### DIFF
--- a/tools/build_wheels.sh
+++ b/tools/build_wheels.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e -x
 
-pybindirs="/opt/python/cp36*/bin /opt/python/cp37*/bin /opt/python/cp38*/bin /opt/python/cp39*/bin"
+pybindirs="$*"
 
 # Install a system package required by our library
 yum install -y librsync-devel
--- a/tools/build_wheels.sh
+++ b/tools/build_wheels.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e -x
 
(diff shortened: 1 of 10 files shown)


ericzolf:
Struggling with the right quoting in .travis.yml...

ericzolf:
OK, I don't think I'll get it any better, so please review and squash/merge.

tools/build_wheels.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e -x
 
-pybindirs="/opt/python/cp35*/bin /opt/python/cp36*/bin /opt/python/cp37*/bin /opt/python/cp38*/bin"
+pybindirs="$*"
```
ikus060, line 4: Any particular reason not to keep the python version here ?

ericzolf: Yes, I forgot first to change it 🙄 . What I mean is that it's easier to maintain it if the lists of Pythons are in as few places as possible, i.e. Travis, Tox and docs.

ikus060: Agree.

With this in mind, we could also move the list of python version out of tox and define them only in travis. But wecan keep it for another PR.

ericzolf: I don't think it would work but if you find a way to overcome those concerns, that would be cool:

1. because I don't think tox/tox.ini can be parametrized in that way
2. I need tox to run locally without Travis involvment

PR #470, opened 2020-10-01
Replace assert through unittest's self.assertSOMETHING in testing

ericzolf:
The assertXXX functions fit better in the unittest framework, give by default more information in case something goes wrong.

Because the os.system calls didn't so well in the concept, I introduced also the xcopytree to make it easier to copy trees independently from the operating system.

### DIFF
--- a/testing/commontest.py
+++ b/testing/commontest.py
@@ -566,5 +566,22 @@ def iter_map(function, iterator):
         yield function(i)
 
 
+def xcopytree(source, dest, content=False):
+    """copytree can't copy all kind of files but is platform independent
+    hence we use it only if the 'cp' utility doesn't exist.
+    If content is True then dest is created if needed and
+    the content of the source is copied into dest and not source itself."""
+    if content:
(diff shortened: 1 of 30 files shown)


testing/commontest.py
```diff
@@ -566,5 +566,22 @@ def iter_map(function, iterator):
...
+    hence we use it only if the 'cp' utility doesn't exist.
+    If content is True then dest is created if needed and
+    the content of the source is copied into dest and not source itself."""
+    if content:
+        subs = map(lambda d: os.path.join(source, d), os.listdir(source))
+        os.makedirs(dest, exist_ok=True)
+    else:
+        subs = (source,)
+    for sub in subs:
+        if shutil.which('cp'):
```
ikus060, line 580: Guessing it's mostly for windows ?

ikus060:
It's a long commit. The changes look good.

testing/commontest.py
ericzolf: Yep, but I didn't want to be specific, BeOS might not have `cp` either :-)

PR #474, opened 2020-10-02
Fix format and improve checks of rdiff-backup-delete

ericzolf:
There were flake8 issues and one could try to delete a file within rdiff-backup-data (not sure it would have gone but it wasn't checked properly).

@ikus060 I see that you've tried to support python 2 but on the other hand, there are bytes constructs which are not covered, e.g. `b'rdiff-backup-data'` or `bytes(...)` calls in the `_unquote` function. Do you plan to make it really work under Python 2 (or perhaps it does already?) or shall I clean-up and simplify the code?

### DIFF
--- a/src/rdiff-backup-delete
+++ b/src/rdiff-backup-delete
@@ -43,7 +43,7 @@ SUFFIXES = [b".missing", b".snapshot.gz", b".snapshot",
 
 
 def _bytes(value):
-    if isinstance(value, bytes if PY3 else str):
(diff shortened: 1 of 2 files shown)


ikus060:
@ericzolf The script was working on python2 yes. For contractual reason. The ideas was to support it for a while until rdiff-backup v1.2.8 is really dead.

ericzolf:
Ok, then I don't clean it up further. It should still work under python 2, can you check?

ikus060 (approved):

PR #476, opened 2020-10-06
Remove or explain asserts and simplify fs_abilities

ericzolf:
Instead of doing an init of the FSA object and then setting the details, this is now just one step.


ericzolf:
@ikus060 I think you oversaw this one PR (as you checked both the one after and the one before)...

ikus060 (approved):

PR #479, opened 2020-10-18
Improve benchmark (aka slow) test

ericzolf:
I'm a bit concerned that code changes might impact the performance without
us noticing. Absolute benchmark times are too much dependent on the sizing
of the testing platform to be relevant, but I'm hoping that the relative
numbers, compared to rsync, are more relevant.


ericzolf:
The relative speed isn't as stable as I had hoped. I still take the chance to simplify the code in this PR. Perhaps I'll find a better way to get a more stable number we could validate against, but it'll be something for a further PR. Ready for review.

ottok:
Good goal to test for performance regressions. Unfortunately I don't have any additional insight into how to best do it.

PR #481, opened 2020-10-28
Sort the elements and mark the API functions

ericzolf:
It's a big but (again) boring pull request: I've defined a default sorting order in CODING.md and accordingly sorted all the modules. Additionally, I've marked all functions making out the API so that it becomes easier for developers to pay attention when touching them (as documented in docs/api/README.md).


ottok:
Skimmed through quickly but I don't have anything in particular to comment here. Sorry, I am not being very useful in this review.

ericzolf:
It's OK, it was a pretty boring PR, thanks for having a look.

PR #483, opened 2020-11-05
Clarify meaning of man-page regarding hard links

ericzolf:
DOC: man page has been clarified regarding --no-hard-links option

Based on the discussion on the mailing list.


ottok (approved):
OK!

Issue #496, opened 2020-12-21
[BUG] Unexpected MemoryError on increment


xastor:
## Bug summary

In some cases a backup increment over ssh fails unexpectedly with a MemoryError.  Subsequent backups have the same error.

## Version, Python, Operating System

Call `rdiff-backup --version --api-version 201` or `rdiff-backup -v9` and replace the following line with the output, repeat for each environment impacted:

```yaml
PS C:\Users\Pieter.Bonne> D:\repos\backup_manager\rdiff-backup-test\rdiff-backup.exe -v9
2020-12-21 21:25:03.230011 +0100  <CLIENT-19452>  Using rdiff-backup version 2.0.5
2020-12-21 21:25:03.230011 +0100  <CLIENT-19452>        with cpython D:\repos\backup_manager\rdiff-backup-test\rdiff-backup.exe version 3.7.7
2020-12-21 21:25:03.230011 +0100  <CLIENT-19452>        on Windows-10-10.0.18362, fs encoding utf-8
2020-12-21 21:25:03.231011 +0100  <CLIENT-19452>  Fatal Error: No arguments given
See the rdiff-backup manual page for more information.
```

## rdiff-backup call

How did you call rdiff-backup, with which parameters:

```
D:\repos\backup_manager\rdiff-backup-test\rdiff-backup.exe -v5 --exclude-symbolic-links --remote-schema "C:\Windows\System32\OpenSSH\ssh.exe %s /usr/local/bin/rdiff-backup --server" --include "/root/db_dumps**" --include "/etc/apache2**" --include "/var/www**"  --exclude "**" --print-statistics "root@1.2.3.4::/" "D:\Backups\product-test"
```

## What happened and what did you expect?

I expected a successful backup increment.  

## More information

I'm backing up from a remote server to a local disk.

- The initial backup was successful.
- After that, I updated one of the database dump files on the server (166Mb)
- Any backups after this have the MemoryError error.

Below is a log where the MemoryError occurs : 

```
PS C:\Users\Pieter.Bonne> D:\repos\backup_manager\rdiff-backup-test\rdiff-backup.exe -v5 --exclude-symbolic-links --remote-schema "C:\Windows\System32\OpenSSH\ssh.exe %s /usr/local/bin/rdiff-backup --server" --include "/root/db_dumps**" --include "/etc/apache2**" --include "/var/www**"  --exclude "**" --print-statistics "root@1.2.3.4::/" "D:\Backups\product-test"
[...]

PR #499, opened 2020-12-22
[DOC] Add how to use putty and Microsoft's OpenSSH as SSH client

ericzolf:
DOC: document how to use Putty as SSH client thanks to @xastor in #496


xastor:
Noticed something : there's "and that it's installed there." at the ending that seems a bit out of place.

ikus060 (approved):

PR #502, opened 2020-12-26
add --gitignore option to not sync gitignored files

apoelstra:


apoelstra:
I know the dev docs say to open an issue for discussion before adding stuff, but this was very quick to do and solved a problem I was having (how to ignore build artifacts etc from git repositories that I've got all over my home directory). I'm happy to clean this up or make other changes, although I can't test on Windows or OS X.

This code probably isn't usable as-is, since I ascii-decode paths, which shouldn't be necessary and is likely broken for many files, but I'm not very familiar with how to work with filepaths in Python.

Feel free to close if this isn't something the project is interested in.

ericzolf:
I haven't done a full review yet of the code but I'm already struggling with this PR:

1. the git library doesn't work properly under Windows according to https://github.com/gitpython-developers/GitPython/issues/525 and rdiff-backup must remain cross-platform
2. it also leaks memory and can crash, which isn't something we want in a backup tool.
3. it is very specific in the sense that we could also add "intelligence" for Mercury, or any other special software. That doesn't seem to scale to me, until we have some kind of plug-in concept.
4. it's perhaps me, but if you don't want to back-up your gitignored files, and don't _need_ to back-up the non-gitignored files (because it's Git, and already backed-up so to say), just ignore the whole Git repo. So what would be the use case for this PR?

apoelstra:
I have no access to Windows machines, nor do I know how to deal with memory leaks in Python. So I think this PR is dead on arrival.

I don't understand your suggestion to ignore the entire repo :). git repos are not "already backed-up".

In any case, even for my personal usage I've moved away from this and instead just added the same flag to rsync using libgit2, which is 1000x faster than this approach, and has a much clearer memory model and more of a hope of being cross-platform.

Thanks for taking a look, and sorry to waste your time.

Issue #280, opened 2020-02-24
Re-implement the arguments parsing using argparse module


ericzolf:
This will make the code more modern and robust, and give us "for free" the `--help` option.

The behaviour shall not change at all, only the implementation, using https://docs.python.org/3/library/argparse.html#module-argparse

ottok:
Another option is https://typer.tiangolo.com/ or https://click.palletsprojects.com/en/7.x/ (which Typer also builds upon).

ericzolf:
Typer doesn't seem to be broadly packaged (not in Debian or Fedora), whereas click seems more present, but argparse is part of the standard library, so it needs to be weighted if those non-standard libraries are worth the additional dependency.

PR #517, opened 2021-01-17
Rewrite argument parsing

ericzolf:
It's a rather important change and I would appreciate a review. The most important news and changes are:

    NEW: rdiff-backup has a --help parameter
    NEW: rdiff-backup has a new interface with actions and sub-options, use --new --help to get the help
    CHG: the old command line interface without actions is considered deprecated
    NEW: new action 'info' to output system information, try 'rdiff-backup info'
    CHG: restoring a specific increment requires now the use of '--restore' parameter

With the exception of the last change, the old CLI still works as it did. The breaking change was unavoidable as the old CLI was guessing the action too late for the next steps I have in mind.

Because this is only the first step, as I have in mind to make the actions pluggable. They are currently defined in arguments.py, but the next step is to give them their own namespace, most probaby `rdiffbackup/actions`, and to also get rid of the many global variables in Main.py.

One more comment: `arguments.py` is in `rdiffbackup` instead of `rdiff_backup` as to get a clean new namespace, which will get filled step by step, as the old stuff remains in rdiff_backup.


ericzolf:
Because I forgot: closes #280

ikus060:
@ericzolf I indent to take time to review your change this weekend. I'm a bit busy lately.

ikus060:
Clearly it's an improvement, but to be honest, the change is so invasive that I have no clue if something is broken or if a backward incompatibility was introduce.

I have not clue why a new set of arguments parsing was introduce (--new). Why do we need to sets of command line !? I really don't foresee this change as an improvement. (1) It make the development harder since will need to support both version. (2) It make the support harder too because will need to provide guidance according to the version of the arguments used ?? (3) documentation will be more difficult cause we need to clarify if it's a new argument or not, etc. 

I know one command line that behave like this `ps` trying to support UNIX, BSD and GNU. That make this command line overly complicated to use on the fly.
image
Is it really where you want to head ?

Lastly, I thought the general idea before working on a big change was to open a ticket and raise a discussion on the subject. As per the guideline:
> Before committing to a lot of writing or coding, please file an issue on Github and discuss your plans and gather feedback. Eventually it will be much easier to merge your change request if the idea and design has been agreed upon, and there will be less work for you as a contributor if you implement your idea along the correct lines to begin with.

I feel as being the bad guy here, just complaining. But I need to raise my concerns.

ericzolf:
Well, the problem is, I can't at the same time:

1. improve things as I'm willing to do,
2. keep compatibility "forever" as you, among others, requested me to do, 
3. and keep things simple, as we both would like to have it.

I'm doing my best to balance all 3 requirements, but I can NOT satisfy all 3 equally. My current strategy is to keep one version in-between compatible with N-1 and N+1, so that upgrade isn't too painful, mark clearly the parts which are only kept for compatibility (the compat200 and the @API things), so that I can easily remove them when my users, you included, allow me to do so.

And it's really unfair to tell me that I didn't steer the discussion: I sent an early version of the command line to the mailing list and asked for feedback, and got exactly 0 answers. I had to continue on my own, or would stop doing any development. Also the API compatibility approach was discussed at length with you and others on the mailing list, where I accepted changes painful for me.

And the change was necessary because I need to simplify and disentangle the code before I can fix certain bugs safely; I also want to have more plugins-alike interfaces to make it easier for others to start with coding. I started with the CLI because it was already very entangled between actions, general options and sub-options, so I needed to disentangle first this aspect, before going deeper into the code and cleaning it up, making sure that the compatibility is kept step after step, developing clear plug-in interfaces, and documenting them at the same time. Difficult enough? Next steps will be even more fun, getting rid of (most) global variables, split the actions into different classes, handled as plug-in, go to the next level of the code, identify the next (plug-in) interface, rinse and repeat...

ikus060:
I think we should discuss this together other the phone or something. 

I believe it's perfectly doable to improve thing, keep compatibility and keep stuff simple if we change the way to deliver rdiff-backup. I've already write my recommendation in a previous PR, but the idea did not make it.

The general idea is to fork rdiff-backup for each breaking change. Having a branch for v1, v2 and v3. And we change our way to deliver rdiff-backup by providing a different binary name for each version. rdiff-backup2. rdiff-backup3, etc.

The benefits:
* It allows the development to go crazy and break everything
* It keep the code clean since we only need to support interoperability for major version

Maybe we need to reconsider this approach again ?

From my point of view and for the sake of rdiffweb and minarca, this is a very supportable model. I could make sure to support each major version quite easily. As of today, v3 could be our next stable version. v2 is our stable version. v1 is our old stable version. I would recommend v2 for production deployment and v3 for lab and experimentation, etc.

It keep stuff clean and simple.

ericzolf:
We can talk but I need to find a time.

Anyway, I think that I have a solution:

1. for the "normal" use cases, we still think to have one compatible version in-between just because we don't want people to upgrade all at once, mainly because if they take the packaged version from their distro, they can't decide when they get the new version, so Debian might be one week later than Fedora or vice-versa, and we don't want people to stumble about those things. This should keep these people happy and it is important, IMHO. The measures taken (API and compat200) allow me to do this with reasonable effort, at the price of a _temporary_ complexity of the code.
2. for the few people with special needs, the solution is actually very simple. Use virtualenvs, something like:

```bash
virtualenv $BASEDIR/rdiff-backup-2.0
$BASEDIR/rdiff-backup-2.0/bin/pip install rdiff-backup==2.0.5
$BASEDIR/rdiff-backup-2.0/bin/rdiff-backup --version  # result is 2.0.5
```
The user then only needs to use the `--remote-schema` option to point at the right version. There are even variations, like each user having a different PATH pointing at the right virtualenv preferably. And I could add a %v placeholder to simplify the use of the `--remote-schema` (it would only be the first or 2 first places of the version, e.g. 2 or 2.0).

What do you think?

Practically, 2.2 would have both CLIs, old and new, but already 2.4 would have only the new one. We also need to document the above, but that's easy.

ikus060:
> mainly because if they take the packaged version from their distro, they can't decide when they get the new version
This is where my view diverge. We can still provide a package rdiff-backup1, another package rdiff-backup2 and so on. It's not the only pacakge where I see this. Java, python, nvidia, but many more too.

ericzolf:
First, you didn't say what you don't like with the approach with virtualenvs. It's a simple one and of course you can also create a wrapper script `/usr/local/bin/rdiff-backup-2` which calls `/usr/local/lib/rdiff-backup-2/bin/rdiff-backup` if this is what you need (assuming `/usr/local/lib/rdiff-backup-2` is your virtualenv).

Second, there are packages which use this parallel numbering but, unless for libraries, it's a small minority and only in exceptional cases (at least, the ones I vaguely remember), none of them make it the normal way to progress versions.

As former packager, I can tell you that there is a reason: each new package (and each rdiff-backup-N would be a new package) requires a more or less heavy procedure to be integrated into a distro, and package maintainers have generally little understanding when the structure of the software they're packaging is changing regularly (at least, I was swearing each time  it happened, and I had to have a discussion with the developers). And I haven't seen any Python library with such a versioned namespace (and we would have to version the namespace rdiff_backup to make it possible). But ask @frankcrawford and @ottok for their opinion.

As developer, having the name of the namespace isn't fun either, Git has problems to follow renamed/moved files and each incompatibility would require a lot of renaming for "nothing". I mind less to use my brain and guarantee compatibility for a transition version, as it profits everybody, and I committed to it. Our users will less notice it, they would notice (and they might oversee it) that a new version requires to install a new package, and it wouldn't solve the issue that the client side of their installation might already have the new version, where the server side doesn't have it yet, and it would still force them to switch both sides at the same time and/or do some configuration magic. However we do it, a versioned rdiff-backup is more intrusive for our users than a compatible rdiff-backup.

frankcrawford:
Eric, Patrik,

I was wondering when I should add a comment in here, as compatibility
between the versions is a big issue packagers have to worry about, and
it is likely there is no easy  solution.

By way of example, we have created both rdiff-backup and rdiff-backup-1
(or equivalent) in multiple distributions, but in reality only the
current version is in the actual distribution.  The older one is
usually in some form of side archive (e.g. COPR for Fedora/EPEL), which
is not usually not known to most end-users.  So, having different
numbering is really not easy on us packagers.

Of course, compatibility over multiple releases is the goal, as with
something like rdiff-backup, there will always be multiple different
versions in an client environment, as there will be different releases
of OSs in any environment.  Again, having experienced the issues with
RHEL 6 & 7, when we released rdiff-backup 2.x, we will always need to
work out how to keep compatibility working, and probably more than just
between N-1 and N.

In my case RHEL 6 will never get updated for most people, RHEL 7 will
probably not get anything except bug fixes going forward, and even RHEL
8 is supposed to be kept pretty stable now.  As such it is likely we
will need to put in place some form of wrapper for any significant
upgrade and also consider if we are talking about the upgrade really
being for the central server or the multiple client version.  It is
easier to get them to update a central server vs lots of clients at the
same time, although often the central server will be a much older host,
running some old version.

So, to summarise, from my point of view, trying to keep some sort of
compatibility across the same name is easier for packaging, than
separate version, but it does need to be available for a fair while.
 It also needs some simple way for the end use to know that there is a
change, because they never read the upgrade notes, and then only query
about issues months later, when they have a little time.

Regards
Frank

ericzolf:
So, for me, we have 3 different types of users:

1. individuals: they have one or two desktops to backup, possibly not even remote back-up, only local to an external disk. I would assume that they mainly use the package of their distro (if Linux). For them we need to make sure that each update/upgrade is rather transparent, but they can upgrade their environment more or less at the same time.
2. [SMB](https://en.wikipedia.org/wiki/Small_and_medium-sized_enterprises) & similar: they will also use the packaged version of their distro, but they have a rather small number of servers, which they have under control, and can upgrade in finite time. They still need some time to upgrade their whole environment.
3. big corporations and service providers (like Patrik): they have many hosts which they don't have under control, some of them very old, and they can only upgrade very slowly, sometimes only once all the old cruft died out (I personally still have customers with RHEL5 :roll_eyes:). The clients might or might not be packaged (for sake of simplicity, I consider "wheels" as not packaged), it's not under their control, but they need to support them all on their backup server(s). I assume that they have the choice if they install via package or wheel/code, whatever, on their servers.

Let me know first if you agree with the above analysis.

Now I continue with the approach we can take, and I think that one isn't sufficient, we need to consider two of them:

1. individuals and SMBs are happy if we keep compatibility between N and N+1 and they can just do `yum|apt upgrade` from time to time, and nothing breaks. Before version N+2 comes out, they have upgraded their whole estate. They will not appreciate to have to install a new package just to upgrade their estate. They might have to change something in their behaviour before N+2, so the software itself should warn them about deprecation (that's something I just realised, i.e. here the old CLI should warn the user that it's deprecated), because as Frank wrote "who reads documentation?".
2. big estates are more difficult and I don't think, we (as in developer or packager) can directly address the issue, or we'll kill ourselves with requirements that were never meant to be addressed by packaging or programming. Support "forever" is just a bitch, and adding versioned packages on top of versioned packages will not help, because also packagers don't want to support old cruft forever. This said, the approach described above is in the hand of the service provider, he can just install as many versions of rdiff-backup as they want on their servers in as many virtualenvs as required. And if we introduce a %v placeholder, the instructions to the users could even be the same for all users without need to change the configuration as they upgrade their client rdiff-backup.

@ikus060 I assume that your customers trigger the backup? So _they_ call `rdiff-backup source user@server::/target` ? Meaning they could add the proper `--remote-schema /usr/local/lib/rdiff-backup-%v...` parameter (the other way around would be uglier as it would require you to know which client has which version installed, but it remains ugly whatever we decide).

The only downside of this approach is that it can't work with packages, which can be an issue in some corporations, but generally it shouldn't be a big issue because then rdiff-backup is at the core of the service delivered, and exceptions apply (I assume that service providers don't care).

ericzolf:
So, for me, we have 3 different types of users:

1. individuals: they have one or two desktops to backup, possibly not even remote back-up, only local to an external disk. I would assume that they mainly use the package of their distro (if Linux). For them we need to make sure that each update/upgrade is rather transparent, but they can upgrade their environment more or less at the same time.
2. [SMB](https://en.wikipedia.org/wiki/Small_and_medium-sized_enterprises) & similar: they will also use the packaged version of their distro, but they have a rather small number of servers, which they have under control, and can upgrade in finite time. They still need some time to upgrade their whole environment.
3. big corporations and service providers (like Patrik): they have many hosts which they don't have under control, some of them very old, and they can only upgrade very slowly, sometimes only once all the old cruft died out (I personally still have customers with RHEL5 :roll_eyes:). The clients might or might not be packaged (for sake of simplicity, I consider "wheels" as not packaged), it's not under their control, but they need to support them all on their backup server(s). I assume that they have the choice if they install via package or wheel/code, whatever, on their servers.

Let me know first if you agree with the above analysis.

Now I continue with the approach we can take, and I think that one isn't sufficient, we need to consider two of them:

1. individuals and SMBs are happy if we keep compatibility between N and N+1 and they can just do `yum|apt upgrade` from time to time, and nothing breaks. Before version N+2 comes out, they have upgraded their whole estate. They will not appreciate to have to install a new package just to upgrade their estate. They might have to change something in their behaviour before N+2, so the software itself should warn them about deprecation (that's something I just realised, i.e. here the old CLI should warn the user that it's deprecated), because as Frank wrote "who reads documentation?".
2. big estates are more difficult and I don't think, we (as in developer or packager) can directly address the issue, or we'll kill ourselves with requirements that were never meant to be addressed by packaging or programming. Support "forever" is just a bitch, and adding versioned packages on top of versioned packages will not help, because also packagers don't want to support old cruft forever. This said, the approach described above is in the hand of the service provider, he can just install as many versions of rdiff-backup as they want on their servers in as many virtualenvs as required. And if we introduce a %v placeholder, the instructions to the users could even be the same for all users without need to change the configuration as they upgrade their client rdiff-backup.

@ikus060 I assume that your customers trigger the backup? So _they_ call `rdiff-backup source user@server::/target` ? Meaning they could add the proper `--remote-schema /usr/local/lib/rdiff-backup-%v...` parameter (the other way around would be uglier as it would require you to know which client has which version installed, but it remains ugly whatever we decide).

The only downside of this approach is that it can't work with packages, which can be an issue in some corporations, but generally it shouldn't be a big issue because then rdiff-backup is at the core of the service delivered, and exceptions apply (I assume that service providers don't care).

ericzolf:
So, for me, we have 3 different types of users:

1. individuals: they have one or two desktops to backup, possibly not even remote back-up, only local to an external disk. I would assume that they mainly use the package of their distro (if Linux). For them we need to make sure that each update/upgrade is rather transparent, but they can upgrade their environment more or less at the same time.
2. [SMB](https://en.wikipedia.org/wiki/Small_and_medium-sized_enterprises) & similar: they will also use the packaged version of their distro, but they have a rather small number of servers, which they have under control, and can upgrade in finite time. They still need some time to upgrade their whole environment.
3. big corporations and service providers (like Patrik): they have many hosts which they don't have under control, some of them very old, and they can only upgrade very slowly, sometimes only once all the old cruft died out (I personally still have customers with RHEL5 :roll_eyes:). The clients might or might not be packaged (for sake of simplicity, I consider "wheels" as not packaged), it's not under their control, but they need to support them all on their backup server(s). I assume that they have the choice if they install via package or wheel/code, whatever, on their servers.

Let me know first if you agree with the above analysis.

Now I continue with the approach we can take, and I think that one isn't sufficient, we need to consider two of them:

1. individuals and SMBs are happy if we keep compatibility between N and N+1 and they can just do `yum|apt upgrade` from time to time, and nothing breaks. Before version N+2 comes out, they have upgraded their whole estate. They will not appreciate to have to install a new package just to upgrade their estate. They might have to change something in their behaviour before N+2, so the software itself should warn them about deprecation (that's something I just realised, i.e. here the old CLI should warn the user that it's deprecated), because as Frank wrote "who reads documentation?".
2. big estates are more difficult and I don't think, we (as in developer or packager) can directly address the issue, or we'll kill ourselves with requirements that were never meant to be addressed by packaging or programming. Support "forever" is just a bitch, and adding versioned packages on top of versioned packages will not help, because also packagers don't want to support old cruft forever. This said, the approach described above is in the hand of the service provider, he can just install as many versions of rdiff-backup as they want on their servers in as many virtualenvs as required. And if we introduce a %v placeholder, the instructions to the users could even be the same for all users without need to change the configuration as they upgrade their client rdiff-backup.

@ikus060 I assume that your customers trigger the backup? So _they_ call `rdiff-backup source user@server::/target` ? Meaning they could add the proper `--remote-schema /usr/local/lib/rdiff-backup-%v...` parameter (the other way around would be uglier as it would require you to know which client has which version installed, but it remains ugly whatever we decide).

The only downside of this approach is that it can't work with packages, which can be an issue in some corporations, but generally it shouldn't be a big issue because then rdiff-backup is at the core of the service delivered, and exceptions apply (I assume that service providers don't care).

frankcrawford:
Eric,
I think your argument is reasonable, and probably accurate.  The one item that does need to added is what sort of timeframe you are thinking of between N and N+1.  There is a very big difference between a week and 6 months.

ericzolf:
Difficult to foresee, but given our/my current speed, one week isn't a danger we need to consider. I would actually say that we're more at 6 to 12 months.

frankcrawford:
> Difficult to foresee, but given our/my current speed, one week isn't a danger we need to consider. I would actually say that we're more at 6 to 12 months.

Yes, that is what I feel, and it is probably a good target considering what type of software this is.

ericzolf:
I added the deprecation message as announced. @ikus060 your decision!

ericzolf:
OK, I psuhed the necessary changes for the man page to describe the few changes necessary to the old CLI. I plan to document the new CLI in a separate man page, using markdown as planned for a long time, but it'll be a separate PR, once this one is merged.

ikus060:
If I'm thinking only about my need, I will manage to keep multiple version of rdiff-backup "forever" under Minarca umbrella. I've already started packaging rdiff-backup v1.2.8 and rdiff-backup v2 within Minarca to be backward compatible. Using virtual environment...

That said, let go along and move forward with breaking changes and will see how the community react. Let not accept status quo.

Plz plz, just make sure we change the version accordingly for each breaking changes.

ericzolf:
OK, thanks. We have anyway time until we release our next version, before then we can reconsider and tune our approach.

PR #518, opened 2021-01-17
[DOC] Fix man-page to clarify format of locations

### LINKED ISSUE #480: man page is wrong and remote host is mandatory before double column
I would say that the man page is wrong and not the code: nothing in the rest of the man page hints at the host being optional (and neither in the code to be clear), and the double columns isn't mandatory, as the man page would hint. Furthermore, what is actually the meaning of `::/some/dir`, localhost via SSH, or simply the same as `/some/dir`? Long story short, I would fix the man page to `rdiff-backup [options] [[[user@]host1.foo::]source_directory] [[[user@]host2.foo::]destination_directory]` (shifting the double-columns into the optional part). Fair enough?

_Originally posted by @ericzolf in https://github.com/rdiff-backup/rdiff-backup/pull/404#issuecomment-716117582_

ericzolf:
DOC: clarify that the host part belongs together with the double colons, closes #480

### DIFF
--- a/docs/rdiff-backup.1
+++ b/docs/rdiff-backup.1
@@ -3,9 +3,9 @@
 rdiff-backup \- local/remote mirror and incremental backup
 .SH SYNOPSIS
 .B rdiff-backup
-.BI [ options ]
-.BI [[[ user@ ] host1.foo ]:: source_directory ]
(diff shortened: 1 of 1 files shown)


ericzolf:
I allow myself to merge without review if nobody volunteers, it's only documentation.

PR #524, opened 2021-02-06
Add version to schema and use format

ericzolf:
As discussed in the previous PR #517, there is a now a placeholder for the version with `{vx}`, `{vy}` and `{vz}`, so that the user can point at a specific version of rdiff-backup corresponding to the current version (e.g. using virtualenvs), and the host placeholder has become `{h}` for consistency.

### DIFF
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -213,6 +213,9 @@ def _fill_schema(host_info):
     """
     Fills host_info and optionally the version into the schema and returns remote command.
     """
+    assert isinstance(host_info, bytes), (
+        "host_info parameter must be bytes not {thi}.".format(
+            thi=type(host_info)))
     try:
         # for security reasons, we accept only specific format placeholders
         # h for host_info, vx,vy,vz for version x.y.z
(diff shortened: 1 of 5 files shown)


src/rdiff_backup/SetConnections.py
```diff
@@ -210,11 +210,32 @@ def _parse_file_desc(file_desc):
...
+                    and b"%s" not in __cmd_schema)):  # compat200
+            raise KeyError
+        if b"{h}" in __cmd_schema:
+            ver_split = Globals.version.split(".")
+            # bytes doesn't have a format method, hence the conversions
+            return os.fsencode(os.fsdecode(__cmd_schema).format(
+                h=os.fsdecode(host_info),
+                vx=ver_split[0], vy=ver_split[1], vz=ver_split[2]))
+        else:  # compat200: accepts "%s" as host place-holder
+            return __cmd_schema % host_info
```
ikus060, line 235: Shoudn't we keep a test to make sure '%s' is still working with this version ?

ericzolf: Good point, I reverted one of the places where `%s` was used and added a `compat200` marker.

ikus060 (approved):

PR #525, opened 2021-02-07
Split actions plugins and improve actions documentation

ericzolf:
As the subject says, it's just re-shuffling of code in accordance with the actions plug-in interface, which I've also started to define.

I've also added a fair amount of comments, which further led me to create new coding standards.

As I worked through all this, I realized that testing servers and verifying hashes has nothing to do with each other so I split the two actions again.


docs/CODING.md
```diff
@@ -73,3 +73,23 @@ and reviewers. This document lists such coding conventions.
   few reasons to use asserts).
 * Any conversion between strings and bytes happens with `os.fsencode/fsdecode`.
 * Messages are to be expressed in strings (log, errors, etc).
+
+## DOCSTRINGS
```
ikus060, line 77: Is it possible to enforce this rules using flake8 ? That would greatly simplify the review process as I'm not the best to follow rules :D

docs/CODING.md
```diff
@@ -73,3 +73,23 @@ and reviewers. This document lists such coding conventions.
...
+
+* docstrings are quoted with three double quotes (see above).
+* each triple quotes are alone on their line (this is a deviation from some
+  recommendations) so that it is readable.
+* each docstring starts with a single line summarizing the purpose of the
+  object.
+* further descriptions are written after a blank line.
+
+## IMPORT
```
ikus060, line 86: Same here we should consider adding flake8

docs/arch/plugins_actions.md
```diff
@@ -0,0 +1,53 @@
...
+2. or a top-level module with a name `rdb_action_[...].py`
+
+All those plug-ins are found by the action manager module `rdiffbackup.actions_mgr`
+and returned as a hash made of key-value pairs by the function `get_discovered_actions`:
+
+* key is the name returned by the Action class (see below the `get_name` method)
+* value is the Action class itself, as returned by the `get_action_class` function
+  of the action plug-in module.
+
+> **NOTE:** more information is provided with `pydoc rdiffbackup.actions_mgr`.
```
ikus060, line 18: I feel we are re-inventing the wheel. Should we consider using setuptools entry-point instead ? setup-tool entry point allow quick extension in a pure python environment with a standard approach.

It's very declarative too. 

If we use setup-tools, we may strip down the code you made. Meaning let code to maintain.

docs/arch/plugins_actions.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Module interface
```
ikus060, line 20: Would this approach allow to integrate rdiff-backup-delete into a plugin ?

ikus060 (commented):
It's clearly nicer to review each "action" in seperate file ! Check my comments

docs/CODING.md
ericzolf: See above and #530

docs/CODING.md
ericzolf: Not a small task, I have no clue if flake8 can do it, and how to configure it. But interesting idea, I've parked it in an enhancement request #530 (perhaps someone else could take it).

docs/arch/plugins_actions.md
ericzolf: My approach is actually coming directly from the Python packaging team, see https://packaging.python.org/guides/creating-and-discovering-plugins/, similar to the approach you suggest, listed as last option.

And let's be clear that the "entry point" is only the module itself; how the module internally looks like so that it can be used as plug-in still needs to be described/implemented.

Personally, I find that dropping a module named `rdb_action_whatever.py` in one's PYTHONPATH is a lot easier than any setup.py approach. We can still add the option if there is a demand later on (as said, it just defines how we find modules, not how we use them).

docs/arch/plugins_actions.md
ericzolf: That's the idea indeed, it was one of the drivers for this plug-in concept :-)

But don't start yet, I expect the interface to need much extension before it can be done properly (as hinted in the docs). It's only a first small step.

ericzolf: This said, my idea was to add a sub-action to the remove action, so it would be implemented as `remove files` vs. currently only `remove increments`. It would hence not be really a plug-in but the fact that plug-ins are possible would make it easier to implement properly (so the plan).
But, anyway, a user could implement it as `delete-files` plug-in and it would work as well.

docs/CODING.md
ikus060: +1 I saw the task

docs/arch/plugins_actions.md
ikus060: I've raise the idea so we could consider using it. I don't mind keeping your implementation.

docs/arch/plugins_actions.md
ikus060: Yeah !

ikus060 (approved):

PR #528, opened 2021-02-13
Man page for new CLI of rdiff-backup in markdown format

ericzolf:
This PR is best described with the following two lines:

* man page can be generated from markdown using pandoc, which is now a dependency
* man page **rdiff-backup-old**(1) describes the old CLI, **rdiff-backup**(1) the new one


tools/build_wheels.sh
```diff
@@ -10,7 +10,7 @@ build_dir=${basedir}/build
 dist_dir=${basedir}/dist
 
 # Install a system package required by our library
-yum install -y librsync-devel
+yum install -y librsync-devel pandoc
```
ikus060, line 13: Have you consider adding `pandoc` to `setup_requires` instead of adding it to `yum install` ?

ikus060 (changes requested):
Could you explain why we are not adding pandoc as python dependencies instead of adding it to apt, yum, etc.
If we are adding it to setup.py, it should simplify the step to build rdiff-backup.

tools/build_wheels.sh
ericzolf: I guess, you're referring to the fact that there is a pandoc library on PyPI. I didn't know this but it depends on the pandoc command line, so I would anyway need to install it.

ikus060: When you install pandoc from pypi, it should install the command line too.

ikus060: I feel, it's more portable if the dependencies is defined in setup.py since it's gonna be fetch automatically. Otherwise, we need to update all the documentation everywhere to include pandoc.

Another alternative is to make the man page creation optional during build.

ericzolf: So, I checked the documentation at https://github.com/boisgera/pandoc#getting-started and it doesn't install automatically the pandoc binary. To be absolutely sure, I tried it myself and it really doesn't  (also how, it could be installed via dnf, yum, chocolatey, etc...), it installs but when tried, it complains about the missing binary in my `$PATH`. As a side note, the stable version is broken, I had to use the alpha version (which is as old as April 2020, not encouraging).

Make the creation of manpages optional: I wouldn't know how to do this! The documentation to extend setuptools is "lean" (to say the least), but/and I haven't seen anything of this order. This said (and that's rather a bug, but well), the way I've implemented it doesn't currently make the build process fail (it'll fail at "dist" time, so I don't care too much for our pipeline).

ikus060: Too bad. Thanks for trying. It would have been more portable.

ikus060 (approved):

PR #529, opened 2021-02-13
[DOC] Describe side-by-side installation and clean-up migration.md

### LINKED ISSUE #523: Document parallel usage of rdiff-backup versions and different use cases
1. for the "normal" use cases, we still think to have one compatible version in-between just because we don't want people to upgrade all at once, mainly because if they take the packaged version from their distro, they can't decide when they get the new version, so Debian might be one week later than Fedora or vice-versa, and we don't want people to stumble about those things. This should keep these people happy and it is important, IMHO. The measures taken (API and compat200) allow me to do this with reasonable effort, at the price of a _temporary_ complexity of the code.
2. for the few people with special needs, the solution is actually very simple. Use virtualenvs, something like:

```bash
virtualenv $BASEDIR/rdiff-backup-2.0
$BASEDIR/rdiff-backup-2.0/bin/pip install rdiff-backup==2.0.5
$BASEDIR/rdiff-backup-2.0/bin/rdiff-backup --version  # result is 2.0.5
```
The user then only needs to use the `--remote-schema` option to point at the right version. There are even variations, like each user having a different PATH pointing at the right virtualenv preferably. And I could add a %v placeholder to simplify the use of the `--remote-schema` (it would only be the first or 2 first places of the version, e.g. 2 or 2.0).

So, for me, we have 3 different types of users:

1. individuals: they have one or two desktops to backup, possibly not even remote back-up, only local to an external disk. I would assume that they mainly use the package of their distro (if Linux). For them we need to make sure that each update/upgrade is rather transparent, but they can upgrade their environment more or less at the same time.
2. [SMB](https://en.wikipedia.org/wiki/Small_and_medium-sized_enterprises) & similar: they will also use the packaged version of their distro, but they have a rather small number of servers, which they have under control, and can upgrade in finite time. They still need some time to upgrade their whole environment.
3. big corporations and service providers (like Patrik): they have many hosts which they don't have under control, some of them very old, and they can only upgrade very slowly, sometimes only once all the old cruft died out (I personally still have customers with RHEL5 :roll_eyes:). The clients might or might not be packaged (for sake of simplicity, I consider "wheels" as not packaged), it's not under their control, but they need to support them all on their backup server(s). I assume that they have the choice if they install via package or wheel/code, whatever, on their servers.

Let me know first if you agree with the above analysis.

Now I continue with the approach we can take, and I think that one isn't sufficient, we need to consider two of them:

1. individuals and SMBs are happy if we keep compatibility between N and N+1 and they can just do `yum|apt upgrade` from time to time, and nothing breaks. Before version N+2 comes out, they have upgraded their whole estate. They will not appreciate to have to install a new package just to upgrade their estate. They might have to change something in their behaviour before N+2, so the software itself should warn them about deprecation (that's something I just realised, i.e. here the old CLI should warn the user that it's deprecated), because as Frank wrote "who reads documentation?".
2. big estates are more difficult and I don't think, we (as in developer or packager) can directly address the issue, or we'll kill ourselves with requirements that were never meant to be addressed by packaging or programming. Support "forever" is just a bitch, and adding versioned packages on top of versioned packages will not help, because also packagers don't want to support old cruft forever. This said, the approach described above is in the hand of the service provider, he can just install as many versions of rdiff-backup as they want on their servers in as many virtualenvs as required. And if we introduce a %v placeholder, the instructions to the users could even be the same for all users without need to change the configuration as they upgrade their client rdiff-backup.

@ikus060 I assume that your customers trigger the backup? So _they_ call `rdiff-backup source user@server::/target` ? Meaning they could add the proper `--remote-schema /usr/local/lib/rdiff-backup-%v...` parameter (the other way around would be uglier as it would require you to know which client has which version installed, but it remains ugly whatever we decide).

The only downside of this approach is that it can't work with packages, which can be an issue in some corporations, but generally it shouldn't be a big issue because then rdiff-backup is at the core of the service delivered, and exceptions apply (I assume that service providers don't care).

_Originally posted by @ericzolf in https://github.com/rdiff-backup/rdiff-backup/issues/517#issuecomment-770180528_

ericzolf:
DOC: docs/migration.md describes how to install rdiff-backup side by side and use old versions 'forever', closes #523

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -1,20 +1,18 @@
+# Migration and Side-by-Side installation
 
+This document is a guide to help you migrate from one version of rdiff-backup
+to the next, when it comes to breach of compatibility.
 
-Migration from v1.2.8 to v2.0.0
--------------------------------
+Alternatively it explains how to keep multiple versions of rdiff-backup
+installed in parallel on the same server.
 
-  
+## Migration from v1.2.8 to v2.0.0
 
-This document is a guide to help you migrate from rdiff-backup v1.2.8
-(legacy version) to rdiff-backup v2.0.0 (new version).
-
-  
-
-This new version of rdiff-backup is compatible with repositories created
-with the legacy version. In other words, the data on the disk is fully
+The new version 2.x of rdiff-backup is compatible with repositories created
+with the legacy version 1.x. In other words, the data on the disk is fully
 compatible between the two.
 
-**The network protocol of the legacy version is NOT compatible with the new version.**
+> **CAUTION:** **The network protocol of the legacy version is NOT compatible with the new version.**
 In other words, you must be running the same version of rdiff-backup locally
 and remotely. For this reason, you might need to have a plan to transition
 from the legacy version to the new version depending of your use case.
@@ -37,8 +35,7 @@ same time might not be possible. The following section describes a way
 for you to mitigate this problem. By installing both versions side by
 side. We recommend installing the new rdiff-backup version in a virtualenv.
 
-Use Case: Local to Local
-------------------------
+### Use Case: Local to Local
 
 You are using rdiff-backup locally if you are running a command line
 where the source and the destination are defined as a path on the same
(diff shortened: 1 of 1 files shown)


guysoft:
Seems legit, adds another section on how to use virtualenv and a few phrasing fixes.

ikus060:
I love it. 

Let hope it gonna help people.

Side note, we should probably clean-up the website documentation that get more and more obsolete and point the user toward the git repo documentation. If you agreed, I will make the change.

ikus060 (approved):

ericzolf:
Thanks @guysoft @ikus060 I've merged it.

PR #535, opened 2021-02-25
Define fully action plugins interface and implement most of it

ericzolf:
As the subject says, this PR is a further step towards having proper action plugins.
The architecture documentation has been completed, and the implementation has been progressed.
I still reserve the implementation of `action.run()` for a further PR because it goes quite deeper into the current code, and I have reached a state which passes the tests, so I'd like it committed to master before I continue.

As a side note, I also introduced more understandable log levels, deprecating some of the old ones, which (I think) were too many to keep in mind (I'll formalize this once I'll properly define the log -plugin- interface):
    ERROR = 1
    WARNING = 2
    DEFAULT = 3
    INFO = 5
    DEBUG = 9

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -55,12 +55,12 @@ rdiff-backup FAQ
       1   Fatal Errors displayed
       2   Warnings
       3   Important messages, and maybe later some global statistics (default)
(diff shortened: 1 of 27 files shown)


ericzolf:
Merging without review...

PR #539, opened 2021-02-28
[DOC] Windows build: 32/64 and linker error explained

rstarkov:
I followed the build instructions and got a difficult to diagnose error. I've explained the cause in these instructions.

I've also modified the instructions to default to the 64 bit build, as that is the recommended python download and is probably what most people have I suspect.


ericzolf:
Thanks, Roman, looks like a meaningful extension. Can you also extend accordingly the Python 3 chapter? It explicitly only lists the 32 bits version currently, which wouldn't be consistent anymore with your changes.
And, if you don't mind the additional effort to test, it would be nice to bump the Python version to 3.9, which is now the default version for the pipeline (my fault, I oversaw the document as I bumped the version).

rstarkov:
Good idea, this will take some time but I'll do it.

ericzolf:
Any news?

rstarkov:
@ericzolf thanks for your patience with me. I've re-tested everything on python 3.9 and updated obsolete instructions in the python 3 chapter.

ericzolf:
Looks good to me, thanks for not giving up on my (hopefully reasonable :smirk: ) request!

PR #544, opened 2021-03-10
Change provision playbook to install Asciidoctor under Windows

ericzolf:
Just a small PR to install asciidoctor under Windows (I didn't yet change the pipeline as it might break it, but it can be used as blueprint once we're so far).


ericzolf:
It doesn't really interest anybody but me, so merged without review...

ikus060:
I'm not sure any body uses this playbook except you. 😅

ericzolf:
But it isn't the case of #535 ...

PR #545, opened 2021-03-19
When using the --remove-older-than option with --tempdir, the --tempdir flag is not honored

averi:
When using the --remove-older-than option with --tempdir, the --tempdir flag is not honored causing it to default to $BACKUPS_PATH. That's fine until the disk hosting the backups goes out of space, make sure the --tempdir flag is really taken into account when used.

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -993,6 +993,7 @@ class RPath(RORPath):
 
         # we have to create our own "uniqueness" as tempfile.mktemp is
         # obsolete and we just want a name agnostic regarding file vs. dir
+        import tempfile
         while True:
             if self.__class__._temp_file_index > 100000000:
                 log.Log("Warning: Resetting tempfile index", 2)
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -1001,9 +1001,9 @@ class RPath(RORPath):
 
             # here to catch the bytes string passed via cli opts
             if type(tempfile.tempdir) == bytes:
-              tempdir = tempfile.tempdir.decode()
+                tempdir = tempfile.tempdir.decode()
             else:
-              tempdir = tempfile.tempdir
+                tempdir = tempfile.tempdir
 
             tf = self.append(os.path.join(tempdir, 'rdiff-backup.tmp.{index:d}'.format(
                              index=self.__class__._temp_file_index)))
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -1001,7 +1001,7 @@ class RPath(RORPath):
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
 
-            if tempfile.tempdir and (shutil.disk_usage(self).free == 0):
+            if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
                 # here to catch the bytes string passed via cli opts
                 if type(tempfile.tempdir) == bytes:
                     tempdir = tempfile.tempdir.decode()
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -997,8 +997,15 @@ class RPath(RORPath):
             if self.__class__._temp_file_index > 100000000:
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
-            tf = self.append('rdiff-backup.tmp.{index:d}'.format(
-                             index=self.__class__._temp_file_index))
+
+            # here to catch the bytes string passed via cli opts
+            if type(tempfile.tempdir) == bytes:
+              tempdir = tempfile.tempdir.decode()
+            else:
+              tempdir = tempfile.tempdir
+
+            tf = self.append(os.path.join(tempdir, 'rdiff-backup.tmp.{index:d}'.format(
+                             index=self.__class__._temp_file_index)))
             self.__class__._temp_file_index += 1
             if not tf.lstat():
                 return tf
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -1000,21 +1000,25 @@ class RPath(RORPath):
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
 
+            # When the file system hosting the rdiff-backup-data directory
+            # is full and when the --tempdir flag is defined, attempt to save
+            # temporary files on a different path / file system.
             if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
-                # here to catch the bytes string passed via cli opts
+                # If tempfile.tempdir is manually passed in via the --tempdir
+                # cli flag, it defaults being a bytes string, as such we need to
+                # convert the target path to a string first
                 if type(tempfile.tempdir) == bytes:
-                    tempdir = tempfile.tempdir.decode()
+                    tempdir = os.fsdecode(tempfile.tempdir)
                 else:
                     tempdir = tempfile.tempdir
             else:
                 tempdir = None
 
+            _tf = 'rdiff-backup.tmp.{index:d}'.format(index=self.__class__._temp_file_index)
             if not tempdir:
-                tf = self.append('rdiff-backup.tmp.{index:d}'.format(
-                                 index=self.__class__._temp_file_index))
+                tf = self.append(_tf)
             else:
-                tf = self.append(os.path.join(tempdir, 'rdiff-backup.tmp.{index:d}'.format(
-                                 index=self.__class__._temp_file_index)))
+                tf = os.path.join(tempdir, _tf)
             self.__class__._temp_file_index += 1
             if not tf.lstat():
                 return tf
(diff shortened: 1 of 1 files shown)


src/rdiff_backup/rpath.py
```diff
@@ -982,6 +982,9 @@ def new_index_empty(self, index):
         return self.__class__(self.conn, self.base, index, {'type': None})
 
     def get_temp_rpath(self, sibling=False):
+        import tempfile
```
ericzolf, line 985: No import in functions please, always import at the top of the file, in alphabetical order is better (I know, the current code doesn't do this, but it should, feel free to fix at the same time).
Reference: https://www.python.org/dev/peps/pep-0008/#id23

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +1000,22 @@ def get_temp_rpath(self, sibling=False):
             if self.__class__._temp_file_index > 100000000:
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
-            tf = self.append('rdiff-backup.tmp.{index:d}'.format(
-                             index=self.__class__._temp_file_index))
+
+            if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
+                # here to catch the bytes string passed via cli opts
+                if type(tempfile.tempdir) == bytes:
```
ericzolf, line 1006: You can simplify those lines to `tempdir = os.fsdecode(tempfile.tempdir)` and this is the (re)commended approach.

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +1000,22 @@ def get_temp_rpath(self, sibling=False):
             if self.__class__._temp_file_index > 100000000:
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
-            tf = self.append('rdiff-backup.tmp.{index:d}'.format(
-                             index=self.__class__._temp_file_index))
+
+            if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
+                # here to catch the bytes string passed via cli opts
```
ericzolf, line 1005: Since recently, we don't save `tempfile.tempdir` as bytes anymore because the tempfile module struggles with it (see comment in Main.py, there is  a bug for this).

I'm hesitating if I should tell you to not do the conversion, at the risk to hit a bug once the issue is solved, so leave it as it is. But modify the comment to clarify the purpose.

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +1000,22 @@ def get_temp_rpath(self, sibling=False):
...
+                if type(tempfile.tempdir) == bytes:
+                    tempdir = tempfile.tempdir.decode()
+                else:
+                    tempdir = tempfile.tempdir
+            else:
+                tempdir = None
+
+            if not tempdir:
+                tf = self.append('rdiff-backup.tmp.{index:d}'.format(
```
ericzolf, line 1014: construct the format first in a temporary variable, then use it alone or jointly with tempfile.tempdir, it avoids duplicate code.

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +1000,22 @@ def get_temp_rpath(self, sibling=False):
...
+                                 index=self.__class__._temp_file_index))
+            else:
+                tf = self.append(os.path.join(tempdir, 'rdiff-backup.tmp.{index:d}'.format(
```
ericzolf, line 1017: This line sounds really wrong: if tempdir is `/var/tmp`, this line will _append_ it to the current path, i.e. `/mnt/mybackup` would then give `/var/tmp/mnt/mybackup/rdiff-backup.tmp.123`. This is surely not what you want.

Rather create a completely new RPath object, probably something like `tf = RPath(self.conn, tempfile.tempdir, (indexed_file_name))`. Notice that I assume `indexed_file_name = os.encode("rdiff-backup.tmp.{index:d}".format(index=self.__class__._temp_file_index)))` (it must be bytes, the type of tempfile.tempdir becomes irrelevant).

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +1000,22 @@ def get_temp_rpath(self, sibling=False):
             if self.__class__._temp_file_index > 100000000:
                 log.Log("Warning: Resetting tempfile index", 2)
                 self.__class__._temp_file_index = 0
-            tf = self.append('rdiff-backup.tmp.{index:d}'.format(
-                             index=self.__class__._temp_file_index))
+
+            if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
```
ericzolf, line 1004: Add a comment what you're trying to do, i.e. use the defined tempdir if there is no place left on the current file system (it assumes that the file system of the temporary directory is a different one, it should be checked that it is the case, at least that it isn't full as well).

ericzolf (changes requested):
A lot of comment for cleaner code, sorry, but the code is complicated enough like it is.

There is nevertheless only one error to correct absolutely, and two questions to answer:
1. does this really work if the tempdir is on a different file system from the repository, and if the repository file system is completely full?
2. is there a noticeable performance impact because more checks are done for each file, even if there is no tempdir defined? Note that with a tempdir on a different file systems, I expect the performance to go down a lot because each file will be copied/deleted from one file system to the next instead of simply being renamed/moved within one file system.

Please test this thoroughly and let us know.

ericzolf:
Sorry, I can be a pain in the bottom when it comes to reviewing, which doesn't mean that I don't appreciate the effort, I just don't say it :stuck_out_tongue_winking_eye:

src/rdiff_backup/rpath.py
averi: Addressed by https://github.com/rdiff-backup/rdiff-backup/pull/545/commits/fab0481dd833a1f67fab980f671d2e9861c8b747.

src/rdiff_backup/rpath.py
averi: Addressed by https://github.com/rdiff-backup/rdiff-backup/pull/545/commits/d48390b80cf4de90f79926b8e3a3d4cdd4a2efea.

src/rdiff_backup/rpath.py
averi: Addressed by https://github.com/rdiff-backup/rdiff-backup/pull/545/commits/d48390b80cf4de90f79926b8e3a3d4cdd4a2efea.

src/rdiff_backup/rpath.py
averi: Instead of building a completely new RPath object, wouldn't it be enough to avoid appending and let tf be composed of the two joint directories? Went that way with https://github.com/rdiff-backup/rdiff-backup/pull/545/commits/d48390b80cf4de90f79926b8e3a3d4cdd4a2efea, please let me know if that's not desired.

src/rdiff_backup/rpath.py
averi: Do we really want to introduce more checks under that while loop? If the user picks a directory that is out of disk space it'll get an exception later on in the code mentioning this. That looks enough to me as the OS exception is pretty clear. Adding more logic here which involves reading the partitions table etc sounds overkill to me.

averi:
> 1. does this really work if the tempdir is on a different file system from the repository, and if the repository file system is completely full?

I tested my changes on that specific scenario, we had an empty network storage we couldn't extend more and free space under a different path, rdiff-backup allowed me to cleanup older increments via --remove-older-than with the above patch.

> 2. is there a noticeable performance impact because more checks are done for each file, even if there is no tempdir defined? Note that with a tempdir on a different file systems, I expect the performance to go down a lot because each file will be copied/deleted from one file system to the next instead of simply being renamed/moved within one file system.

I honestly haven't noticed any performance impact against backup stores of several hundred GBs, the way I see passing the --tempdir flag is when there's an absolute need of doing so, i.e backup partition is out of disk and I really need to clean up older increments for subsequent backup runs to succeed. I honestly never used --tempdir for any other case than this as the default (use the same partition as the backup store for tmp files) works as expected (until you have a full disk :))

src/rdiff_backup/rpath.py
```diff
@@ -997,8 +999,26 @@ def get_temp_rpath(self, sibling=False):
...
-                             index=self.__class__._temp_file_index))
+
+            # When the file system hosting the rdiff-backup-data directory
+            # is full and when the --tempdir flag is defined, attempt to save
+            # temporary files on a different path / file system.
+            if tempfile.tempdir and (shutil.disk_usage(self.path).free == 0):
+                # If tempfile.tempdir is manually passed in via the --tempdir
+                # cli flag, it defaults being a bytes string, as such we need to
+                # convert the target path to a string first
+                if type(tempfile.tempdir) == bytes:
```
ericzolf, line 1010: Sorry, if I wasn't clear but you don't need to check for bytes vs. str, `os.fsdecode` does it for you, so you can replace the four lines just with one `tempdir = os.fsdecode(tempfile.tempdir)` and it will work even if `tempfile.tempdir` is already a string. See https://docs.python.org/3/library/os.html#os.fsdecode

ericzolf (changes requested):
Just one last small change, else it looks good to me and I'll merge it once done. Thanks for the effort!

src/rdiff_backup/rpath.py
averi: For a moment I missed the fact os.fsdecode() would cope just fine with both a byte string and a string type, defaulting to return a string, change made, thanks!

ericzolf (approved):

ericzolf:
Thanks for the effort, the review was a bit bumpy but we got there at last :smile:

PR #550, opened 2021-04-06
Refactor Main.py into the actions' run() methods

ericzolf:
It's a rather huge PR but as all the actions were entangled into the Main.py module, it was more or less impossible to split them one by one, so I had to split them all at once, refactoring them into the run() method of the new actions. There is still potential for improvement but at least, it's clear now which action belongs where.

At the same time, I had to create a new "locations" kind of object, which represent either a directory or a (backup) repository. It wasn't the main purpose of the PR but a first logical step.

This aspect will need to be revisited later on though, as it's kind of duplicate with the *Struct classes present in `rdiff_backup/backup.py` and `rdiff_backup/restore.py`, and of course, those action-looking-modules should also be merged with the new action plug-ins. Where it gets messy, is that those aspects are more often used in client/server mode, hence will impact the API. But this is for one or more follow-up PR(s)...

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -54,13 +54,13 @@ rdiff-backup FAQ
       0   No information given
       1   Fatal Errors displayed
       2   Warnings
-      3   Important messages, and maybe later some global statistics (default)
+      3   Important notes, and maybe later some global statistics (default)
       4   Some global settings, miscellaneous messages (obsolete)
-      5   Mentions which files were changed
+      5   Mentions which files were changed, and other info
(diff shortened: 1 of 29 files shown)


ikus060:
I've quickly scroll over the changes. This PR is way to big for me to review carefully. At this point with all the changes you made recently, the real test will be the release of alpha version. :/

ericzolf:
Yeah, I know, hopefully it'll get better as the code stabilizes but it'll still take a while as I see it for now...

PR #561, opened 2021-04-12
Make rpath easier to handle as path or str

### LINKED ISSUE #84: [ENH] Make the RPath object a path-like object
<a href="https://github.com/ericzolf"></img></a> **Issue by [ericzolf](https://github.com/ericzolf)**
_Sunday Jul 14, 2019 at 12:07 GMT_
_Originally opened as https://github.com/ericzolf/rdiff-backup/issues/34_

----

https://docs.python.org/3/library/os.html#os.PathLike - it seems it would mean just implementing an __fspath__() function returning the path of the file as a string (or bytes). It would allow to make usage slightly more transparent but would also mean dropping support for Python 3.5.

ericzolf:
DEV: RORPath and RPath classes are now PathAlike and can be safely output as string, closes #84
This meant also getting rid of the get_safepath and get_safeindex
functions, and refactoring a lot of log messages.

### DIFF
--- a/src/rdiff-backup-statistics
+++ b/src/rdiff-backup-statistics
@@ -77,7 +77,7 @@ def parse_args():
         os.path.join(os.fsencode(args[0]), b"rdiff-backup-data"),
     )
     if not Globals.rbdir.isdir():
-        sys.exit("Directory %s not found" % (Globals.rbdir.get_safepath(),))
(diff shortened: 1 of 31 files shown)


ikus060:
Yes !

From the start, I was thinking of this. Good job !

ericzolf:
Thanks for the quick review, I'll leave it open for now, just to not change the lines reference compared to the feedback on the alpha version.

PR #562, opened 2021-04-14
rdiff-backup-delete permissions fixes

ikus060:
I've discovert two issues with rdiff-backup-delete lately.
1. Permissions error are raised when the directories has `r-x` This seams to happen often for windows to Linux backup
2. Need to support `win_access_control_lists`

### DIFF
--- a/src/rdiff-backup-delete
+++ b/src/rdiff-backup-delete
@@ -164,7 +164,8 @@ def _remove_from_metadata(repopath, file, dry_run):
             return line == b'File ' + repopath.metaquote + b'\n' or line.startswith(b'File ' + repopath.metaquote + b'/')
 
     elif (os.path.basename(file).startswith(b'extended_attributes')
-          or os.path.basename(file).startswith(b'access_control_lists')):
+          or os.path.basename(file).startswith(b'access_control_lists')
+          or os.path.basename(file).startswith(b'win_access_control_lists')):
         start_marker = b'# file: '
 
         def matches(line):
--- a/src/rdiff-backup-delete
+++ b/src/rdiff-backup-delete
@@ -226,23 +226,49 @@ def _remove_increments(path, dry_run):
 
 def _rmtree(path, dry_run):
     """
-    Custom implementation of shutil.rmtree() to support deletion of symlink.
+    Custom implementation of shutil.rmtree() to handle permission errors,
+    fifo and symlink deletion.
     """
+    if dry_run:
+        return
+
+    # Try to change the permissions of the file or directory to delete them.
+    def on_error(func, path, exc):
+        """
+        Handle permissions error while deleting file or directory by changing
+        the permissions to allow deletion.
+        """
+        # Parent directory must allow o+rwx
+        parent_dir = os.path.dirname(path)
+        if not os.access(parent_dir, os.W_OK | os.R_OK | os.X_OK):
+            prev_mode = os.stat(parent_dir).st_mode
+            os.chmod(parent_dir, prev_mode | stat.S_IRWXU)
+        if not os.access(path, os.W_OK | os.R_OK):
+            os.chmod(path, 0o0600)
+        return func(path)
+
     try:
         mode = os.lstat(path).st_mode
-    except os.error:
+    except FileNotFoundError:
+        # Nothing to delete
         mode = 0
+    except PermissionError as exc:
+        mode = on_error(os.lstat, path, exc).st_mode
     if stat.S_ISDIR(mode):
         names = []
         names = os.listdir(path)
         for name in names:
             fullname = os.path.join(path, name)
             _rmtree(fullname, dry_run)
-        if not dry_run:
+        try:
             os.rmdir(path)
+        except PermissionError as exc:
+            on_error(os.rmdir, path, exc)
     elif mode:
-        if not dry_run:
+        try:
             os.remove(path)
+        except PermissionError as exc:
+            on_error(os.remove, path, exc)
 
 
 def _unquote(name):
--- a/testing/rdiffbackupdeletetest.py
+++ b/testing/rdiffbackupdeletetest.py
@@ -43,8 +43,7 @@ class RdiffBackupDeleteTest(unittest.TestCase):
         # Copy the required repo to a temporary location.
         # We need to use os command line to properly copy and delete special files.
         self.repo = os.path.join(abs_test_dir, b'deletetest')
-        if os.path.exists(self.repo):
-            subprocess.check_call([b'rm', b'-Rf', self.repo])
+        self.assertFalse(os.path.exists(self.repo), 'repository should not already exists, previous test fail to clean-up. Run rm -Rf %s' % self.repo.decode('utf-8'))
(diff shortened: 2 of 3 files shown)


ikus060:
@ericzolf Any tips regarding why it's failing ? Cause everything is fine when I'm running this locally.

ericzolf:
I guess, you've found the issue in the meantime, else I don't understand your question? The PR is still marked WIP, so I don't review.

ikus060:
I've found part of the issue. I need to revert one of the unit tests to
make it work. I'm guessing GitHub tasks are not running with the same
permission then 'make test' Will debug this Friday.

ericzolf:
Sorry, I seem to be misunderstanding you: all the checks have passed, so where is the problem?

ikus060:
@ericzolf This is ready for review

src/rdiff-backup-delete
```diff
@@ -225,23 +226,45 @@ def _remove_increments(path, dry_run):
...
     """
+    if dry_run:
+        return
+
+    # Try to change the permissions of the file or directory to delete them.
+    def onerror(func, path, exc_info):
+        if exc_info[0] == PermissionError:
+            # Parent directory must allow rwx
+            if not os.access(os.path.dirname(path), os.W_OK | os.R_OK | os.X_OK):
+                os.chmod(os.path.dirname(path), 0o0700)
```
ericzolf, line 240: This would be less intrusive (the owner might want the group to still be able to read the directory):

```python
parent_dir = os.path.dirname(path)
st = os.stat(parent_dir)
os.chmod(parent_dir, st.st_mode | stat.S_IRWXU)
```
You could create a small helper function to do it, because you would need to do something similar for the file itself.

src/rdiff-backup-delete
```diff
@@ -225,23 +226,45 @@ def _remove_increments(path, dry_run):
...
+            if not os.access(path, os.W_OK | os.R_OK):
+                os.chmod(path, 0o0600)
```
ericzolf, line 242: Make it less intrusive as for the parent directory.

src/rdiff-backup-delete
```diff
@@ -225,23 +226,45 @@ def _remove_increments(path, dry_run):
...
+        raise
+
     try:
         mode = os.lstat(path).st_mode
-    except os.error:
+    except FileNotFoundError:
+        # Nothing to delete
         mode = 0
+    except OSError:
+        mode = onerror(os.lstat, path, sys.exc_info()).st_mode
```
ericzolf, line 252: the usage of `sys.exc_info()` is unusual, and IMHO unnecessary: you can just use `except OSError as exc:` in the line before, and use `isinstance(exc, PermissionError)` in the function.

src/rdiff-backup-delete
```diff
@@ -225,23 +226,45 @@ def _remove_increments(path, dry_run):
...
+            return func(path)
+        raise
+
     try:
         mode = os.lstat(path).st_mode
-    except os.error:
+    except FileNotFoundError:
+        # Nothing to delete
         mode = 0
+    except OSError:
```
ericzolf, line 251: why don't you directly `except PermissionError` and remove the check from the `on_error` function? Then you wouldn't even need to pass the exception to the function.

src/rdiff-backup-delete
```diff
@@ -225,23 +226,45 @@ def _remove_increments(path, dry_run):
...
+                os.chmod(os.path.dirname(path), 0o0700)
+            if not os.access(path, os.W_OK | os.R_OK):
+                os.chmod(path, 0o0600)
+            return func(path)
+        raise
+
     try:
         mode = os.lstat(path).st_mode
-    except os.error:
+    except FileNotFoundError:
```
ericzolf, line 248: Generic remark: have you tested your code under Windows? I ask because, under Windows, I noticed recently, as I worked on the Windows specific issues raised on the alpha (e.g. long path one) that the exceptions thrown might be "surprising"... And `os.chmod` doesn't really work as expected under Windows...

testing/rdiffbackupdeletetest.py
```diff
@@ -67,7 +66,13 @@ def assertNotFound(self, search):
...
     def tearDown(self):
         if os.path.exists(self.repo):
-            subprocess.check_call(['rm', '-Rf', self.repo])
+            for root, dirs, files in os.walk(self.repo, followlinks=False):
+                for d in dirs:
+                    os.chmod(os.path.join(root, d), 0o777)
+                for f in files:
+                    if not os.path.islink(os.path.join(root, f)):
+                        os.chmod(os.path.join(root, f), 0o777)
+            subprocess.check_call([b'rm', b'-Rf', self.repo])
```
ericzolf, line 75: I know that we do it in the current (outdated) code, but we should try to avoid calling (OS specific) commands instead of using pure python. https://docs.python.org/3.8/library/os.html#os.walk contains an example on how to delete a complete tree, you would only need to adapt it slightly:

```python
for root, dirs, files in os.walk(top, topdown=False):
    for name in files:
        os.remove(os.path.join(root, name))
    for name in dirs:
        os.rmdir(os.path.join(root, name))
```

Extra points if you make it a generic function in commontest to reuse elsewhere...

testing/rdiffbackupdeletetest.py
```diff
@@ -103,6 +108,13 @@ def test_delete_with_directory(self):
         rdiff_backup(1, 1, self.repo, None, extra_options=b"--verify")
         self.assertNotFound(b'tmp')
 
+    def test_delete_with_directory_with_perm(self):
```
ericzolf, line 111: short description of what is being tested doesn't hurt...

ericzolf (changes requested):
Looks overall good to me but needs a few changes.

src/rdiff-backup-delete
ikus060: Not sure it's required to be done for the file. It's gonna get deleted.

src/rdiff-backup-delete
ikus060: Not sure it's required to be done for the file. It's gonna get deleted.

src/rdiff-backup-delete
ikus060: Will make the adjustment

src/rdiff-backup-delete
ikus060: Was previously handling a different kind or error in an intermediate version of the code. It become unnecessary. I agree.

src/rdiff-backup-delete
ikus060: I have no way of testing this on Windows and I'm not sure the current implementation is compatible with Windows either. As part of this task, I just wanted to fix the permissions error on Linux. Making it work on Windows seams out of scope for this commit.

testing/rdiffbackupdeletetest.py
ikus060: done

src/rdiff-backup-delete
ericzolf: Fair enough.

ericzolf (approved):

PR #571, opened 2021-04-24
improve and simplify security handling

ericzolf:
CHG: the restrict mode and path options are specific to the server action, and not generic options
* Simplify code in Security.py and improve message
* Fixing tests replacing Globals with Security
* Get rid of global security variables

### DIFF
--- a/src/rdiff_backup/Globals.py
+++ b/src/rdiff_backup/Globals.py
@@ -243,18 +243,6 @@ select_mirror = None
 # object.  Access is provided to increment error counts.
 ITRB = None
 
-# security_level has 4 values and controls which requests from remote
-# systems will be honored.  "all" means anything goes. "read-only"
-# means that the requests must not write to disk.  "update-only" means
-# that requests shouldn't destructively update the disk (but normal
-# incremental updates are OK).  "minimal" means only listen to a few
-# basic requests.
-security_level = "all"
-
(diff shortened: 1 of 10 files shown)


ericzolf:
Any review?

ikus060:
I tried to review last night. But could make any sense of it. Will try to review asap

ericzolf:
Asking questions is also a form of review. Thanks.

ikus060:
I would love to review it, but I'm not enough into this peace of code to make sens of it. I'm trusting your judgment and understand it's only refactoring. So it should be good.

I don't want it to block you, and it seams nobody is reviewing it either...

aka, I feel this security context should be move else where. Most likely within the `Location` class you just created to very each call make or something similar. But it's probably too early to happen.

ericzolf:
Let me try to explain even if appreciate your trust :-) but would indeed also appreciate to have more reviewers :-(

Anyway, the security module is currently a singleton, and it might stay like this, but currently it does its work a bit too implicitly for my taste: `Server.vet_request` is being called by the connection each time a "message" goes between client and server. So it's _not_ related to a location but to the connection between client and server. And it's not easy to make the link between connection and security more explicit because the server action initiates the setup in the connection.Server() function.

I might change (and document) this in the future, but the current state is good enough (in the sense that I understand it, and can follow the flow). Other aspects are more convoluted.

PR #572, opened 2021-04-25
Use set instead of dictionary for connections

ericzolf:
Makes the code slightly simpler, and is a better fit for the logic


ikus060:
+1

PR #576, opened 2021-05-09
add tox based pipeline for Windows

ericzolf:
* the tox_win.ini wasn't working any more with Python 3.9 and recent changes
DEV: Windows is now part of the test pipeline even if tests are limited.

### DIFF
--- a/tools/win_test_rdiffbackup.sh
+++ b/tools/win_test_rdiffbackup.sh
@@ -30,4 +30,11 @@ export LIBRSYNC_DIR
 ver_name=rdiff-backup-$(${PYEXE} setup.py --version)
 py_ver_brief=${PYTHON_VERSION%.[0-9]}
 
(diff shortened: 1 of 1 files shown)


ericzolf:
There aren't a lot of tests yet, but that's a start, hence ready for review.

ericzolf:
I added two tests that were missing but had to comment them out because they don't work under Windows (for the same reason, probably because Windows doesn't recognize a python script as executable if it doesn't end in `.py`). Anyway, let's merge it like it is, I'll look into adding more tests later on.

ericzolf:
After all, I couldn't resist to fix the issue, and it was as easy as creating a `rdiff-backup.bat` file which is recognized by Windows as "rdiff-backup" executable and which just calls the windows equivalent of `python $0 "$@"` (roughly).

ikus060 (approved):
Really nice !

PR #586, opened 2021-05-16
Fix handling of backslashes in locations

### LINKED ISSUE #585: [BUG] rdiff-backup is too eager to unescape backslashes in locations making it fail for valid Windows paths
## Bug summary

While trying to make the tox-tests work under Windows, I noticed that the following command fails:

```
$ rdiff-backup --remote-schema '%s' test 'ssh -C vagrant@192.168.121.166 C:\\temp\\rdiff-backup\\rdiff-backup.exe --server'::C:/temp
'C:temprdiff-backuprdiff-backup.exe' is not recognized as an internal or external command,
operable program or batch file.
Fatal Error: Truncated header string (problem probably originated remotely)

Couldn't start up the remote connection by executing

    ssh -C vagrant@192.168.121.166 C:\temp\rdiff-backup\rdiff-backup.exe --server
```

whereas `rdiff-backup --remote-schema '%s' test 'ssh -C vagrant@192.168.121.166 C:/temp/rdiff-backup/rdiff-backup.exe --server'::C:/temp` succeeds.

Doesn't really matter: 2.0 and 2.1 show the same behavior.

## More information

The number of backslashes doesn't really matter, one, two or three, the result is more or less the same.
The issue is surely in the `SetConnections.parse_location(file_desc)` function, which does too eagerly unescape backslashes... In my opinion, this isn't necessary because the only aspect we really care is the "double-colon" (`::`), which could need escaping (in very rare cases).

ericzolf:
FIX: backslashes were removed too eagerly in locations, making the use of Windows paths impossible, closes #585

The fix doesn't yet allow to run more tests under Windows but it's an important step towards this possibility, as it allows now to give Windows commands (with backslashes in the path). And, if you wonder, as the OS' own shell is used, it can't be slashes, they must be backslashes.

### DIFF
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -240,13 +240,13 @@ def parse_location(file_desc):
     # before colons.
     # Hence we split along double ones, remove single ones in each element,
     # and join back with a single backslash.
-    # And then we make sure that paths under Windows use / instead of \
     if file_host:
         file_host = b'\\'.join(
             [x.replace(b'\\:', b':') for x in file_host.split(b'\\\\')])
-        file_host = file_host.replace(b"\\", b"/")
     file_path = b'\\'.join(
         [x.replace(b'\\:', b':') for x in file_path.split(b'\\\\')])
+    # And then we make sure that paths under Windows use / instead of \
+    # (we don't do it for the host part because it could be a shell command)
     file_path = file_path.replace(b"\\", b"/")
 
     return (file_host, file_path, None)
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -194,31 +194,60 @@ def parse_location(file_desc):
     # paths and similar objects must always be bytes
     file_desc = os.fsencode(file_desc)
     # match double colon not preceded by an odd number of backslashes
-    file_match = re.fullmatch(rb"^(?P<host>.*[^\\](?:\\\\)*)::(?P<path>.*)$",
-                              file_desc)
-    if file_match:
-        file_host = file_match.group("host")
-        # According to description, the backslashes must be unquoted, i.e.
-        # double backslashes replaced by single ones, and single ones removed.
-        # Hence we split along double ones, remove single ones in each element,
-        # and join back with a single backslash.
-        file_host = b'\\'.join(
-            [x.replace(b'\\', b'') for x in re.split(rb'\\\\', file_host) if x])
-        file_path = file_match.group("path")
-    else:
-        if re.match(rb"^::", file_desc):
-            return (None, None,
-                    "No file host in {desc} starting with '::'.".format(
-                        desc=file_desc))
+    file_parts = [x for x in re.split(rb"(?<!\\)(\\{2})*::", file_desc)
+                  if x is not None]
+    # because even numbers of backslashes are grouped as part of the split,
+    # we need to stitch them back together,e.g.
+    # "host\\\\::path" becomes ["host","\\\\","path"]
+    # which then becomes ["host\\\\", "path"]
+    concat_parts = []
+    keep = None
+    for part in reversed(file_parts):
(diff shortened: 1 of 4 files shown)


testing/commontest.py
```diff
@@ -140,7 +140,10 @@ def _get_locations(src_local, dest_local, src_dir, dest_dir):
     """
     Return a tuple of remote or local source and destination locations
     """
-    remote_location = "cd {rdir}; {tdir}/server.py::{dir}"
+    if os.name == "nt":
+        remote_location = "cd {rdir} & {tdir}\\server.py::{dir}"
```
ikus060, line 144: @ericzolf I don't understand why we need double backslash `\\`

I would expect a single backslash.

testing/setconnectionstest.py
```diff
@@ -17,16 +16,18 @@ def testParsing(self):
...
+        self.assertEqual(pl(rb"test\\ing\::more::and more\\.."),
+                         (b"test\\ing::more", b"and more/..", None))
+        self.assertEqual(pl(rb"strangely named\::file"),
+                         (None, b"strangely named::file", None))
+        self.assertEqual(pl(rb"foobar\\"), (None, b"foobar/", None))
+        self.assertEqual(pl(rb"not\::too::many\\\::paths"),
+                         (b"not::too", b"many/::paths", None))
 
         # test missing path and missing host
-        self.assertIsNotNone(pl(rb"hello\:there::")[2])
```
ikus060, line 28: Why is this test not relevant anymore ? 
`hello\:there::`

tox_win.ini
```diff
@@ -54,8 +54,8 @@ commands =
     python testing/rdifftest.py
 #    python testing/securitytest.py  # no os.getuid + backslash in path handlingon --server
 #    python testing/killtest.py  # cannot create symbolic link
-#    python testing/backuptest.py  # backslash in path handling on --server
-#    python testing/comparetest.py  # backslash in path handling on --server
+#    python testing/backuptest.py  # No module named rdiff_backup in server.py
```
ikus060, line 57: What is the meaning of this comment ?

src/rdiff_backup/SetConnections.py
```diff
@@ -194,31 +194,60 @@ def parse_location(file_desc):
...
+    # before colons.
+    # Hence we split along double ones, remove single ones in each element,
+    # and join back with a single backslash.
+    if file_host:
+        file_host = b'\\'.join(
+            [x.replace(b'\\:', b':') for x in file_host.split(b'\\\\')])
+    file_path = b'\\'.join(
+        [x.replace(b'\\:', b':') for x in file_path.split(b'\\\\')])
+    # And then we make sure that paths under Windows use / instead of \
+    # (we don't do it for the host part because it could be a shell command)
```
ikus060, line 249: The host could be a shell command ?

Do you have an example for me ?

ikus060 (commented):
I have more question then comments ! :/

testing/commontest.py
ericzolf: Backslash is strings or bytes is an escape character,  so you need to escape it with itself. Unless you use raw, the r-prefix, _but_ you still might need to escape it if you use the string as regex.

tox_win.ini
ericzolf: I think, Windows doesn't inherit the environment variables, so PYTHONPATH isn't properly set to point to the tox environment, so it doesn't find the rdiff-backup module.

src/rdiff_backup/SetConnections.py
ericzolf: Yes, this is how all the client/ server tests are faked,  see commontest.py, the function `rdiff_backup`. The remote schema is set to `%s`.

testing/setconnectionstest.py
ericzolf: I just renamed the string to make the test more explicit, but it's basically the same test.

PR #588, opened 2021-05-26
rdiff-backup-delete: Use buffered subprocess gzip or buffered io

ikus060:
Using buffered io improve the performance when reading and writing gzip
files.

CHG: rdiff-backup-delete: improve performance on gzip files

### DIFF
--- a/src/rdiff-backup-delete
+++ b/src/rdiff-backup-delete
@@ -24,14 +24,17 @@
 # metadata.
 #
 
-import os
-import sys
-
 import getopt
 import gzip
+import io
+import os
 import re
 import stat
 import struct
+import subprocess
+import sys
+from distutils import spawn
+
 
 PY2 = sys.version_info < (3,)
 PY26 = sys.version_info > (2, 6) and sys.version_info < (2, 7)
@@ -58,6 +61,67 @@ def _str(value):
         return value.decode('utf-8')
     else:
         return value.decode('utf-8', errors='replace')
+
+
+# Check if gzip is available.
+_GZIP = spawn.find_executable('gzip')
+
+
+class WrapClose:
+    """
+    Helper for _open() -- a proxy for a file whose close waits for the process.
+    """
+
+    def __init__(self, stream, proc):
+        self._stream = stream
+        self._proc = proc
+
+    def close(self):
+        if self._proc.stdin:
+            self._proc.stdin.close()
+        returncode = self._proc.wait()
+        if self._proc.stdout:
+            self._proc.stdout.close()
+        return returncode
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
+    def __getattr__(self, name):
+        return getattr(self._stream, name)
+
+    def __iter__(self):
+        return iter(self._stream)
+
+
+def _open(fn, mode):
+    """
+    Wrapper to open a file with or without compression using gzip executable or
+    pure-python implementation.
+    """
+    compress = fn.endswith(b'.gz')
+    buffered = io.BufferedReader if 'r' in mode else io.BufferedWriter
+    # Open file directly if compression is not required
+    if not compress:
+        return buffered(open(fn, mode))
+
+    # Open file using python gzip if zcat and gzip are not available.
+    if not _GZIP:
+        return buffered(gzip.open(fn, mode))
+
+    # When available, open file using subprocess gzip for better performance
+    if 'r' in mode:
+        proc = subprocess.Popen([_bytes(_GZIP), b'-cd', fn],
+                                stdout=subprocess.PIPE)
+        return WrapClose(proc.stdout, proc)
+    else:  # wb
+        proc = subprocess.Popen([_GZIP],
+                                stdin=subprocess.PIPE,
+                                stdout=open(fn, mode))
+        return WrapClose(proc.stdin, proc)
 
 
 def _print_usage(error_message=None):
@@ -175,7 +239,6 @@ def _remove_from_metadata(repopath, file, dry_run):
         return
 
     print('removing entries `%s` from %s' % (_str(repopath.relpath), _str(file)))
-    _open = gzip.open if file.endswith(b'.gz') else open
     input = _open(file, 'rb')
     tmp_file = os.path.join(os.path.dirname(file), b'.tmp.' + os.path.basename(file))
     output = _open(tmp_file, 'wb')


ikus060:
After more testing, zcat/gzip are still faster then python implementation. So I'm making required changes to use zcat and gzip when available and fallback to gzip with buffered io.

```
rdiff-backup original

  real	11m43.436s
  user	11m37.380s
  sys	0m4.915s
  
rdiff-backup io.BufferedReader()

  real	8m51.554s
  user	8m46.373s
  sys	0m4.263s
  
rdiff-backup subprocess zcat/gzip 

  real	3m49.299s
  user	7m42.138s
  sys	0m18.507s
```

src/rdiff-backup-delete
```diff
@@ -25,13 +25,17 @@
 #
 
 import os
```
ericzolf, line 27: Overall, the imports should be in alphabetical order.

src/rdiff-backup-delete
```diff
@@ -25,13 +25,17 @@
...
+import re
 import sys
 
+from distutils import spawn
```
ericzolf, line 31: See https://github.com/rdiff-backup/rdiff-backup/blob/master/docs/CODING.md#import

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
         return value.decode('utf-8', errors='replace')
 
 
+# Check if zcat is available.
```
ericzolf, line 67: Comment says zcat but executable is gzip

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
...
+GZIP = spawn.find_executable('gzip')
+
+
+class _wrap_close:
```
ericzolf, line 71: Class names are Camel case.

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
...
+    def __getattr__(self, name):
+        return getattr(self._stream, name)
+
+    def __iter__(self):
+        return iter(self._stream)
+
+
+def _open(fn, mode):
+    """
+    Wrapp to open a file with or without compression using zcat/gzip or
```
ericzolf, line 105: Wrap (one p)

ericzolf: And why do you talk again about zcat?

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
...
+        return buffered(open(fn, mode))
+
+    # Open file using python gzip if zcat and gzip are not available.
+    if not GZIP:
+        return buffered(gzip.open(fn, mode))
+
+    # When available, open file using subprocess gzip for better performance
+    if 'r' in mode:
+        proc = subprocess.Popen([_bytes(GZIP), b'-cd', fn],
+                                shell=False,
```
ericzolf, line 121: shell=False is the default and isn't required

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
...
+                                stdout=subprocess.PIPE)
+        return _wrap_close(proc.stdout, proc)
+    else:  # wb
+        proc = subprocess.Popen([GZIP],
+                                shell=False,
```
ericzolf, line 126: shell=False is the default and isn't required

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
         return value.decode('utf-8', errors='replace')
 
 
+# Check if zcat is available.
+GZIP = spawn.find_executable('gzip')
```
ericzolf, line 68: It's an internal "constant", its name should start with underscore.

src/rdiff-backup-delete
```diff
@@ -60,6 +64,71 @@ def _str(value):
...
+        self._proc = proc
+
+    def close(self):
+        if self._proc.stdin:
+            self._proc.stdin.close()
+        returncode = self._proc.wait()
+        if self._proc.stdout:
+            self._proc.stdout.close()
+        if returncode == 0:
+            return None
```
ericzolf, line 87: Why doing this? Why not returning directly the returncode? It's mostly a moot point as I didn't see that you do anything with it, but IMHO having consistently an integer being returned is easier to handle.

ericzolf (changes requested):
Only small things, looks good to me even if it hurts to need such workarounds to get the necessary performance.

ikus060:
> Only small things, looks good to me even if it hurts to need such workarounds to get the necessary performance.

I completely agree. gzip performance in python is a shame. It one of the reason why I recommend disabling it by default and let the compression of FS do it's thing. With zstd-4 in ZFS it even boost the performance !

src/rdiff-backup-delete
ikus060: Be more specific here. I have no clue what is the problem. `spawn` is a module.

src/rdiff-backup-delete
ikus060: I first started by using zcat ang gzip, but finally figure out how to use gzip for both operation. And I didn't review my comments.

src/rdiff-backup-delete
ikus060: fixed

src/rdiff-backup-delete
ikus060: fixed

src/rdiff-backup-delete
ikus060: fixed

src/rdiff-backup-delete
ericzolf: Ups. Sorry, my bad. You're right.

src/rdiff-backup-delete
ericzolf: https://www.python.org/dev/peps/pep-0008/#imports doesn't say anything about it, I have a slight preference for the 2nd approach, which is also what you chose, so fine with me.

src/rdiff-backup-delete
```diff
@@ -60,6 +63,67 @@ def _str(value):
...
+    pure-python implementation.
+    """
+    compress = fn.endswith(b'.gz')
+    buffered = io.BufferedReader if 'r' in mode else io.BufferedWriter
+    # Open file directly if compression is not required
+    if not compress:
+        return buffered(open(fn, mode))
+
+    # Open file using python gzip if zcat and gzip are not available.
+    if not GZIP:
```
ericzolf, line 112: It should be _GZIP. That's apparently why the flake8 test fails.

ericzolf (approved):
Just fix the tests (flake8/_GZIP) and then you can merge.

PR #590, opened 2021-05-30
Convert from Markdown to Asciidoc documentation

frankcrawford:
Automated and cleaned up most markdown documents to asciidoc, but have not converted documents in .github/ISSUE_TEMPLATE, as they seem to be required in markdown format, or tools/windows/roles/* as they appear to be included from another distribution.

Also, have not yet converted docs/rdiff-backup.1.md yet, as it requires special treatment.

Finally, will also remove all the related *.md files once happy with this PR.


ericzolf:
I browsed quickly through it and it looks good to me. Just one thing: credits.md and example.md haven't been converted. The rest indeed shouldn't be converted. 

Do you want to remove the markdown files in the same PR or in a new one?

frankcrawford:
@ericzolf Thanks.  I'll fix up those 2 files, don't know how I missed them, as well as fixing the man page.

I'll also do the removal of the markdown files in this PR, but was just holding off the actual delete until we are sure that it all looks good.

ericzolf:
IMHO you can remove the markdown files and call it a day for now. As I went through the documentation, I noticed how much old stuff is in there, so we'll need after to do some more clean-up but this has nothing to do with the transcription to asciidoc.

frankcrawford:
I suspect the failures are related to the conversion of rdiff-backup.1 from md to adoc, and the automated testing doesn't know how to generate the man file from an AsciiDoc file.

frankcrawford:
@ericzolf it looks like I need to update setup.py, Dockerfile and tools/build_wheels.sh if this is to work.
Do you want me to follow this up?

ericzolf:
You should only need to change `setup.py` lines 288 and 289, replacing the pandoc call through an asciidoctor one. If I'm not mistaking:
```
"pandoc --standalone --to man --variable date=\"{date}\" "
 "--variable footer=\"Version {ver}\" {infile} -o {outfile}"
```
should become something like:
```
"asciidoctor -b manpage -a revnumber=\"{ver}\" -a revdate=\"{date}\'
"-o {outfile} {infile}"
```
You could then add the `{revdate}` at the end of the `:manmanual:` variable (according to https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/ there should be no blank in-between).

frankcrawford:
@ericzolf except we probably also need to make sure that asciidoctor or asciidoc or whatever is in the Docker images for testing as well, but I don't think I have rights to update that.

BTW, what do you want to make the dependency asciidoctor or asciidoc?

ericzolf:
Ah, and you surely need to replace pandoc with asciidoctor as dependency in `.github/workflows/*.yml` (under Windows, there is no choco package for asciidoctor, so you'll need to install asciidoctor via gems, see https://docs.asciidoctor.org/asciidoctor/latest/install/windows/).

ericzolf:
Ok. Let's rollback the man page change and I'll take it from there after we've merged.

frankcrawford:
Yeah, I was just thinking that, or rather just roll back the delete of rdiff-backup.1.md for now.

ericzolf:
Good idea, even better.

frankcrawford:
And now to update setup.py for all the doc files to be copied!

frankcrawford:
> You could then add the `{revdate}` at the end of the `:manmanual:` variable (according to https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/ there should be no blank in-between).

BTW, the document shows no space, but their example file has spaces in there.  I suspect the space is optional.

ericzolf:
> BTW, the document shows no space, but their example file has spaces in there. I suspect the space is optional.

OK, thanks, I'll keep an eye on it.

frankcrawford:
@ericzolf I'll do another pass of rdiff-backup.1 and also do one for rdiff-backup-statistics.1, but I'll submit them as a new PR.

PR #591, opened 2021-06-07
Additional manpage conversions

frankcrawford:
Updated rdiff-backup.1.adoc
Added rdiff-backup-statistics.1.adoc

### DIFF
--- a/docs/rdiff-backup-statistics.1.adoc
+++ b/docs/rdiff-backup-statistics.1.adoc
@@ -1,0 +1,86 @@
+= RDIFF-BACKUP-STATISTICS(1)
+:doctype: manpage
+:docdate: {revdate}
+:man source: rdiff-backup-statistics
+:man version: {revnumber}
+:man manual: Rdiff-Backup-Statistics Manual {revnumber}
+
+== NAME
+
+rdiff-backup-statistics - summarize rdiff-backup statistics files
+
+== SYNOPSIS
+
+*rdiff-backup-statistics* [*--begin-time* _time_] [*--end-time* _time_] [*--minimum-ratio* _ratio_] [*--null-separator*] [*--quiet*] [*-h*|*--help*] [*-V*|*--version*] _repository_
+
+== DESCRIPTION
+*rdiff-backup-statistics*
+reads the matching statistics files in a backup repository made by
+*rdiff-backup*
+and prints some summary statistics to the screen.  It does not alter
+the repository in any way.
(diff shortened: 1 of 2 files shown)


docs/rdiff-backup-statistics.1.adoc
```diff
@@ -0,0 +1,86 @@
...
+total.  The default value is .05, or 5 percent.
+
+--null-separator::
+Specify that the lines of the file_statistics file are separated by
+nulls ('\0').  The default is to assume that newlines separate.  Use
+this switch if rdiff-backup was run with the *--null-separator* when
+making the given repository.
+
+--quiet::
+Suppress printing of the '[.code]``Processing statistics from session...``"
```
ericzolf, line 69: the code section starts with a single quote and ends with a double one.

And I think, it would be nicer to have three backquotes instead of the `[.code]` thingy, i.e.

```asciidoc
'```bla bla```'
```

It would be more readable as raw code, and it would even give us curved quotes, which look slightly nicer, see https://docs.asciidoctor.org/asciidoc/latest/text/quotation-marks-and-apostrophes/

This last comment is of course valid for the whole document. Else everything looks good to me.

frankcrawford:
Updated.

The reason why it is use [.code] all through the file is that it had the most consistent result in the rdiff-backup.1.adoc file.  During the discussion of globbing and pattern matching, there were lots of issue, that required random backslashes, etc, to resolve.  I converted to using them almost exclusively, and just did it through out the files.

docs/rdiff-backup.1.adoc
```diff
@@ -333,11 +331,11 @@ The default is to include other filesystems.
 
 --min-file-size _sizeinbytes_:: Exclude files that are smaller than the given size in bytes.
 
-[[statistics-options]]
+[[satistics-options]]
```
ericzolf, line 334: Sorry,  I see a missing t here.

frankcrawford: Done

frankcrawford: Something broke in the testing, it can't find setuptools-scm.

ericzolf: There were CDN issues during the day,  just amend the last commit and force push it.

ericzolf: Sorry, there is a re run jobs button in the checks tab, no need to force push!

PR #598, opened 2021-06-12
Use adoc for manpages 

ericzolf:
It wasn't as easy as I thought it would be, mainly due to the old manylinux versions, but now it works! No pandoc left, only asciidoc. Thanks again to @frankcrawford for doing most of the work.


frankcrawford:
@ericzolf where are you trying to find asciidoctor in build_wheel.sh?  rubygem-asciidoctor is in EPEL, although it really would end up similar to your current process.

ericzolf:
build_wheel.sh is being run within the manylinux images, and I just checked, EPEL is even activated. OK, let me fix this.

ericzolf:
Yep, but asciidoctor isn't in the EPEL of manylinux_2010 (CentOS 6), so let's keep like this, at least it works!

PR #599, opened 2021-06-12
Improve quoting testing and quoting itself

ericzolf:
Initially, I only wanted to improve the testing coverage for
FilenameMapping (which I did from ~60 to 77%), then I noticed that
the requoting algorithm wasn't quite correct.

CHG: requoting a repository can be enforced using --chars-to-quote, use with care


ikus060 (approved):
I didn't see anything special here.

Issue #122, opened 2019-09-11
Insufficient defense against concurrent operations on a backup repository


zjw:
rdiff-backup blocks certain types of concurrent operations that could cause problems, such as the initiation of either a backup or a restore while a backup is already ongoing.  But there are a couple cases of concurrency that are not blocked, but should be.

It is currently possible (in v. 1.2.8) for multiple instances of `rdiff-backup --check-destination-dir` to be running simultaneously on a single backup repository.  I am not sure if this could lead to corruption of the repository, but at a minimum, it is a waste of processing time & resources.

A real-life scenario of this occurring was mentioned [here](https://lists.nongnu.org/archive/html/rdiff-backup-users/2019-09/msg00007.html).  In that particular instance, the user was performing a regression after a failed backup.  The regression was taking many hours (as is common for large repositories).  Meanwhile, the user's regularly scheduled daily backup automatically started, which initiated another regression of the repository, even while the earlier regression was still ongoing.

Proper behavior would be for a regression to terminate immediately upon detecting that another regression was already in progress.

Another case of concurrency that is not blocked is the initiation of a backup while a restore operation is already ongoing.  This type of concurrency should also be blocked.

zjw:
Expanding on my original comments -- The first concurrency problem I mentioned can lead to a situation with a backup running concurrently with a regression:

1. Start a backup after a previously failed backup.  The failed backup is detected, and a regression is initiated.
2. Before the above regression is complete, start another backup, which initiates another regression.
3. Once the first regression is complete, the first backup begins, even while the second regression is ongoing.

Again, I don't know if this could lead to corruption.  If nothing else, the second regression will finish and will give up when attempting to remove a `current_mirror` file that was already removed when the first regression finished.

ikus060:
In additional, the `current_mirror` file with a pid is not reliable. multiple rdiff-backup process might be running on different server with the same NFS mount point. Or running inside a docker container... PID is not the way to go.

Locking mechanism that is reliable is not simple if we want to make it work on all the file-system (local or network). File Lock is simple not supported properly on network file system.

The only real implementation I have experience with and that is working properly is H2DB FileLock:
http://www.h2database.com/html/advanced.html#file_locking_protocols

ericzolf:
I've split out the more reliable approach over network file system topic into a separate issue (as enhancement) because it sounds much more difficult to realise than just making sure that check-destination-dir follows the current locking mechanism.

[...]

PR #603, opened 2021-06-17
Add locking mechanism to repositories

### LINKED ISSUE #415: [ENH] make locking of backup repo reliable on network file system
In additional, the `current_mirror` file with a pid is not reliable. multiple rdiff-backup process might be running on different server with the same NFS mount point. Or running inside a docker container... PID is not the way to go.

Locking mechanism that is reliable is not simple if we want to make it work on all the file-system (local or network). File Lock is simple not supported properly on network file system.

The only real implementation I have experience with and that is working properly is H2DB FileLock:
http://www.h2database.com/html/advanced.html#file_locking_protocols

_Originally posted by @ikus060 in https://github.com/rdiff-backup/rdiff-backup/issues/122#issuecomment-566047927_

ericzolf:
NEW: repositories are now locked to avoid concurrent write actions, i.e. backup, regress and remove, closes #415


ikus060:
OMG, I'm waiting for that feature for a while now.

src/rdiffbackup/locations/repository.py
```diff
@@ -86,6 +92,9 @@ def check(self):
         if self.must_be_writable and not self._is_writable():
             ret_code |= 1
 
+        if self._is_locked():
```
ikus060, line 95: If I'm reading this the right way, this might return True in case a previous backup failed and leave the ".lock" file.

src/rdiffbackup/locations/repository.py
```diff
@@ -267,6 +286,79 @@ def _is_writable(self):
...
+        and verify that no other process took the lock by comparing its
+        content.
+
+        Return True if the lock could be taken, False else.
+        """
+        fqdn = socket.getfqdn()
+        currtime = Time.curtimestr
+        pid = os.getpid()
+        cmd = sys.argv[0]
+        identifier = "{} {} {} {}".format(fqdn, currtime, pid, cmd)
```
ikus060, line 324: If you are doing this.

Somewhere in the code we write the pid to current_mirror. That become useless don't you think ?

ikus060 (changes requested):
I'm not sure it's an improvement. Running rdiff-backup, alot of automated backup are running and doesn'T complete because of flaky Internet connection as an example. In this scenario, the ".lock" file will stay on the remote server and will effectively "lock" the repository until something or someone delete the file.

I don't have yet a solution for this, but the direction is good. We simply need to define an heuristic to get the file deleted automatically when nothing is really happening.

We could try to use FileSystem Lock for this ".lock" file. But it's not working on all filesystem. POSIX and NTFS are supporting this.

I'm sure python have some nice library for this too. Worth having a look.

src/rdiffbackup/locations/repository.py
ericzolf: Correct. Unless you have the force option.

src/rdiffbackup/locations/repository.py
ericzolf: Yeah. I didn't change it for backward compatibility reasons.

ericzolf:
Overall. I need to rethink it a bit, so draft mode for now. Sorry for the noise.

Regarding your question about existing library, I couldn't find one which is platform independent and convinced me, and I didn't want a solution too dependent on the file system type. That restricts a lot my options...

But, as said, I need to rethink it so throw your ideas at me.

ikus060:
The only implementation that I know that work cross-platform and on every file system, network or anything, is H2DB file lock.

More or less:
A demon thread write  a signature into the lock file and continually, check if the file changed. When the file changes, it replace the content by it's signature.

A process that want to aquire lock, need to write it's signature to the lock file. Then when to check if it get replaced within x amount of time. If not, we consider the lock to be acquired.

Read more here: http://www.h2database.com/html/advanced.html#file_locking_protocols

And here:
https://github.com/h2database/h2database/blob/3653ab38381c47cc37a57ef47bd133af15d496c4/h2/src/main/org/h2/store/FileLock.java#L311

It's relatively simple to implement. Only disadvantage is we need to start a daemon thread. I know for a fact that this mechanism is working over sshfs, smbfs, windows mount point, nfs, on linux, windows, etc.

ericzolf:
I've looked at the protocol and it's very similar to what I implemented with one complication due to the fact that the same process might want to write multiple times to the database, and you need to have some interleaving of write requests. It isn't our case here as one process takes the lock, does the backup/regress/remove/whatever, then releases the lock and leaves.

You also need to consider that a lock can be overwritten/ignored using the `--force` option, so that you're not blocked if the lock has remained. This said, I should output the content of the file if there is a lock so that the user knows better what they're facing (and add the action name to the content).

ericzolf:
Just for the record, the issue with the pipeline is that the lock mechanism isn't allowed by the current security rules for remote access, so I need to add those rules somehow, but I realized also that it would be better, for security and performance reason, to execute the whole function remotely, and not each individual step.

src/rdiffbackup/locations/repository.py
ericzolf: New version does check if the process which created the lockfile is still running with the same PID _and_ name.

ericzolf:
PR is ready for review, it's almost a complete rewrite, limited to the usage of API 201 (because it requires new functions is_locked/lock/unlock to work).

ericzolf:
I realize that this PR closes as well #122

ikus060:
Hello @ericzolf 

The locking mechanism is better then previous version,. but is not robust enough and might be improved a bit. e.g.:

1. If rdiff-backup crash, the lock file will remain on the filesystem and someone need to manually remove the lock file or run rdiff-backup with `--force`. Which is not a good solution for all the automated backup running in cron job.
2. Verification of the PID and CMD is a good protection, but is not enough nowaday. If the process is runing in a container like LXC with shared filesystem or Docker with a mount point or simply trough a shared NFS the PID might not match because it's running on a different system.

The locking mechanism could be more robust if we use filesystem locking like `fcntl` and `msvcrt.locking`.

1. Using `fcntl` or `msvcrt.locking` would allow rdiff-backup to fail, but the next process will be able to lock file file again.
2. On any shared mount point like NFS or event Docker with mount point, the `fcntl` would prevent multiple process to lock file file.

ericzolf:
Point 1 isn't true for two main reasons: the code in `__exit__` should be executed in almost all cases, _and_ a stale lockfile is recognized by its PID/process name. Now point 2 is indeed relevant. OK, let me think about it.

ikus060:
Regarding point #2, it might be interesting to add the local "hostname" to the lock file data that way it's possible to identify the lxc container, physical host or docker container... Mostly for debugging.

ericzolf:
3rd tentative: Linux locking using fcntl.flock is now used + hostname is part of the information stored. I still need to implement the Windows side of the house, hence back to draft mode. @ikus060 if you could test on NFS that it works as expected, that would be much appreciated.

ericzolf:
OK, I did some final tests under Windows and it seems to work as well. Thanks for the constructive review, it was a bit painful but the result is probably a better solution. Now a final review and we should be done.

ikus060:
I'm not sure I will have time to compile and test. Have you put in place something to generate the binaries on every build ? That would ease my pain to simply install this branch version instead of git checkout than compile.

ericzolf:
In general you only need to push the branch to a remote branch with underscore at the end and the packages will get built, e.g.

```
git push origin ericzolf-add-locking-repo-415:ericzolf-add-locking-repo-415_
# wait until build is finished
git push origin -d ericzolf-add-locking-repo-415_
```
But I just started it for you...

ericzolf:
You should now see the artifacts under https://github.com/rdiff-backup/rdiff-backup/releases/tag/untagged-b840a1978ab67b8644d0 (only visible as project admin, sorry for others).

ericzolf:
Got any time for a review?

src/rdiffbackup/utils/psutil.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get_pid_name(pid):
+    """
+    Gets a Process ID as integer or string and returns a process/image name
+    Returns None if no process is running with the given PID
+    """
+    if os.name == "nt":
+        cmd = ("tasklist", "/nh", "/fi", "pid eq {pp}".format(pp=pid))
+    else:
+        cmd = ("ps", "q", str(pid), "o", "comm=", "o", "pid=")
```
ikus060, line 38: Using `ps` might not be the most stable approach. To my knowledge, the arguments are not stable between Linux and Unix. 
On Debian:
```
$ ps q 1493225 o comm= o pid=
gnome-screensho 1493225
```

On MacOS:
```
% ps q 37511 o comm= o pid=
ps: illegal option -- q
usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
          [-u]
          [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
       ps [-L]
```

I'm a bit surprise psutil is not covering our need. On Windows I get this:
image

On Linux Debian I get this: 
image

On MacOS
image

src/rdiff_backup/rpath.py
```diff
@@ -1167,15 +1167,17 @@ def write_from_fileobj(self, fp, compress=None):
     def write_string(self, s, compress=None):
         """Write string s into rpath"""
         assert not self.lstat(), "File '{rp!r}' already exists".format(rp=self)
-        with self.open("w", compress=compress) as outfp:
-            outfp.write(s)
+        outfp = self.open("w", compress=compress)
+        outfp.write(s)
+        outfp.close()
```
ikus060, line 1172: How is it related to the locking !?

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -1048,6 +1060,88 @@ def init_owners_mapping(cls, users_map, groups_map, preserve_num_ids):
...
+    # @API(RepoShadow.unlock, 201)
+    @classmethod
+    def unlock(cls, lockfile, exclusive):
+        """
+        Remove any lock existing.
+
+        We don't check for any content because we have the lock and should be
+        the only process running on this repository.
+        """
+        if cls._lockfd:
```
ikus060, line 1138: Shoudn't we call `locking.unlock()` somewhere ?

ikus060 (changes requested):

src/rdiffbackup/utils/locking.py
```diff
@@ -0,0 +1,65 @@
...
+                raise BlockingIOError(-1, exc.strerror,
+                                      file.name, exc.winerror)
+            else:
+                raise
+
+    def unlock(file):
+        hfile = win32file._get_osfhandle(file.fileno())
+        win32file.UnlockFileEx(hfile, 0, 0xffff0000, __overlapped)
+
+elif os.name == 'posix':
```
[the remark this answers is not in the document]
ericzolf: It's neither nt nor java, so it must have been posix: https://docs.python.org/3/library/os.html#os.name

src/rdiff_backup/rpath.py
ericzolf: The tests were failing with those lines. I assume that two `with` one within the other wasn't working so well.

src/rdiffbackup/locations/_repo_shadow.py
ericzolf: Closing the file releases the lock automatically.

src/rdiffbackup/utils/psutil.py
ericzolf: Under WIndows, psutil was still listing processes which were not running anymore, they were still listed but had a state `stopped` (or something like this). I can't reproduce it right now (of course), and we should indeed keep MacOS running. I assumed it would follow the BSD format... OK, let me look into this once again.

ericzolf:
I know what I didn't like:

```
>>> import psutil 
>>> import os 
>>> os.getpid() 
4844 
>>> psutil.Process(4844) 
psutil.Process(pid=4844, name='python.exe', status='running', started='05:46:55'
)
>>> psutil.Process(4845) 
psutil.Process(pid=4845, name='python.exe', status='terminated', started='05:46:
55')
>>> psutil.Process(4845).is_running() 
True 
>>> psutil.Process(4844).is_running() 
True 
>>> psutil.Process(4845).status() 
[...]
psutil.NoSuchProcess: process no longer exists (pid=4845)
```
That doesn't make a lot of sense, but OK, I'll implement both side by side, and we'll see.

ericzolf:
Does `ps -o comm= -p 46470` work under MacOS and return only the command (without parameters)?

ikus060:
> Does `ps -o comm= -p 46470` work under MacOS and return only the command (without parameters)?

Indeed it's working:
```
% ps -o comm= -p 1071 
/Applications/Minarca.app/Contents/MacOS/minarca
```

ikus060 (approved):

ericzolf:
Thanks for the deep review, it costed time but I think, it made the code better.

PR #606, opened 2021-07-09
Move get_inclist from restore to rpath

ericzolf:
- move restore.get_inclist to rpath.get_incfiles_list to clean-up the code
- avoid recursive import between FilenameMapping and rpath by overloading the QuotedRpath.dirsplit method, making also the code easier

### DIFF
--- a/src/rdiff-backup-statistics
+++ b/src/rdiff-backup-statistics
@@ -30,7 +30,6 @@ from rdiff_backup import (
     FilenameMapping,
     robust,
     rpath,
-    restore,
 )
 
 begin_time = None  # Parse statistics at or after this time...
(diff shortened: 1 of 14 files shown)


ikus060:
@ericzolf I would need some context to better understand your change. Why are you moving this function ?

ikus060 (approved):
Technically the changes is good.

ericzolf:
The main reason was that it's a generic file/repo function, so it has nothing to do in the restore module. It was used all over the place and the restore module was often imported only for this one function. And the rpath module already contained the is_incfile function m it fits better there. The code looks also cleaner.

ericzolf:
Thanks for approving!

PR #607, opened 2021-07-15
Deprecate remote functions in Main and backup

ericzolf:
DEV: remote functions in Main and backup are deprecated from the API and replaced by class methods in _repo_shadow and _dir_shadow

Architecture documentation has been extended to explain the shadow classes.

Further PRs will follow to do the same for other actions, and probably more. Check the docs/arch/locations.adoc changes in this PR for the idea behind those changes.

### DIFF
--- a/src/rdiff_backup/Security.py
+++ b/src/rdiff_backup/Security.py
@@ -237,47 +237,63 @@ def _set_allowed_requests(sec_class, sec_level):
             "restore.MirrorStruct.initialize_rf_cache",
             "restore.MirrorStruct.close_rf_cache",
             "restore.MirrorStruct.get_diffs", "restore.ListChangedSince",
-            "restore.ListAtTime", "backup.SourceStruct.get_source_select",
-            "backup.SourceStruct.set_source_select",
-            "backup.SourceStruct.get_diffs",
+            "restore.ListAtTime",
             "compare.RepoSide.init_and_get_iter",
             "compare.RepoSide.close_rf_cache", "compare.RepoSide.attach_files",
             "compare.DataSide.get_source_select",
             "compare.DataSide.compare_fast", "compare.DataSide.compare_hash",
-            "compare.DataSide.compare_full", "compare.Verify"
+            "compare.DataSide.compare_full", "compare.Verify",
+            # API < 201
+            "backup.SourceStruct.get_source_select",
+            "backup.SourceStruct.set_source_select",
+            "backup.SourceStruct.get_diffs",
+            # API >= 201
+            "_dir_shadow.ShadowReadDir.get_select",
+            "_dir_shadow.ShadowReadDir.set_select",
+            "_dir_shadow.ShadowReadDir.get_diffs",
         ])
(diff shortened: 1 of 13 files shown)


ericzolf:
The test fails because the coverage has dropped below 80%. Fair point, I'll add new tests to cover the new code with API 201.

The new code can already be reviewed though.

ericzolf:
Back to 83% coverage, you can review!

ericzolf:
Someone to review please?

src/rdiff_backup/Security.py
```diff
@@ -237,47 +237,63 @@ def _set_allowed_requests(sec_class, sec_level):
...
-            "backup.SourceStruct.set_source_select",
-            "backup.SourceStruct.get_diffs",
+            "restore.ListAtTime",
             "compare.RepoSide.init_and_get_iter",
             "compare.RepoSide.close_rf_cache", "compare.RepoSide.attach_files",
             "compare.DataSide.get_source_select",
             "compare.DataSide.compare_fast", "compare.DataSide.compare_hash",
-            "compare.DataSide.compare_full", "compare.Verify"
+            "compare.DataSide.compare_full", "compare.Verify",
+            # API < 201
```
ikus060, line 246: Are you introducing another version  for the API ?
Do you think it's necessary since 201 was not even releases.

ikus060:
Will take the time to review tomorrow. Sorry, I did not notice that PR.

Another question for you in regards to remote calls. I notice while developing Minarca lately that our call to remote schema calling ssh will use stdin and stdout to communicate. This is fine. But the stderr is untouched and that allow the remote server to print details trough the stderr.

I'm not sure this is intentional and I would believe it's supposed to go trough the Log class instead with a remote cal.

src/rdiff_backup/Security.py
ericzolf: No, it's only 200 and 201. I'm unsure why you think it might be so...

ericzolf:
It goes through the log class but the log class uses stderr to output the message,  without going through the usual way, which is reserved for "real" data. I guess,  it would have been too anyway to keep stderr clean, with exceptions&co.

ikus060:
Ok, I took the time to really review your changes. I understand it goes along the line of refactoring the API to centralize everything under one roof. I remember you wrote about this direction and I think it's the best. I also understand that this change is a first step in the process.

Side note regarding the API changes it make the code overly complex with duplicated code. I understand the reason, but I also understand it's unavoidable,

Good job, as always

ericzolf:
Thanks for the review and the kind words :-)

And, yes, looking forward to version 3.x and be able to remove all the backward compatibility cruft. But we're by far not yet there...

What I was thinking BTW:
* version 2.2 will have API 200 as default, API 201 possible (with --api-version)
* version 2.4 will have API 201 as default, API 200 possible
* version 2.x (x>4) we'll see..., but API 200 and 201 will remain possible
* version 3.0 will have API 201 as default (or perhaps 202 or later, depending how 2.x evolves) and get rid of API 200.

PR #610, opened 2021-07-28
Fix server calls to new schema

ericzolf:
It avoids too many warnings about deprecation, some remain to still test old interface
FIX: handling of RDIFF_BACKUP_VERBOSITY was broken after recent changes


ikus060 (approved):

ericzolf:
Thanks!

PR #634, opened 2021-11-13
Get rid of most manage functions

### LINKED ISSUE #616: [BUG] `remove increments --older-than` returns an error when there is no increment to remove in v2.1.0a1
## Bug summary

Using v2.1.0a1, calling *with the new or old interface*, removing increments with the `--older-than` option returns 1 (error) if there is no increment old enough to be removed.

```yaml
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - c:\Program Files\rdiff-backup\rdiff-backup.exe
  - info
  version: 2.1.0a1
python:
  executable: c:\Program Files\rdiff-backup\rdiff-backup.exe
  name: cpython
  version: 3.9.1
system:
  fs_encoding: utf-8
  platform: Windows-10-10.0.19041-SP0
...
```

## rdiff-backup call

```
"C:\Program Files\rdiff-backup-v210a1\rdiff-backup.exe" remove increments --older-than 3Y D:\  # returns 1
"C:\Program Files\rdiff-backup-v205\rdiff-backup.exe" --remove-older-than 3Y D:\  # returns 0
```

`rdiff-backup` v2.1.0a1 returns `1` while v2.0.5 returns `0`.

As a user, I would expect the command to return `0` (success) if it can access the backup directory and list its increments, meaning it runs without unexpected error, even if there is no matching increment to remove and it terminates without removing any.
This expectation seems consistent with how the `man` entry describes `rdiff-backup` return value:
````
If rdiff-backup finishes successfully, the exit status will be  0.   If
       there  is  an unrecoverable (critical) error, it will be non-zero (usu-
       ally 1, but don't depend on this  specific  value).   When  setting  up
       rdiff-backup  to  run  automatically (as from cron(8) or similar) it is
       probably a good idea to check the exit code.
````

Note that this "no matching increment found" case is different from the "specified increment not found" case of removing a given specific increment, where it does make sense to consider it an error if the increment doesn't exist.

## More information

My use case is running `rdiff-backup` in automated scripts and performing cleanup of old backups automatically.
The current behavior makes it impossible to check the exit code for potential execution error (e.g. "backup repository unaccessible") as suggested by the manual in such a use case.

ericzolf:
- the list increments have been moved to repository and list_
- the delete/remove function has been copied to repository and remove
  action (and the old one deprecated)
- add list and remove test cases for API >= 201

Note that the delete_earlier_than(_local) functions remain for
compatibility reasons.

NEW: parsable list of increments is in YAML format for easier parsing
CHG: human readable list of increments with size has slightly changed and is in the same order as the list _without_ size for consistency
FIX: do not return an error if no increment is old enough to be removed, closes #616

### DIFF
--- a/docs/api/v201.adoc
+++ b/docs/api/v201.adoc
@@ -55,7 +55,7 @@
 * `Main.backup_close_statistics` **deprecated**
 * `Main.backup_remove_curmirror_local` **deprecated**
 * `Main.backup_touch_curmirror_local` **deprecated**
-* `manage.delete_earlier_than_local`
+* `manage.delete_earlier_than_local` **deprecated**
 * `regress.check_pids`
 * `regress.Regress`
 * `restore.ListAtTime` **deprecated**
@@ -102,6 +102,7 @@
 * `locations._repo_shadow.ShadowRepo`  **new**
 ** `.close_rf_cache`
 ** `.close_statistics`
+** `.delete_increments_older_than`
 ** `.get_diffs`
 ** `.get_increment_times`
 ** `.get_mirror_time`
--- a/src/rdiff_backup/Security.py
+++ b/src/rdiff_backup/Security.py
@@ -290,14 +290,15 @@ def _set_allowed_requests(sec_class, sec_level):
             "rpath.delete_dir_no_files",
             "fs_abilities.restore_set_globals",
             "fs_abilities.single_set_globals", "regress.Regress",
-            "manage.delete_earlier_than_local",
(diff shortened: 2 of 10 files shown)


ericzolf:
Ready for review!

ikus060:
My understanding, this changes is mostly refactoring.

Everything seams allright

ericzolf:
Thanks, very much appreciated!

PR #635, opened 2021-11-13
Don't count coverage for server action breakpoint

ericzolf:
It's unnecessary as it's only a debug function

Very small PR...


ikus060 (approved):
+1

PR #639, opened 2021-11-19
Refactor the regress action

ericzolf:
This led to the need to rename and formalize the action.repo and
action.dir, and some re-shuffling of backup and restore (because
they use regress methods).

I also fixed a few errors in the action tests.

### DIFF
--- a/docs/api/v201.adoc
+++ b/docs/api/v201.adoc
@@ -63,8 +63,8 @@
 * `Main.backup_remove_curmirror_local` **deprecated**
 * `Main.backup_touch_curmirror_local` **deprecated**
 * `manage.delete_earlier_than_local` **deprecated**
-* `regress.check_pids`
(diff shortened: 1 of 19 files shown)


ericzolf:
@ikus060 as you've been so fast to answer in the recent past, I assume that it was an oversight.

ikus060:
> @ikus060 as you've been so fast to answer in the recent past, I assume that it was an oversight.

I'm trying to keep up the review. I don't want it to stop your development. 

Since all your latest changes are refactoring the new actions, those are fast to review.

ericzolf:
I meant with my remark that you probably oversaw this PR because you reviewed the other one, created at almost the same time.

ikus060 (approved):

ericzolf:
Thanks!

PR #652, opened 2021-12-31
Use Globals.set_all instead of SetConnections.UpdateGlobal

ericzolf:
It is more logical to have set_all together with set and set_local.

A simple one, but it depends on #651 and will need to be rebased before it can be merged.


ericzolf:
I've deprecated  Globals.postset_regexp_local in the same PR to avoid one review, not very clean but who cares?

ericzolf:
A rather simple PR ready for your review...

dio4ev:
I haven't looked at PR #651 yet, but this one looks good from my side.

ericzolf:
Thanks!

Issue #633, opened 2021-11-10
[BUG] rdiff-backup fails to build with Python 3.11: error: lvalue required as left operand of assignment


frankcrawford:
## Bug summary

Test build of rdiff-backup 2.0.5 fails to build with Python 3.11.0a2.  See below for full details of the issue.

Note, this is more of a pre-warning, as Python 3.11 has not yet been released, and this is preparation for Fedora 37, which is up to 12 months away.

## Version, Python, Operating System

rdiff-backup: 2.0.5
Python: 3.11.0a2
OS: Fedora rawhide

## rdiff-backup call

N/A

## What happened and what did you expect?

Details from Red Hat Bugzilla – Bug 2021946 (https://bugzilla.redhat.com/show_bug.cgi?id=2021946)

```
rdiff-backup fails to build with Python 3.11.0a2.

src/_librsyncmodule.c:543:36: error: lvalue required as left operand of assignment
  543 |   Py_TYPE(&_librsync_SigMakerType) = &PyType_Type;
      |                                    ^
src/_librsyncmodule.c:544:38: error: lvalue required as left operand of assignment
  544 |   Py_TYPE(&_librsync_DeltaMakerType) = &PyType_Type;
      |                                      ^

Since Py_TYPE() is changed to a inline static function, Py_TYPE(obj) = new_type must be replaced with Py_SET_TYPE(obj, new_type): see the Py_SET_TYPE() function (available since Python 3.9). For backward compatibility, this macro can be used:

#if PY_VERSION_HEX < 0x030900A4 && !defined(Py_SET_TYPE)
static inline void _Py_SET_TYPE(PyObject *ob, PyTypeObject *type)
{ ob->ob_type = type; }
#define Py_SET_TYPE(ob, type) _Py_SET_TYPE((PyObject*)(ob), type)
#endif

(Contributed by Victor Stinner in bpo-39573.)
```

## More information

Will continue to track this with the Python test team, and may end up adding a patch as listed to test out if there are any issues with Python 3.11

ericzolf:
As long as the solution is backward compatible, we can integrate it once you're ready.

IIRC it's the same issue you had already plugged for 3.10 but where the change was taken back?

PR #655, opened 2022-01-02
Added patch for Python3.11 as per bpo-39573.

frankcrawford:
FIX: Added patch for Python3.11 as per bpo-39573. as noted in Issue #633.

### DIFF
--- a/src/_librsyncmodule.c
+++ b/src/_librsyncmodule.c
@@ -25,6 +25,16 @@
 #include <librsync.h>
 #define RSM_JOB_BLOCKSIZE 65536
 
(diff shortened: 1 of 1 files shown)


ericzolf (approved):
Guessing that you know what you're doing, and given that the tests passed on older versions of Python, I approve the PR.

frankcrawford:
Just for reference, it builds happily against Python3.10 (Fedora 35) and Python3.11 (COPR testing) and runs fine on Python3.10.  I don't actually have a Python3.11 installation to try to run it on, but I believe the code is same for 3.10 and 3.11 anyway.

Thanks for accepting it.

PR #656, opened 2022-01-02
[DOC] Updated details for RHEL based systems

frankcrawford:
DOC: Update for standard RHEL based installs from EPEL  Only use COPR for development releases.
DOC: Finally added migration details for RHEL based systems.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -33,10 +33,8 @@ If the optional dependencies pylibacl and pyxattr are installed, rdiff-backup wi
 In older Linux distributions the rdiff-backup versions are of the 1.x series, which is not recommended for new installs anymore.
 Follow the instructions below to install the latest 2.x release of rdiff-backup.
 
-____
 CAUTION: rdiff-backup 1.x and 2.x aren't compatible and can't be mixed in server/client mode!
 See xref:docs/migration.adoc[how to migrate side-by-side].
-____
 
 === Ubuntu Focal or Debian Bullseye or newer (has 2.0)
 
@@ -61,9 +59,7 @@ sudo yum install rdiff-backup
 sudo yum install py3libacl pyxattr
 ----
 
-____
 NOTE: the last line is optional to get ACLs and EAs support.
-____
 
 === CentOS, AlmaLinux, Rocky Linux and RHEL 8 (from EPEL)
 
@@ -72,13 +68,11 @@ sudo dnf install epel-release
 sudo dnf --enablerepo=PowerTools install rdiff-backup
 ----
 
-____
 NOTE: you can add the option `--setopt=install_weak_deps=False` to the last line if you don't need ACLs and EAs support.
 You can install `python3-pylibacl` and `python3-pyxattr` also separately.
 Under RHEL, the repo to enable is https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository[codeready-builder-for-rhel-8-x86_64-rpms] in order to get access to pyxattr, instead of PowerTools.
 
 NOTE: This does not enable updates for `PowerTools`, check the distribution documentation for details on how to do this.
-____
 
 === Fedora 34+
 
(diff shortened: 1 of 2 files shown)


docs/migration.adoc
```diff
@@ -120,11 +149,39 @@ Then install the wrapper script to auto detect the version of rdiff-backup.
...
+sudo dnf copr enable frankcrawford/python2-rdiff-backup
+sudo dnf copr enable frankcrawford/python2-pyxattr
+sudo dnf copr enable frankcrawford/python2-pylibacl
+sudo dnf install python2-rdiff-backup
+----
 
 Once both version of rdiff-backup are installed side-by-side, you need to adapt your command line to make use of the rdiff-backup-wrap script that is used to auto-detect the version of rdiff-backup to be used.
 
-  rdiff-backup-wrap user@10.255.1.102:/source /destination
+  rdiff-backup-1 user@10.255.1.102:/source /destination
```
ericzolf, line 184: You can't do this because `rdiff-backup-wrap` is coming from the Debian instructions. The question is, if the rdiff-backup-wrap script could also detect between rdiff-backup and rdiff-backup-1 (Debian names them rdiff-backup2 vs. rdiff-backup). I think that the answer is negative. IMHO, you should simply move the paragraph to the Debian section as it's useless for you.

docs/migration.adoc
```diff
@@ -120,11 +149,39 @@ Then install the wrapper script to auto detect the version of rdiff-backup.
...
+
+Fedora (From COPR)
+ 
+----
+sudo dnf install dnf-plugins-core epel-release
+sudo dnf copr enable frankcrawford/python2-rdiff-backup
+sudo dnf copr enable frankcrawford/python2-pyxattr
+sudo dnf copr enable frankcrawford/python2-pylibacl
+sudo dnf install python2-rdiff-backup
+----
```
ericzolf, line 180: If I'm not mistaking, it's exactly the same as above, why not just [reference](https://docs.asciidoctor.org/asciidoc/latest/macros/xref/)?

ericzolf (changes requested):

ericzolf:
Side note: only `[DOC]` avoids lengthy checks...

README.adoc
```diff
@@ -66,21 +65,23 @@ ____
 *NOTE:* the last line is optional to get ACLs and EAs support.
```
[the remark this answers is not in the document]
frankcrawford: Done through both files.

docs/migration.adoc
frankcrawford: Yes, I see that now.  I've fixed the documentation

I will add a small change to the wrapper script to allow either version, maybe by expanding the test a little further.  Will test it a bit before I push the update.

docs/migration.adoc
```diff
@@ -139,7 +175,7 @@ The idea is to have a central backup server where multiple clients can connect t
 Because all the clients can't migrate at the same time, it must be made sure that the server is able to support multiple versions of rdiff-backup at the same time.
 
 ____
-*NOTE:* the same approach can be used to support multiple clients of 	different versions but the use case doesn't seem as useful, hence 	it is left to the interpretation of the reader.
+*NOTE:* the same approach can be used to support multiple clients of different versions but the use case doesn't seem as useful, hence it is left to the interpretation of the reader.
```
ericzolf, line 178: Sorry, I know I can be a pain, but you've forgotten to remove the asterisks (and also the 4 underscores before and after, I realize).

docs/migration.adoc
```diff
@@ -194,7 +230,7 @@ rdiff-backup --remote-schema 'ssh -C {h} rdiff-backup-{Vx}.{Vy} server' \
 ----
 
 ____
-*TIP:* for older versions of rdiff-backup, one could surely write a 	wrapper script mimicking the same behaviour, using something along the 	line of `$(rdiff-backup --version | awk -F'[.
+*TIP:* for older versions of rdiff-backup, one could surely write a wrapper script mimicking the same behaviour, using something along the line of `$(rdiff-backup --version | awk -F'[.
```
ericzolf, line 233: same thing for TIP...

docs/migration.adoc
frankcrawford: Done, I forgot to look at the other file!!!

docs/migration.adoc
frankcrawford: And fixed here too.

It now looks much better.

ericzolf (approved):

PR #658, opened 2022-01-03
Allow rdiff-backup-wrapper to test for multiple executables

frankcrawford:
A rewrite of rdiff-backup-wrapper to est for multiple possible executables, using the POSIX compliant command.

### DIFF
--- a/tools/misc/rdiff-backup-wrap
+++ b/tools/misc/rdiff-backup-wrap
@@ -30,8 +30,8 @@ done
 
 # Check if the source is remote.
 case "$SOURCE" in
-    *::*) REMOTE_SOURCE=1;;
-    *) REMOTE_SOURCE=0;;
+    *::*) REMOTE_SOURCE=1 ;;
+    *) REMOTE_SOURCE=0 ;;
 esac
 
 # Pick the right version of rdiff-backup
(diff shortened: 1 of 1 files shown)


tools/misc/rdiff-backup-wrap
```diff
@@ -35,14 +35,20 @@ case "$SOURCE" in
...
+do
+	if command -v $cmd 2>&1 > /dev/null
+	then
+		CMD=$cmd
+	fi
+done
+
+if [ $REMOTE_SOURCE -eq 1 ]; then
     if rdiff-backup --test-server "$SOURCE" 2>&1 > /dev/null; then
         CMD="rdiff-backup"
```
ericzolf, line 50:
```suggestion
        CMD="rdiff-backup"
    else
        # because the version we found previously might also not work
        if ! $CMD --test-server "$SOURCE" 2>&1 > /dev/null; then
            echo "ERROR:   Couldn't find a working version of rdiff-backup, tried 'rdiff-backup' and '$CMD'" >&2
            exit 1
        fi
```
I didn't test it but it seems worth a try.

tools/misc/rdiff-backup-wrap
```diff
@@ -35,14 +35,20 @@ case "$SOURCE" in
...
-    else
-        CMD="rdiff-backup2"
     fi
 fi
 
-"$CMD" $@
\ No newline at end of file
+"$CMD" $@
```
ericzolf, line 54:
```suggestion
"$CMD" "$@"
```
See man-page of bash and search for `$@`...

ericzolf (changes requested):

tools/misc/rdiff-backup-wrap
ericzolf: https://man7.org/linux/man-pages/man1/bash.1.html#PARAMETERS -> Special parameters

tools/misc/rdiff-backup-wrap
frankcrawford: Okay, I'll add that.  Previously such a test didn't exist, but this will make it more bulletproof.

tools/misc/rdiff-backup-wrap
frankcrawford: Tossing up should I also "exec" this, rather than just running it as a separate process?  What do you think?

ericzolf: Indeed I personally prefer to use `exec` for wrapper scripts, but it's more a personal preference (the overhead of bash is limited compared to the footprint of rdiff-backup).

ericzolf (approved):

ericzolf:
Thanks! Merged!

PR #659, opened 2022-01-03
[DOC] Add FAQ regarding support of exotic file systems

### LINKED ISSUE #595: The system cannot find the file specified (Network Share, Windows 10)
Problem Description
I get "The system cannot find the file specified" errors (FileNotFoundError: [WinError 2]) and rdiff-backup stops working. **The file which rdiff-backup is referring too is always located in the target directory.** I think it is important to mentioned that the backup target is a network share. The problem is reproducible, and executing rdiff-backup with the same parameters several times in a row always leads to another file which couldn't be found.

Which rdiff-backup was used?
I get this error with 2.0.5 and v2.1.0a1

Which operating system do you use?
Windows 10 20H2

More information
I have the feeling that this is a synchronization problem because the "missing" file is always available when I search it manually. Maybe the network share does not reflect immediately the already transferred files but rdiff-backup tries to access them (this is a guess).

Log File (rdiff-backup 2.0.5)
[example-error-log.txt](https://github.com/rdiff-backup/rdiff-backup/files/6622788/example-error-log.txt)

ericzolf:
DOC: add FAQ regarding support of too slow file systems like exFAT and ZFS on SMB, closes #595


dio4ev:
Looks fine for me and the documentation is clear.

ericzolf:
Thank you!

PR #663, opened 2022-01-12
Refactor time functions

ericzolf:
- merge local and remote get_mirror_time functions of repo
- slightly improve get_increment_times and add repo shadow function
- rename repo.init_quoting to setup_quoting to clarify call during setup
- add repo.setup_paths to define default paths of a repository
- slightly clean-up the Time module
- deprecate Time.setprevtime and Time.setprevtime_local
- deprecate Time.setcurtime, replacing with Gloabals variables

PR #664, opened 2022-01-14
Clarify usage of set vs. set_local

ericzolf:
- deprecate Globals.set as remote function, replace with set_local

Actually a small PR but it depends on the previous one #663


ericzolf:
Ready for (quick) review...

dio4ev:
Looks also fine :)

ericzolf:
Thanks, much appreciated!

PR #675, opened 2022-02-22
Output logs as text instead of bytes

### LINKED ISSUE #546: [BUG] tests fail with '_io.StringIO' object has no attribute 'buffer' when called with --buffer
## Bug summary

When calling the test cases with `--buffer` (for a cleaner output), the tests fail with something like:

```
    log.Log(
  File "/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.9/rdiff_backup/log.py", line 75, in __call__
    self.log_to_term(message, verbosity)
  File "/home/ericl/Comp/rdiff-backup/build/lib.linux-x86_64-3.9/rdiff_backup/log.py", line 91, in log_to_term
    termfp = sys.stderr.buffer
AttributeError: '_io.StringIO' object has no attribute 'buffer'
```

Call `rdiff-backup info` (version >= 2.1) or `rdiff-backup -v9` and replace the following line with the output, repeat for each environment impacted:

```yaml
2.0.6.etc...
```

## rdiff-backup call

How did you call rdiff-backup, with which parameters:

```
PATH=build/scripts-3.9:$PATH PYTHONPATH=build/lib.linux-x86_64-3.9 python testing/backuptest.py --buffer
```

The test fails because we try to output as bytes directly to the terminal's (stdout) buffer but the buffered tests don't have a buffer because they are redirected as string.

## More information

We should look into logging as string, and get rid of the bytes output (but only once I've cleaned up the whole thing).

ericzolf:
FIX: allow --buffer option while testing by logging to terminal as string, closes #546

### DIFF
--- a/src/rdiff_backup/log.py
+++ b/src/rdiff_backup/log.py
@@ -110,26 +110,25 @@ class Logger:
     def log_to_term(self, message, verbosity):
         """Write message to stdout/stderr"""
         if verbosity <= 2 or Globals.server:
-            termfp = sys.stderr.buffer
+            termfp = sys.stderr
         else:
-            termfp = sys.stdout.buffer
+            termfp = sys.stdout
         if Globals.get_api_version() < 201:  # compat200
             tmpstr = self._format(message, self.term_verbosity)
-            termfp.write(_to_bytes(tmpstr, encoding=sys.stdout.encoding))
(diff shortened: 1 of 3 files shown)


dio4ev:
Looks fine from my side :)

ericzolf:
Thanks!

PR #680, opened 2022-03-07
[DOC] Include both old and new CLI formats in examples

lostromb:
There's a bit of a documentation gap where the new CLI has been in place since 2.1, but all of the examples on the documentation page are still written in the legacy syntax. This PR updates the examples to put legacy and modern commands side-by-side. The intention is that new users can learn the new syntax, and legacy users can still get documentation on the legacy syntax while also getting some guidance on how to update their commands.

### DIFF
--- a/docs/examples.adoc
+++ b/docs/examples.adoc
@@ -6,7 +6,7 @@
 
 == Old vs. New Command Line Interface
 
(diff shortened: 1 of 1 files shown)


ericzolf:
The changes looked good but the underlying format was a bit broken because converted from markdown to asciidoc. I'll merge once the checks are finished. Thanks a lot!

PR #688, opened 2022-03-29
Fix issue with initial uncompressed backup

ericzolf:
It uncovered that the detection of an initial failed backup was done twice in an inconsistent manner


ikus060:
Seams all right.

dio4ev:
Looks good

ericzolf:
Hey, you're back! Thanks for the review.

PR #689, opened 2022-04-01
Do not compress metadata files when no compression

### LINKED ISSUE #402: [BUG] mirror_metadata are compressed even with --no-compression
Even when running `rdiff-backup --no-compression`, mirror_metadata files are created as gz file.

rdiff-backup 2.0.3

Looking at the way MetadataFile compression is handled, it's not using the global compression variable to termine if the files should be created with or without compression.

ericzolf:
FIX: (API 201 only) do not compress metadata files if the --no-compression option is given, BEWARE that such a repo can't be read by rdiff-backup 2.0, closes #402

I changed a bit my opinion and pushed the change already, but it does only apply to version 201 of the API, so that people can decide to keep their old repo version. Also with the old API version, repos with uncompressed metadata can be read (but not written).

### DIFF
--- a/src/rdiffbackup/meta/__init__.py
+++ b/src/rdiffbackup/meta/__init__.py
@@ -208,7 +208,8 @@ class FlatFile:
         """
         return cls._is_main
 
-    def __init__(self, rp_base, mode, check_path=1, compress=1, callback=None):
+    def __init__(self, rp_base, mode, check_path=1,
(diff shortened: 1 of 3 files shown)


ericzolf:
Do I get a review?

ikus060:
Seams good. So many little changes to get this fixed. No wonder I was not able to figure it out last time I check.

Thanks for the change.

PR #691, opened 2022-04-12
remove support for python v2 from rdiff-backup-delete

fireartist:
was broken with python 2.7

### DIFF
--- a/src/rdiff-backup-delete
+++ b/src/rdiff-backup-delete
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # Copyright (C) 2020 Patrik Dufresne<info@patrikdufresne.com>
 #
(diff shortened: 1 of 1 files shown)


ikus060:
If it's working for you. I'm fine with your changes

ericzolf:
Looks good to me as well. Would you mind perhaps creating a test case for rdiff-backup-delete, we're missing at least one? It could be as part of this PR or in a separate one, as you wish.

ikus060:
@ericzolf Don't we have these tests already ?
https://github.com/rdiff-backup/rdiff-backup/blob/master/testing/rdiffbackupdeletetest.py

ericzolf:
My bad, I was convinced of the contrary. We're even better than I thought! 🤗 Then let's merge! Thanks @fireartist

ikus060:
Perfect !

Thanks @fireartist Your contribution is mush appreciated.

PR #694, opened 2022-04-17
Fix restoring of sub-dir with include/exclude

### LINKED ISSUE #463: [BUG] restoring a subdirectory doesn't play well with include/exclude filters
## Bug summary

rdiff-backup crashes due to a Python OSError exception while trying to restore a subdirectory and I also specify include/exclude filters.

rdiff-backup 2.0.5, Python 3.8.5, Fedora 32

Call `rdiff-backup -v9` and paste the output between the backquotes below, repeat for each environment impacted:

```
2020-09-11 22:59:21.708502 +0200  <CLIENT-7215>  Fatal Error: No arguments given
2020-09-11 22:59:21.703167 +0200  <CLIENT-7215>  Using rdiff-backup version 2.0.5
2020-09-11 22:59:21.703273 +0200  <CLIENT-7215>  	with cpython /usr/bin/python3 version 3.8.5
2020-09-11 22:59:21.708363 +0200  <CLIENT-7215>  	on Linux-5.8.6-201.fc32.x86_64-x86_64-with-glibc2.2.5, fs encoding utf-8
```

## rdiff-backup call

I have a fully reproducible test case. First, let's create a few files and directories to play with:

```
$ mkdir -p backup original/dir{1,2}
$ cd original
$ echo file1 | tee dir1/file1.txt > dir2/file1.txt
$ echo file2 | tee dir1/file2.txt > dir2/file2.txt
$ echo base > base.txt
$ cd -
```

Create our first backup in the empty backup directory, then overwrite a file with garbage and :
```
$ rdiff-backup original backup
$ echo garbage > original/dir1/file2.txt
```

We can now try to selectively restore dir1:
```
$ rdiff-backup -r now --force --include original/dir1/file2.txt --exclude '**' backup/dir1 original/dir1
Exception '[Errno 39] Directory not empty: b'original/dir1'' raised of class '<class 'OSError'>':
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 395, in error_check_Main
    Main(arglist)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 417, in Main
    take_action(rps)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 373, in take_action
    Restore(rps[0], rps[1], 1)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 720, in Restore
    restore.Restore(
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 42, in Restore
    TargetS.patch(target, diff_iter)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 337, in patch
    ITR.Finish()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/rorpiter.py", line 277, in Finish
    to_be_finished.end_process()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 762, in end_process
    self.base_rp.rmdir()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/rpath.py", line 1222, in rmdir
    self.conn.os.rmdir(self.path)

  File "/usr/bin/rdiff-backup", line 32, in <module>
    rdiff_backup.Main.error_check_Main(sys.argv[1:])
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 395, in error_check_Main
    Main(arglist)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 417, in Main
    take_action(rps)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 373, in take_action
    Restore(rps[0], rps[1], 1)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/Main.py", line 720, in Restore
    restore.Restore(
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 42, in Restore
    TargetS.patch(target, diff_iter)
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 337, in patch
    ITR.Finish()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/rorpiter.py", line 277, in Finish
    to_be_finished.end_process()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/restore.py", line 762, in end_process
    self.base_rp.rmdir()
  File "/usr/lib64/python3.8/site-packages/rdiff_backup/rpath.py", line 1222, in rmdir
    self.conn.os.rmdir(self.path)
OSError: [Errno 39] Directory not empty: b'original/dir1'
```

I expected `rdiff-backup` to restore the `dir1/file1.txt` instead of crashing. Even worse, file2.txt disappeared completely after the crash, but it fortunately still be recovered from the backup directory:
```
$ find original/ -type f
original/dir2/file2.txt
original/dir2/file1.txt
original/dir1/file1.txt
$ ls backup/dir1
file1.txt  file2.txt
```

## More information

Here's the `-v9` output:

[rdiff-backup.log](https://github.com/rdiff-backup/rdiff-backup/files/5211604/rdiff-backup.log)

Possible workarounds:
- restore the entire directory, losing changes to the other files: `rdiff-backup -r now --force backup/dir1 original/dir1`
- use include/exclude filters without subdirectories: `rdiff-backup -r now --force --include 'original/dir1/file2.txt' --exclude '**' backup original`
- use rsync, which offers its own filters (this only works for the ltest backed up version, not older ones): `rsync -a --include 'file2.txt' --exclude '**' backup/dir1/ original/dir1`

The test case might seem convoluted, since `rdiff-backup backup/dir1/file1.txt original/dir1/file1.txt` is easier, but I discovered this using `rdiff-backup` with a globbing filelist to restore several file in the `/root` directory - it was a bit scary to see it tried to remove the `/root` directory, running with root privileges.

I've been using rdiff-backup since at least 2009, many thanks for reviving the project!

ericzolf:
DEV: get rid of unused RepoShadow.update_quoting function
DEV: introduce properly reference path, index, inc and type (partially renamed from restore_....)
FIX: restoring from sub-path while selecting is now forbidden to avoid data loss, workaround documented in FAQ, closes #463


ericzolf:
Please review! @lpancescu you're a candidate! Thanks.

lpancescu:
Hi @ericzolf, many thanks for looking into this! I had the entire distro wiped out at least 3 times by something similar like `rdiff-backup -r now --force --include /home/lpancesc --include /usr/local --exclude '**' /mnt/backup/ /`, although somtimes it works as expected (fresh install, restore from the same backup). Preventing this from occurring is good.

I'm not familiar at all with the rdiff-backup code, so I deduced some things from the variable names, but from looking at the patch, it looks fine to me. Something I don't understand: rdiff-backup seems to use bytes (not strings) for the paths, but the 'base' in increments seems to be a string? I had some older files from an old Windows machine with German characters (encoded 8859-1 on the disc), copied as they were by `cp -a`. They always produced exceptions during `rdiff-backup --verify` because the bytes don't decode to UTF-8, I ended up writing a script to scan for such file names and rename them to file names using valid UTF-8 encoding. In such situations, couldn't it happen that two different increments with invalid UTF-8 encoding are transformed to the same base directory after calling `string.decode(errors='replace')`? Is the base directory only used for printing on the screen or logging?

ericzolf:
The base shouldn't be a string, but an error is always possible. What makes you think so?

Actually, the only times where paths are transformed from bytes to strings or vice-versa, should be input (from the command line parameters) or output (log messages), else it should remain bytes all the way. There are a few exceptions in the test scripts, but they shouldn't be relevant. So the conversion can't lead to files losses.

I'm a bit surprised that `rdiff-backup --verify` failed on non-UTF-8 compliant filenames. With which version of rdiff-backup did it happen? I had similar issues with rdiff-backup 1.2/1.3 IIRC but it shouldn't happen with 2.x (perhaps early versions of 2.0).

ikus060:
> failed on non-UTF-8 compliant filenames

I'm missing the context. here, but I've never seen rdiff-backup failed on non-utf-8. It's an essential feature for my use case here and rdiff-backup is very resilient. Since v1.2.8 I never had trouble with non-utf8 filenames on Linux or MacOs operating system. On Windows it might be another subject.

lpancescu:
Hi @ericzolf , sorry for the late reply. I saw `mirror_path.dirsplit()[1].decode(errors="replace")` in [repository.cpp:443](https://github.com/rdiff-backup/rdiff-backup/blob/master/src/rdiffbackup/locations/repository.py#L433) and assumed `mirror_path.dirsplit()[1]` returns a bytes object which is then decoded as utf-8, replacing invalid characters with question marks or something similar.

Regarding non-utf-8 filenames, it was probably version 1.2.8 on Debian Buster, I see verify and restore work fine with rdiff-backup 2.0.5, even with the same file name with different encodings:

```
>>> for encoding in 'utf-8', 'iso-8859-15':
...     with open('/tmp/src/Böblingen.txt'.encode(encoding), 'w') as f:
...             f.write('German city')
```

ericzolf:
Ha, ha, Böblingen is my home town... Anyway, `get_increments` is only a function used by `list increments` for output to the end-user, so all good. I'll merge now.

PR #701, opened 2022-05-01
[DOC] Clarify FAQ instructions to set timezone in Windows (#692)

### LINKED ISSUE #692: [DOC] FAQ entry on using UTC as TZ on Windows seems outdated
The FAQ, which is referred to in several other places in the documentation, has [an entry](<https://github.com/rdiff-backup/rdiff-backup/blob/master/docs/FAQ.adoc#22-how-can-rdiff-backup-use-utc-as-the-timezone>) about using UTC as timezone:

> 22. How can rdiff-backup use UTC as the timezone?
> Like other Unix and Python programs, rdiff-backup respects the `TZ` environment variable, which can be used to temporarily change the timezone. On Unix, simply set `TZ=UTC` either in your shell, or on the command line used to run rdiff-backup. On Windows, the command USE `TZ=UTC` sets the `%TZ%` environment variable, and can be used either in a batch script, or at the DOS prompt.

This is an important use case, but I can't get the Windows directives to work (under Windows 10 at least): there is no `TZ` environment variable preset and there is no `USE` command as far as I know.

If it is correct, it should be clarified.
I think that it is outdated though, in which case it should be updated, possibly with a way to set a specific timezone in the scope of a process, like in Linux. The only ways I know of in Windows modify the system timezone.

desseim:
As discussed in #692, clarifies the FAQ instructions on how to set the timezone on Windows. Also slightly rewords the entry, adding the reference to the format of the Windows `TZ` variable.

Let me know in case the style or the content is not fully appropriate or can be improved and I'll update it.

Resolves #692

### DIFF
--- a/docs/FAQ.adoc
+++ b/docs/FAQ.adoc
@@ -350,11 +350,12 @@ After the text "Bad index order", the error message will indicate which files ha
 
 If you get this message for an unrelated reason, try contacting the mailing list.
 
-== How can rdiff-backup use UTC as the timezone?
+== How can rdiff-backup use a specific timezone, like UTC?
 
 Like other Unix and Python programs, rdiff-backup respects the `TZ` environment variable, which can be used to temporarily change the timezone.
-On Unix, simply set `TZ=UTC` either in your shell, or on the command line used to run rdiff-backup.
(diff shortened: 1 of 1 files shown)


ericzolf:
Looks good to me, I'll merge once the tests have run. Thanks a lot!

PR #702, opened 2022-05-13
Implement forced regress

### LINKED ISSUE #10: --regress switch
Enhancement request. While --check-destination-dir regresses a damaged archive, there is no official way to regress an undamaged archive. I have written [rdiff-backup-regress.sh](http://www.timedicer.co.uk/programs/help/rdiff-backup-regress.sh.php) which does this but in an unofficial way. There are legitimate reasons why one might want to regress an undamaged archive (including wishing to reverse the bloating that can happen when some large data gets unintentionally included in a new backup session). 

At its simplest this could be a --regress switch which does what --check-destination-dir does but without first requiring that there be more than one current_mirror file. If this doesn't break anything...

ericzolf:
FIX: regress option in API 201 was actually still using the old API function

NEW: (API 201 only) regress action can be forced on a sane repository, i.e. the last backup can be removed, closes #10

CHG: start to introduce more specific return codes (1 for error, 2 for warning, 4 for single file error, etc), which can be combined (e.g. 3 for error _and_ warning), this will take time to introduce everywhere

It's a bit mixing different topics but this became necessary to
differentiate the rejection of forced regress, which is a warning,
from an actual failure to regress


ericzolf:
Any reviewer volunteering?

src/rdiff_backup/Main.py
```diff
@@ -74,7 +74,7 @@ def _main_run(arglist, security_override=False):
 
     # validate that everything looks good before really starting
     ret_val = action.pre_check()
-    if ret_val != 0:
+    if ret_val & Globals.RET_CODE_ERR:
```
ikus060, line 77: If my understanding is right, the process will continue running is pre_check, or check return RET_CODE_FILE ?

What is the rational behind this ?

src/rdiffbackup/actions/__init__.py
```diff
@@ -558,33 +561,42 @@ def _operate_regress(self, try_regress=True, noticeable=False):
...
         else:
             if self.repo.needs_regress():
                 if not try_regress:
-                    return 1
+                    return Globals.RET_CODE_ERR
                 log.Log("Previous backup seems to have failed, regressing "
                         "destination now", log.WARNING)
                 return self.repo.regress()
+            elif force:
+                if self.repo.force_regress():
```
ikus060, line 588: That function already existed ?

ikus060 (approved):
The change is more related to return code modification then forece-regress if you ask me. ;)

src/rdiff_backup/Main.py
ericzolf: As I wrote, it's not fully implemented but the idea is that only a "real" error stops the program workflow, a warning or a single file error (especially at backup time) doesn't stop the whole backup.

IMHO it's not a change of logic: already today, the backup doesn't stop just because a few files can't be saved (e.g. because not readable).
I only formalize the approach with predefined return codes.

src/rdiffbackup/actions/__init__.py
ericzolf: No, you missed the crunchy part:smile:

Added line 299 of repository.py

PR #704, opened 2022-05-15
Upgrade Python and Visual Studio under Windows

ericzolf:
CHG: rdiff-backup is now compiled with Python 3.10.4 and Visual Studio 2022, users shouldn't notice any difference


ericzolf:
Merging without review, no code change anyway.

PR #705, opened 2022-05-20
Fix comparaison code inconsistency

### LINKED ISSUE #643: [BUG] compare action uses the SingleRepoSetGlobals class even though it compares with a directory
## Bug summary

The compare action uses the `fs_abilities.SingleRepoSetGlobals` class even though it compares with a directory, this is aligned with the previous approach using the old `fs_abilities.single_set_global` function but that doesn't sound right.

Development version

## More information

It's unclear at this stage if the Repo2Dir or the Dir2Repo SetGlobals class would be best, but it would need to keep repo _and_ directory read-only (using the "ReadDir" with an optional Repo might be the right approach).

ericzolf:
DEV: use file system object for two file systems in comparaison action, closes #643

DEV: split RET_CODE_FILE in RET_CODE_FILE_ERR AND _WARN so that file warnings can be used to detect comparaison differences

A purely internal change to make code slightly more consistent.

### DIFF
--- a/docs/rdiff-backup.1.adoc
+++ b/docs/rdiff-backup.1.adoc
@@ -690,7 +690,8 @@ Also note that they can be combined, so that for example a return code 3 might b
 0 - OK:: the action was completely successful
 1 - ERROR:: something fatal happened, the whole action failed
 2 - WARNING:: any kind of unexpected behavior without complete failure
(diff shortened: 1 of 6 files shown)


ericzolf:
Merging without review...

PR #712, opened 2022-08-01
Fix AsciiDoc syntax for literal text

desseim:
Fix a couple instances of monospace text which wasn't made [literal](https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace/) and as a consequence was being rendered wrongly as AsciiDoc text formatting was applied.

This syntax must have come from the previous markdown version of the file.


ikus060 (approved):

ericzolf:
Thanks!

PR #719, opened 2022-09-07
Update versions of different dependencies

ericzolf:
CHG: embed Python 3.10.7 instead of 3.10.4 in Windows rdiff-backup, shouldn't impact end-users DEV: remove dependency on importlib-metadata 1.x, it can be now any version

- reformat the changelog which was slightly broken
- add documentation on where to change dependency versions (all over the place)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,8 +16,8 @@ defaults:
     shell: bash
 
 env:
-  WIN_PYTHON_VERSION: 3.9.1
-  WIN_LIBRSYNC_VERSION: v2.2.1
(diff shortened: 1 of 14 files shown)


ikus060 (approved):

ericzolf:
I also had to drop Python 3.6 and add Python 3.10 as Python 3.6 made the release pipeline fail. I might have found a solution but as I anyway planned to drop 3.6 as it's not supported anymore, it just happens a bit earlier.

ericzolf:
OK, there was just a hick-up in the pipeline of GitHub, which made the upload fail.

PR #728, opened 2022-09-08
Allow again for Python 3.6

ericzolf:
Still not a recommended version as it isn't regularly tested but it should continue to work. This said, it will most probably be the last non-bug-fix release to support it.


ericzolf:
@frankcrawford for you to review!

frankcrawford (approved):
Looks correct and thanks for that.

Issue #731, opened 2022-09-10
[BUG] rdiff-backup.exe doesn't find its modules under Windows


ericzolf:
## Bug summary

The Windows executable built from the pipeline is failing with

```
> rdiff-backup.exe --version
Traceback (most recent call last):
  File "rdiff-backup", line 19, in <module>
ModuleNotFoundError: No module named 'rdiffbackup'
[1156] Failed to execute script 'rdiff-backup' due to unhandled exception!
```

## Version, Python, Operating System

2.1.2+

## More information

I thought that the issue was due to the version of PyInstaller, where 5.0/5.1 does work, but 5.2/5.3 doesn't. This is at least the state on my Vagrant VM, _but_ there the binary doesn't fail miserably, it just doesn't find the "action" modules, so it might be a secondary issue.

Sadly, the logs are too old that I could check the version of PyInstaller last successfully working.

I'm slightly at loss what the relevant difference between the pipeline and my VM is.

ericzolf:
On my VM, `rdiff-backup info`:

```
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - rdiff-backup.exe
  - info
  parsed: null
  version: 2.1.2b2.dev3+gf8e4dd5
python:
  executable: C:\Users\vagrant\Develop\rdiff-backup\dist\rdiff-backup-2.1.2b2.dev3+gf8e4dd5-64\rdiff-backup.exe
  name: cpython
  version: 3.10.4
system:
  fs_encoding: utf-8
  platform: Windows-10-10.0.17763-SP0
...
```

The pipeline has:

* Python 3.10.7 (tags/v3.10.7:6cc6b13, Sep  5 2022, 14:08:36) [MSC v.1933 64 bit (AMD64)]
* Microsoft Windows Server 2022, 10.0.20348, Datacenter

So perhaps the Python version difference is relevant...

ericzolf:
Upgraded my VM to 3.10.7, and the resulting build is failing like in the pipeline, so probably two issues:

1. too new version of Python
2. too new version of PyInstaller

ericzolf:
By downgrading Python to 3.5 and PyInstaller to 5.1, I come back to a working combination, but this is only a workaround so this issue remains open.

ericzolf:
If you're interested, follow the discussion at https://github.com/orgs/pyinstaller/discussions/7078

ericzolf:
[...]

PR #732, opened 2022-09-10
Fix missing modules in Windows binary

ericzolf:
I had basically to do two things to address #731

1. downgrade PyInstaller from 5.2/5.3 to 5.1
2. downgrade Python 3.10.7 to 3.10.5

Note that I won't close the issue with this PR, I consider this to be only a workaround and hope to find a better solution during the beta phase.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -83,11 +83,10 @@ jobs:
       with:
         python-version: ${{ env.WIN_PYTHON_VERSION }}
         architecture: ${{ matrix.arch }}
-    # FIXME PyInstaller 5.2 and 5.3 fail to grab all modules
     - name: Install dependencies
       run: |
         python.exe -VV
-        pip.exe install --upgrade pywin32 pyinstaller==5.1 wheel certifi setuptools-scm tox PyYAML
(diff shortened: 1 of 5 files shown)


ericzolf:
OK, this is ready for review, we're back to 3.10.7, which was important from a security perspective. Getting PyInstaller back to a newer version will take more effort, something for another PR.

ericzolf:
OK, now really no more changes. Please review that we can merge and release a new working beta version.

ericzolf:
OK, sorry, the circular dependency was much easier to fix than I thought, so I've created PR #734 which makes this PR obsolete.

ericzolf:
Bah, change leads to an error in the tests, so back to this PR, I'll fix the other one later on. It's now more important that we get more feedback on the beta.

ikus060:
Not sure what is the difference with #734

ericzolf:
It's only a workaround because I couldn't use PyInstaller > 5.1, #734 _additionally_ removes a circular dependency so that PyInstaller continues to work.

ericzolf:
And, now, after reviewing your comments on #734, I understand better your confusion. :shrug: I've got it sorted.

PR #734, opened 2022-09-11
fix meta circular dependency 

ericzolf:
Closes definitely #731 and makes #732 obsolete, which it contains.

### DIFF
--- a/tools/win_build_rdiffbackup.sh
+++ b/tools/win_build_rdiffbackup.sh
@@ -18,26 +18,18 @@ else
 fi
 
 PYEXE=python.exe
-PYINST=PyInstaller.exe
+TOXEXE=tox.exe
 if [[ -n ${PYDIRECT} ]]
 then
 	py_dir=C:/Python${bits}
 	PYEXE=${py_dir}/${PYEXE}
-	PYINST=${py_dir}/Scripts/${PYINST}
+	TOXEXE=${py_dir}/Scripts/${TOXEXE}
 fi
 
 LIBRSYNC_DIR=${HOME}/librsync${bits}
 export LIBRSYNC_DIR
 
-ver_name=rdiff-backup-$(${PYEXE} setup.py --version)
-py_ver_brief=${PYTHON_VERSION%.[0-9]}
+RDIFF_BACKUP_VERSION=rdiff-backup-$(${PYEXE} setup.py --version)-${bits}
+export RDIFF_BACKUP_VERSION
 
-${PYEXE} setup.py bdist_wheel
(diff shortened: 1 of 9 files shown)


ericzolf:
The error in the Windows tests seems to have been just a hick-up, still let's first merge #732 then this one, so that I can roll-back if necessary.

tools/win_build_rdiffbackup.sh
```diff
@@ -18,26 +18,18 @@ else
...
-	PYINST=${py_dir}/Scripts/${PYINST}
+	TOXEXE=${py_dir}/Scripts/${TOXEXE}
 fi
 
 LIBRSYNC_DIR=${HOME}/librsync${bits}
 export LIBRSYNC_DIR
 
-ver_name=rdiff-backup-$(${PYEXE} setup.py --version)
-py_ver_brief=${PYTHON_VERSION%.[0-9]}
+RDIFF_BACKUP_VERSION=rdiff-backup-$(${PYEXE} setup.py --version)-${bits}
```
ikus060, line 32: Would it be possible to move that into tox.ini ? So that someone that want to build a exe, could simply call `tox -e buildexe`

ikus060: This is a nice to have.

tox_win.ini
```diff
@@ -97,3 +96,14 @@ exclude =
     __pycache__
     build
 max-complexity = 40
+
+[testenv:buildexe]
+# add hook file to hooks-dir for each new plugin manager!
+commands =
```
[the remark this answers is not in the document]
ericzolf: It is already in the default dependencies (see `[testenv]` -> `deps =`).
Fix the pyinstaller dependency: yeah, let's do it, I've added a note to 730, we should anyway reconsider and clean everything.

tox_win.ini
```diff
@@ -7,7 +7,7 @@
 # Use tox_slow.ini for longer running tests.
 
 [tox]
-envlist = py, flake8
+envlist = py, flake8, buildexe
```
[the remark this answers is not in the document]
ericzolf: Good point. Done.

tools/win_build_rdiffbackup.sh
ericzolf: I wouldn't know if it's possible in tox.ini... I couldn't find anything at https://tox.wiki/en/latest/config.html

ericzolf:
Now that #732 is merged, the scope of this PR should be much clearer (and smaller).

ericzolf:
Replaced by #736

PR #735, opened 2022-09-14
[DOC] Mention EuroLinux

aronowski:
Mention [EuroLinux](https://en.euro-linux.com/eurolinux/what-is/) as a supported RHEL replica.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -51,7 +51,7 @@ sudo apt update
 sudo apt install rdiff-backup
 ----
 
-=== CentOS and RHEL 7 (from EPEL)
+=== CentOS, EuroLinux and RHEL 7 (from EPEL)
 
 ----
 sudo yum install epel-release
@@ -61,7 +61,7 @@ sudo yum install py3libacl pyxattr
 
 NOTE: the last line is optional to get ACLs and EAs support.
 
-=== CentOS, AlmaLinux, Rocky Linux and RHEL 8 (from EPEL)
+=== CentOS, AlmaLinux, EuroLinux, Rocky Linux and RHEL 8 (from EPEL)
 
 ----
 sudo dnf install epel-release
--- a/README.adoc
+++ b/README.adoc
@@ -51,7 +51,7 @@ sudo apt update
 sudo apt install rdiff-backup
 ----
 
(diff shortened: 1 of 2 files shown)


ericzolf:
Hesitating, I didn't know about this distro, and don't want to list zillions of obscure distros, just because they happen to exist. Do you have some numbers/arguments for me?

aronowski:
Naturally!

The company behind the distribution has a way faster release cycle than let's say Rocky. We can compare them on Distrowatch:
- https://distrowatch.com/table.php?distribution=eurolinux

Feature | 9.0 | 8.6
-- | -- | --
Release Date | 2022-06-14 | 2022-05-11

- https://distrowatch.com/table.php?distribution=rocky

Feature | 9.0 | 8.6
-- | -- | --
Release Date | 2022-07-14 | 2022-05-16

EuroLinux usually releases one business day after Red Hat does

dw-eurolinux-8-6

EuroLinux is also in the TOP 100 list on Distrowatch. That's got to mean something.

README.adoc
```diff
@@ -51,7 +51,7 @@ sudo apt update
 sudo apt install rdiff-backup
 ----
 
-=== CentOS and RHEL 7 (from EPEL)
+=== CentOS, EuroLinux and RHEL 7 (from EPEL)
```
ikus060, line 54: I would simply rephrase that to "RedHat based distribution". No need to list all the RedHat distribution that exists.

And do something similar for Debian, Ubuntu, etc.: "Debian based distribution"

ericzolf: I like the term "RHEL Replica" used by @aronowski , it differentiates Centos & Co from Fedora, but else I agree on the principle.

Side note: Red Hat, with a space in-between.

ikus060: Good for me.

ericzolf: @aronowski do you mind updating your PR accordingly? You could link to https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux_derivatives (I'm also fine with "RHEL derivatives" BTW, to be consistent).

aronowski: @ericzolf Done. Thank you!

ericzolf:
Thanks for the effort. Merging...

PR #736, opened 2022-09-14
Remove circular dependency in meta

### LINKED ISSUE #731: [BUG] rdiff-backup.exe doesn't find its modules under Windows

ericzolf:
FIX: remove circular dependency in meta to rpath to allow for newer PyInstaller under Windows, closes #731

Corrected version of PR #734

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -1440,13 +1440,13 @@ class RPath(RORPath):
 
 
 class MaybeGzip:
-    """Represent a file object that may or may not be compressed
+    """
+    Represent a file object that may or may not be compressed
 
     We don't want to compress 0 length files.  This class lets us
     delay the opening of the file until either the first write (so we
     know it has data and should be compressed), or close (when there's
     no data).
-
     """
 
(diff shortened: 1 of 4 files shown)


ikus060 (approved):

ericzolf:
Again an hick-up in the Windows testing, that starts to worry me a bit, as it's happening more and more apparently. Let's keep an eye on it...

PR #743, opened 2022-09-22
Consolidate all dependencies into one place

### LINKED ISSUE #730: [ENH] consolidate the many places where dependencies are listed

As I need to modify dependencies, I noticed in how many different places they are defined.

Find a way to centralize the dependencies

I've already started to document where to look for dependencies, but it's still not perfect.

## Additional context

I'm a lazy bastard :smile_cat:

ericzolf:
DEV: consolidate all Python and binary dependencies into requirements.txt/requs/*.txt respectively bindep.txt, see DEVELOP.adoc for details, closes #730


ikus060:
@ericzolf I don't really agree with this approach. I think the new way to define multiple dependencies is to use setuptools extras_require. This way you may define multiple dependencies depending on you need. Usage of `requirements.txt` is considered deprecated to my knowledge.

Using `extras_require` you may then install extra dependencies with `pip install rdiffweb[test]` or similar.

ericzolf:
i would tend to disagree:
1. I can't find any sign that requirements.txt should be deprecated, it's still documented with [latest pip](https://pip.pypa.io/en/stable/reference/requirements-file-format/) without any sign otherwise. The document even shows that you can use extras in requirements.txt.
2. IMHO you're mixing up two things, the extras_require is meant for extra features of the actual product, they could be a good fit for xattr/libacl because they are dependencies allowing for extra features _but_ they only work on Linux, so I would hesitate to declare them because it would require extra handling (and that would be a fully different topic).
3. requirements.txt on the other hand defines any kind of dependencies, also (and especially) for development purposes, or to declare a complete Python environment. If I want to develop rdiff-backup, I need dependencies but I actually do _not_ want to install rdiff-backup to install those dependencies. This is where requirements.txt comes in place.

ericzolf:
This said, I just learned about https://peps.python.org/pep-0508/#environment-markers

ericzolf:
We need to come to an agreement but else I'm done with my changes. This said, I've added the optional dependencies as extras_require to setup.py. As explained, I still don't think that the dev/test/build dependencies belong in there.

ikus060:
> As explained, I still don't think that the dev/test/build dependencies belong in there.

Actually, it's wildly use in projects t define the testing dependencies as a replacement for test_requirements that is now deprecated.

For build dependencies, you might take a look at `project.toml` which is the new thing to define build system dependencies.

I'm not here to convince you, I'm just sharing my little knowledge on the subject. Python packaging as been stagnant for decade and recently of lot of stuff are moving. So I'm trying to keep up with it. So while you are modifing it, better go in the right direction. Another new thing to work with dependencies and I did not had time to work with yet it: https://pdm.fming.dev/latest/

frankcrawford:
Can I just chime in here, `project.toml` is the great new thing, but it is a different structure, and I think it is a bit late to move to it.  In terms of packaging for Fedora, I've seen they have a whole separate lot of macros to handle it vs the current scheme.

It is probably something you need to look at for a future release.

ericzolf:
It's not the first time I think about setup.cfg or indeed project.toml, but we have quite some code in setup.py which I don't think can be expressed in config files, so that I never thought it worth the trouble.
Pdm BTW is Python 3.7+ so that it's out of question for this time, after having promised to keep 3.6 compatibility.
And there is another reason for keeping requirements.txt: GitHub is scanning it for vulnerabilities.
OK, I'll merge. Good discussion, gentlemen, I truly appreciate your imput, it keeps me on my feet.

PR #746, opened 2022-09-24
Update RPM specs to latest requirements and standards

frankcrawford:
Clean up the RPM files, although I'm not certain the generic one will always work, as I don't have a different system to test on.

### DIFF
--- a/tools/rdiff-backup.spec.template-fedora
+++ b/tools/rdiff-backup.spec.template-fedora
@@ -1,33 +1,28 @@
-# This spec file works for Fedora, but is NOT the same as the official one
+# This spec file works for Fedora, but is NOT necessarily the official one
 # used to build the package.
 
-%define PYTHON_NAME %((rpm -q --quiet python3 && echo python3) || echo python)
-%define PYTHON_VERSION %(%{PYTHON_NAME} -c 'import sys; print(sys.version[:3])')
-%define NEXT_PYTHON_VERSION %(%{PYTHON_NAME} -c 'import sys; print("%d.%d" % (sys.version_info[0], sys.version_info[1]+1))')
+%global gittag {{ version }}
 
-Version: {{ version }}
+Version: %{gittag}
 Summary: Convenient and transparent local/remote incremental mirror/backup
 Name: rdiff-backup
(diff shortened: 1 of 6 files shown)


tools/rdiff-backup.spec.template-fedora
```diff
@@ -42,75 +37,31 @@ securely back a hard drive up to a remote location, and only the
...
+%{python3_sitearch}/rdiff_backup/
+%{python3_sitearch}/rdiff_backup-*.egg-info
+%{python3_sitearch}/rdiffbackup/
 %{_datadir}/bash-completion/completions/rdiff-backup
-%ghost %{_libdir}/python%{PYTHON_VERSION}/site-packages/rdiff_backup/__pycache__/*.pyo
-%doc CHANGELOG.adoc COPYING README.adoc docs/*.adoc
+%doc CHANGELOG.adoc README.adoc
+%doc docs/credits.adoc docs/DEVELOP.adoc docs/examples.adoc
+%doc docs/FAQ.adoc docs/migration.adoc
+%doc docs/Windows-README.adoc docs/Windows-DEVELOP.adoc
```
ericzolf, line 61: I don't think we need to put the Windows documentation in the Linux package

ericzolf:
I will mostly trust you on this one, but shouldn't the changelogs contain also/at least the version 2.0.5? I also hesitate to remove the full history and former maintainers' name, did you have a specific reason for doing it?

tools/rdiff-backup.spec.template-fedora
frankcrawford: True, although it is actually me being lazy, as otherwise I need to remove them in the install statement.
I'll take them out, as I noted in the Fedora packaging guidelines, it also says to do not include "non-essential" docs, such as those for Windows.

frankcrawford:
If you want for the generic one, I'll include it all,  For Fedora the recommendation is to trim it regularly.

The other reasoning is that we haven't really added them in for the recent releases so I should fake a few entries.  After all it was just a template.

ericzolf:
Looks good to me, merging! Thanks.

PR #752, opened 2022-10-01
Fix terminating PermissionError under Windows

### LINKED ISSUE #619: [BUG] Terminating error occurs in the handling of a `PermissionError` produced by certain files in use
## Bug summary

An error occurs in the handling of a `PermissionError` produced by a Firefox lock file being in use.

```yaml
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - C:\Program Files\rdiff-backup\rdiff-backup.exe
  - info
  version: 2.1.0a1
python:
  executable: C:\Program Files\rdiff-backup\rdiff-backup.exe
  name: cpython
  version: 3.9.1
system:
  fs_encoding: utf-8
  platform: Windows-10-10.0.19041-SP0
...
```

## rdiff-backup call

```
rdiff-backup.exe -v5 backup --exclude-symbolic-links --include-globbing-filelist C:/filelist C:/ D:/
```

`rdiff-backup` threw a terminating `PermissionError`, aborting the backup and leaving the repository in an inconsistent state.

I expected the file to produce a non-terminating `ListError` and for the backup to continue, skipping this file.

## More information

Several other files being in use by the `System` process produce an `[Errno 13] Permission denied` error, but (I guess) the terminating `PermissionError` is caught and only a non-terminating `ListError` is eventually thrown and the backup continues afterwards.
But when backing up this `parent.lock` file being opened by `Firefox`, there seems to be an error while processing the `PermissionError`, aborting the script:
```
ListError: 'Users/TheUser/AppData/Local/Packages/MicrosoftWindows.Client.CBS_xxxxxxxxxxxxx/Settings/settings.dat.LOG2' [Errno 13] Permission denied: 
b'C:/Users/TheUser/AppData/Local/Packages/MicrosoftWindows.Client.CBS_xxxxxxxxxxxxx/Settings/settings.dat.LOG2'
ListError: 'Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock' [Errno 13] Permission denied: 
b'C:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock'
  File "rdiff_backup\rpath.py", line 863, in delete
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: b'D:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock'
During handling of the above exception, another exception occurred:
  File "rdiff-backup", line 32, in <module>
  File "rdiff_backup\Main.py", line 38, in main_run_and_exit
  File "rdiff_backup\Main.py", line 95, in _main_run
  File "rdiffbackup\actions\backup.py", line 143, in run
  File "rdiff_backup\backup.py", line 55, in Mirror_and_increment
  File "rdiff_backup\backup.py", line 210, in patch_and_increment
  File "rdiff_backup\rorpiter.py", line 146, in __call__
  File "rdiff_backup\backup.py", line 821, in fast_process_file
  File "rdiff_backup\rpath.py", line 869, in delete
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: b'D:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock'
[1300] Failed to execute script rdiff-backup
```
and:
```
Incrementing mirror file D:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/favicons.sqlite-wal
Exception '[Errno 13] Permission denied: b'C:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock'' raised of class '<class 'PermissionError'>':
  File "rdiff_backup\robust.py", line 94, in check_common_error
  File "rdiff_backup\Rdiff.py", line 40, in get_delta_sigrp_hash
  File "rdiff_backup\rpath.py", line 1100, in open

Processing changed file Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock
Incrementing mirror file D:/Users/TheUser/AppData/Roaming/Mozilla/Firefox/Profiles/xxxxxxxx.default-release/parent.lock
Cleaning up
```

In my environment, this happens consistently, always on the Firefox profile's `parent.lock` file.
When this file doesn't exist, the backup otherwise performs successfully to the end (despite several `ListError`s on other system files).

ericzolf:
FIX: avoid abort on PermissionError if the source file had been successfully backed up once, closes #619

DEV: add testing/action_backuprestore_test.py to Windows tests

I didn't manage to test the fixed issue, opening the file isn't enough to create a proper lock.


desseim:
I see, so if I understand correctly, on executing `librsync.DeltaFile(rp_signature.open("rb"), hash.FileWrapper(rp_new.open("rb")))`, `rp_new.open("rb")` would fail leaving `rp_signature` open as it was expected to be closed [within `DeltaFile`'s constructor](<https://github.com/rdiff-backup/rdiff-backup/blob/b819fdec629c3078e56e2d8c6b547c54a6544a67/src/rdiff_backup/librsync.py#L150>)?

If so, wouldn't it be a bit more future- and fool-proof to close `rp_signature` where it is open? I.e. either

1. open it in `DeltaFile` constructor right before it is needed
2. close it at call site of `DeltaFile` constructor in a `finally`

I think it'd make sense since `DeltaFile` constructor doesn't seem to wrap `rp_signature` the way it does for `rp_new`, but simply use it at construction time.

That being said, I don't know the code base and I understand it might be totally wrong or inappropriate. These were just my thoughts after trying to understand the issue and its fix, and I thought I would at least share them.

src/rdiff_backup/Rdiff.py
```diff
@@ -34,8 +34,12 @@ def get_delta_sigrp_hash(rp_signature, rp_new):
     """Like above but also calculate hash of new as close() value"""
     log.Log("Getting delta (with hash) of file {fi} with signature {si}".format(
         fi=rp_new, si=rp_signature), log.DEBUG)
-    return librsync.DeltaFile(
-        rp_signature.open("rb"), hash.FileWrapper(rp_new.open("rb")))
+    try:
+        return librsync.DeltaFile(
+            rp_signature.open("rb"), hash.FileWrapper(rp_new.open("rb")))
+    except OSError:
```
desseim, line 40: In any case, wouldn't a `finally` be a bit safer (i.e. more robust) here?

ericzolf:
All your comments make perfectly sense, and I don't have a good answer, just that I still have a lot of refactoring to do.

src/rdiff_backup/Rdiff.py
ericzolf: No, because in success case, rp_signature would have been closed by DeltaFile and can't be closed a 2nd time (it can be checked but then it becomes quickly ugly). As said, lot of re factoring while keeping compatibility etc.

desseim: I see, from the name I thought `close_if_necessary()` was idempotent.

I also understand the need for minimal changes to the current structure of the code base.
I first thought of something like this:

```python
    rp_new_fw = hash.FileWrapper(rp_new.open("rb"))
    rp_signature_fp = rp_signature.open("rb")
    return librsync.DeltaFile(rp_signature_fp, rp_new_fw))
```

to induce minimal behavioral change while addressing similar potential bugs where `rp_new.open()` would throw another type of unexpected exception.

But I guess `FileWrapper` wouldn't automatically close `rp_new` if not registered to the new `DeltaFile`, so we'd create the mirror bug where `rp_signature.open()` throws and `rp_new` doesn't get closed ¯\\_(ツ)_/¯

desseim:
I understand your stance, and totally agree on separating refactoring changes. And on the risks in changing the current code structure.

When comes the time though, I've seen that python supports `with` statements, I might ease stream management quite a bit and get rid of similar bugs :bulb:

That's it for my 2 cents.

More importantly though, I think a similar guard is necessary around all calls to `DeltaFile`'s constructor.
I've only identified [this one](<https://github.com/rdiff-backup/rdiff-backup/blob/b819fdec629c3078e56e2d8c6b547c54a6544a67/src/rdiff_backup/Rdiff.py#L49>), but since the execution of `get_signature(basis)` opens `basis`, I'm afraid it is theoritically prone to the same bug.

ericzolf:
`with` works generally well, and I've introduced it to manage the client-server connection but it can mean a lot of refactoring: if you have, which is often the case in this code, a file opened in function A, then closed in function B, then a function C which calls function A then function B, it means that you have to move the open and close activities back into function C, which can be difficult if you have a lot of logic around them in function A resp. B. It becomes even more complicated if you have the open function not in A, but in a sub-function and/or a class' __init__ (which was the case here BTW). And because objects are sometimes passed between client and server, you need to pay attention that this doesn't break their interface or you break the backward compatibility. Not saying this wouldn't be the right thing to do, but it might not be as easy as it sounds, and hence not for this version of rdiff-backup.

desseim:
Totally agree and understand :+1: 

What about the similar instance in `write_delta()` otherwise?
Am I wrong about that or is it simply not worth touching before it breaks?

ericzolf:
> What about the similar instance in `write_delta()` otherwise?
> Am I wrong about that or is it simply not worth touching before it breaks?

Just a case of never touch a running system (I'll look into it at once and globally)

desseim:
Gotcha, it ain't broke yet.

Thanks for taking the time :bow:

Issue #602, opened 2021-06-16
[ENH] Provide a console-script entrypoint to automatically generate executable script on all platform


ikus060:
## Is your feature request related to a problem? Please describe.

current version of rdiff-backup provide a mini script to launch rdiff-backup. While this script is "working" on Linux, it's not working on Windows platform. As a simple replaceent, I recommand using a setuptools entrypoint to automatically generate an executable when rdiff-backup is installed using `pip install rdiff-backup`

## Describe the solution you'd like

In setup.py we need to add 
```
entry_points = { "console_scripts": ["rdiff-backup = rdiff_backup.Main:error_check_Main" ] }
```
We also need to adapt a bit the main function call to make the arglist optional.

The same is probably true for `rdiff-backup-statistics`

ericzolf:
Did you try it? I mean, the rdiff-backup script runs under Windows, Windows is just too stupid to recognize it as Python script,so you explicitly need to call Python. I mean, we can try, I'm just not convinced, it will solve anything.

From the docs, we might have to switch to a setup.cfg.

ikus060:
Yes, I have tried, When installing with pip it's not possible to call `rdiff-backup` from command line. 
From a user perspective, it's not working. Since it's relatively simple to fix with an entrypoint, I would add it.

ericzolf:
My question was more ifyou've tried it with an entry point and if it really works under Windows. I'm just wondering which kind of magic could make it work...

ikus060:
Yes, it's working with an entry_point. When an entry_point is created, setuptools generate a "rdiff-backup.exe" along with a "rdiff-backup-script.py" that get launch when calling rdiff-backup.

I'm using it for Minarca as an example. I will submit a PR if you are ok with this.

ericzolf:
Please do, that sounds really nice. I'm a bit concerned by the special imports in the current script, but perhaps Python does there also the needed magic.

ericzolf:
It sounds like this could be the solution to #733 (the "scripts [must] be called .py [under Windows]" part). Did you do any progress on this resp. could this be something you do for 2.2.0?

ericzolf:
This has been done, I'm surprised the issue hasn't been closed automatically...

PR #754, opened 2022-10-02
Fix windows wheel

### LINKED ISSUE #733: [BUG] the Windows wheel(s) should contain librsync.dll and scripts be called .py
## Bug summary

The wheel doesn't work under Windows without the library, and Windows doesn't recognize `rdiff-backup` as python script, it should be called `rdiff-backup.py`, I never noticed because I don't use the wheel (and apparently nobody does either).

ericzolf:
FIX: make Windows wheel runnable by adding rsync.dll and renaming scripts, closes #733, #602

Use entry_points instead of scripts in setup.py. @ikus060 you were right in #602, it rocks (but makes things a bit more complicated with PyInstaller).

The scripts have disappeared from the source code and moved to modules.
The modules run_delete and run_stats are in rdiff_backup because ultimately they should become actions and disappear.
The setup.py has become yet a little more ugly to copy the DLL to the right place, but it'll need anyway a bigger revamp at some point in time. Most important, it works.


src/rdiffbackup/run.py
```diff
@@ -23,6 +23,18 @@
 from rdiff_backup import Globals, log
 from rdiffbackup import arguments, actions_mgr, actions
 
+try:
+    import msvcrt
+
+    msvcrt.setmode(0, os.O_BINARY)
```
ikus060, line 29: A comment would be helpful to understand why it's needed

setup.py
```diff
@@ -240,10 +257,19 @@ def run(self):
...
+            'rdiff-backup-delete = rdiff_backup.run_delete:main',
+            'rdiff-backup-statistics = rdiff_backup.run_stats:main',
+        ]
+    },
     packages=["rdiff_backup", "rdiffbackup",
               "rdiffbackup.actions", "rdiffbackup.utils", "rdiffbackup.meta",
               "rdiffbackup.locations", "rdiffbackup.locations.map"],
     package_dir={"": "src"},  # tell distutils packages are under src
+    include_package_data=True,
+    package_data={"rdiff_backup": ["*.dll"]},
```
ikus060, line 272: Would that include the dll file for Linux too !?

ericzolf: No, it just wouldn't exist. And that would simply be ignored.

src/rdiffbackup/run.py
ericzolf: To be honest, I took over the code as it stands, and never cared as it worked. But, good point, and clarified.

PR #758, opened 2022-10-13
Add completion for new CLI

### LINKED ISSUE #725: [BUG] bash completion doesn't support new CLI of rdiff-backup
As the subject says...

ericzolf:
FIX: bash completion works for new _and_ old CLI, closes #725

NEW: the action plug-in `complete` makes writing completion scripts beyond bash much easier, afficionados of alternative shells are called to write their own, see docs/arch/completion.adoc for details

Also added a new test script for the new complete action.


ikus060:
I did not test the bash completion. But my understanding is it get auto created.
+1

ericzolf:
Mostly, yes.  Basically rdiff-backup complete provides a list of possible completions, and one needs only little glue code in bash itself.

PR #763, opened 2022-10-22
Preparing release v2.1.3b3

ericzolf:
Fix warnings in pipeline due to saulmaldonado/skip-workflow and fix typo
Possibly not the right timing to do all these changes...


ericzolf (commented):
Automatically created and merged by make_release.sh

PR #775, opened 2022-11-04
Improve overall test coverage

ericzolf:
* test internal function of simpleps to test both versions
* exclude locking lines for neither NT nor posix platforms
* add DOS files quoting test to locations.map.filenames
* add action_calculate_test test case, testing also rdiff-backup-statistics
* remove MacOS specific parts from coverage in stdattr.py
* remove unused functions from user_group
* add failure test cases for restore and backup
* simplify log internal function _exception_to_string
* add test_debug_output function to api_test
* test --exclude/include-if-present in selectiontest

NEW: make --include-if-present work, it was documented but not implemented

This raised the test coverage from 81 to 85%, the rest are either Windows or MacOS specific code (even though I've tried to remove most MacOS parts from coverage), and very specific and hard to simulate system errors (e.g. obscure file system errors).


ikus060:
Good Job !

ericzolf:
Thanks, still an issue under Windows but that seems to be a triffle.

ericzolf:
It wasn't as easy as it seemed: apparently rdiff-backup-statistics has never worked under Windows, now it does.

PR #784, opened 2022-11-14
NEW: Optionally show size of removed increments

### LINKED ISSUE #740: [ENH] when removing increments, optionally show the size of each removed increment

Tools like rdiff-backup make it very easy to create and manage backups.  The biggest problem I have is managing the storage for the backups.  I have a script that
1. displays the storage capacity and usage of the backup destination volume before the backup
2. removes increments before some time period (e.g. 1 year)
3. does the backup showing backup statistics (i.e. --print-statistics)
4. displays the storage capacity and usage of the backup destination volume after the backup
5. verifies the backup
6. lists the changes in the last backup (--list-changed-since 1B)
7. lists all the increments and their sizes

This provides me some information about how/why my backup storage usage changed each day, but there are two things that would be helpful.
1. when removing increments, optionally show the size of each removed increment
2. when listing changes since ..., optionally show the size of each new/deleted file and the change in size of each changed file

For me, item 2 would be nice-to-have, but item 1 is important.

In rdiff-backup 2.2 (syntax) added a --size option to the "remove increments" action that would display the sizes of each removed increment.

pcanning:
Adds a --size option to the remove increments command that causes it to add these increment size to the lines showing the removed increments.

Closes #740

## Changes done and why

> **NOTE:** if your explanation needs to be (very) different from your Git commit message, your Git commit might be insufficient, please re-think it and amend it! Your message must contain `Closes #NNN` if it [closes an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

## Checklist

- [x] commit contains a description of changes relevant to users prefixed by DOC, FIX, NEW and/or CHG (see top of the changelog for details)

### DIFF
--- a/docs/rdiff-backup.1.adoc
+++ b/docs/rdiff-backup.1.adoc
@@ -208,8 +208,8 @@ all the increments older than the given time will be deleted.
 See <<_time_formats,TIME FORMATS>> for details.
 
 --size;;
-Show the size of each increment being removed. The default
-is to _not_ show sizes.
+Show the size of each increment being removed.
+The default is to _not_ show sizes.
 
 restore [<<_creation_options,CREATION OPTIONS>>] [<<_compression_options,COMPRESSION OPTIONS>>] [<<_selection_options,SELECTION OPTIONS>>] [<<_filesystem_options,FILESYSTEM OPTIONS>>] [<<_user_group_options,USER GROUP OPTIONS>>] [*--at* _time_|*--increment*] _source_ _targetdir_::
 restore a source backup repository at a specific time or a specific     source increment to a target directory.
--- a/docs/rdiff-backup.1.adoc
+++ b/docs/rdiff-backup.1.adoc
@@ -195,7 +195,7 @@ If an rdiff-backup session fails, this action will undo the failed directory.
 This happens automatically if you attempt to back-up to a directory and the last backup failed.
 You can use the *--force* option to undo the last backup even if it wasn't failed.
 
-remove *increments* *--older-than* _time_ _repository_:: Remove the incremental backup information in the destination directory that has been around longer than the given time, or the oldest one if no time is provided.
+remove *increments* *--older-than* _time_ [*--size*] _repository_:: Remove the incremental backup information in the destination directory that has been around longer than the given time, or the oldest one if no time is provided.
 +
 By default, rdiff-backup will only delete information from one session at a time.
 To remove two or more sessions at the same time, supply the *--force* option (rdiff-backup will tell you if it is required).
@@ -206,6 +206,10 @@ Thus if you deleted a file two weeks ago, backed up immediately afterwards, and 
 --older-than _time_;;
 all the increments older than the given time will be deleted.
 See <<_time_formats,TIME FORMATS>> for details.
+
+--size;;
+Show the size of each increment being removed. The default
+is to _not_ show sizes.
 
 restore [<<_creation_options,CREATION OPTIONS>>] [<<_compression_options,COMPRESSION OPTIONS>>] [<<_selection_options,SELECTION OPTIONS>>] [<<_filesystem_options,FILESYSTEM OPTIONS>>] [<<_user_group_options,USER GROUP OPTIONS>>] [*--at* _time_|*--increment*] _source_ _targetdir_::
 restore a source backup repository at a specific time or a specific     source increment to a target directory.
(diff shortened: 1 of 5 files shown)


ericzolf:
At first glance, it looks OK, but I need to look deeper into it. I'm definitely missing tests and documentation/manpage update. You have time, I won't probably merge before version 2.4, too late in the development cycle.
Summary: good job :smile:

ericzolf:
Oh, also consider the 3rd check in the list. It's a NEW: feature.

ericzolf:
Regarding the commit check, I think you misunderstood it :  the important part of your commit message must start with NEW: so that it comes (half) automatically into the changelog, and it must end with the closes statement on the same line.

ericzolf:
I didn't check the details but the test fails on the removal tests, so seems to be related to your changes (sometimes, it's just a hick-up of the pipeline). Are you able to see the details of the pipeline and solve the issue yourself @pcanning ?

pcanning:
I found one problem.  While `self.actions.size` is always set when using the new command line interface, its not set when using the old command line interface.  Do you have a recommendation (or an example) of the best way to set it to `False` (or `None`) when using the old command line interface, or should I just use `getattr(self,actions, "size", False) `instead if `self.actions.size`?

ericzolf:
Don't care, the old interface will disappear with 2.4 :shrug: see #793 
Should you nevertheless decide to care, the function `_make_values_like_new_compat200` in `src/rdiffbackup/arguments.py` is the one making sure that the old CLI gives results like the old one.

ericzolf:
As an afterthought, if you do care and support the old CLI, I could add your PR to a version v2.2.1, then you wouldn't need to wait years for your PR to be released with v2.4.0, which is probably an attractive offer. Your change is after all rather limited, so I could do that if you want (and are ready).

BTW, you'll have also to rebase your PR before it can be merged: 
image

ericzolf:
Two more things:
1. I won't review seriously before all the points in the checklist have been ticked off by you (in case you expected me to do it as part of the review, it's the other way around, I probably should clarify this)
2. there was a hick-up with the Windows tests, I restarted them and you're green now, so far so good.

pcanning:
The reason I haven't checked off 

> changes to the code have been covered by new/modified tests

is that I haven't added or modified any tests. My change doesn't affect the behavior of rdiff-backup, it only affects the output produced by the remove  increments command. I looked at testing/action_remove_test.py and AFAICT it doesn't test anything about the output produced by the command.  I also looked at testing/action_list_test.py and didn't see any tests for the --size option (which is what I based my change on).  I could easily replicate test_action_removeincsolderthan in testing/action_remove_test.py and add "--size" to all the invocations of remove  increments.  @ericzolf do you think that is worthwhile?

ericzolf:
Few points here:

1. it's not only about the impact the code has but also about making sure that all paths through the code are covered, else the coverage score goes down (and believe me, it's a pain to get it up again). As you introduced new paths, you need to make sure they're being gone through.
2. I indeed generally don't validate the output because I don't want to rely on error/warning messages which might change and IMHO are not relevant to the interface (at some point, they might even get localized)
3. you missed one test in testing/action_list_test.py which uses the `--size` parameter _and_ checks the output, but only in parsable (computer readable) mode, for exactly the reasons pointed at in the previous point
4. I don't always check everything with everything but just make sure that I use each option at least once (e.g. I switch happily between remote and local calls, but don't test everything first source & target local, then both remote, then each one only remote; that would take too much time)

Many words to just say one thing: you can do as you wrote, but it's also sufficient to add the `--size` parameter to one or two of the remove calls (e.g. one local and one remote), as long as there are enough instance of the call still without `--size`. It would have been different if there would have been a "parsable" output.

ericzolf (changes requested):
Only one little formatting change required, and of course the test is still missing.

Good job!

pcanning:
Are the tests in testing/action_remove_test.py local or remote, and where are the other tests (remote or local)?

ericzolf:
> Are the tests in testing/action_remove_test.py local or remote, and where are the other tests (remote or local)?

It's defined by the first two parameters of the comtst.rdiff_backup_action function (or rdiff_backup function, both in `testing/commontest.py`), True/1 standing for local execution.

ericzolf (approved):

ericzolf:
Now that 2.2.0 is out,  I can merge.  Thanks for the good work and your patience!

pcanning:
Thanks Eric.  Just curious, how soon do expect there will be some sort of (e.g. alpha) release that contains this feature?  I can certain build it myself for my Linux machines, but I also use rdiff-backup on my Windows 10/11 machines and I don't know how to build the Windows versions.

PR #794, opened 2022-11-25
Fix recognition of remote Windows OS

### LINKED ISSUE #788: [BUG] using conn.os.name has never worked
## Bug summary

The value returned by `conn.os.name` is always `os` (insted of `nt` & co), because it is the name of the `EmulateCallable` created by `conn.os`. There are actually two underlying issues:

1. `os.name` isn't a callable but a variable hence it returns in all cases the `EmulateCallable` object instead of the return value of the callable
2. Further more, `EmulateCallable` has an attribute `name` which would also overlap a method called `name()`

n/a - always been there

## rdiff-backup call

```
$ grep -r conn.os.name src
src/rdiff_backup/rpath.py:            if self.conn.os.name != 'nt' or not self.isdir():
src/rdiffbackup/locations/directory.py:        if (self.base_dir.conn.os.name == 'nt'
```

I noticed because it makes the remote call to compare fail if a symlink is involved in the comparison test.
I expect that the compare tests work under Windows (without need to skip, which is the workaround in my current branch).

## More information

I still need to look at the details but I think I'd like a generic `get_sys_info` method on all locations which could provide more information about remote locations than just the os name. I still need to check how to make it reachable in rpath.py.

It's a bigger change so I don't want to pack it in my current Windows tests branch.

ericzolf:
## Changes done and why

FIX: failed to properly recognize remote OS as being Windows, closes #788
    
Also reduced a bit the impact of the previous change, because a normal exception might be catched at the file level, whereas FatalError would always exit

Modify comparetest.py to use new feature. Allow remote parts of comparetest to work under Windows

## Checklist

- [x] commit contains a description of changes relevant to users prefixed by DOC, FIX, NEW and/or CHG (see top of the changelog for details)

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -766,7 +766,11 @@ class RPath(RORPath):
         except OSError:
             # It's not possible to set a modification time for
             # directories on Windows.
-            if self.conn.os.name != 'nt' or not self.isdir():
+            if Globals.get_api_version() < 201:  # compat200
+                is_windows = self.conn.os.name == "nt"
+            else:
+                is_windows = self.conn.platform.system() == "Windows"
+            if not (is_windows and self.isdir()):
                 raise
         else:
             self.data['mtime'] = modtime
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -766,12 +766,14 @@ class RPath(RORPath):
         except OSError:
             # It's not possible to set a modification time for
             # directories on Windows.
-            if Globals.get_api_version() < 201:  # compat200
-                is_windows = self.conn.os.name == "nt"
-            else:
-                is_windows = self.conn.platform.system() == "Windows"
-            if not (is_windows and self.isdir()):
+            if not self.isdir():
                 raise
+            if Globals.get_api_version() < 201:  # compat200
+                if self.conn.os.name != "nt":  # doesn't work, just historical
+                    raise
+            else:
+                if self.conn.platform.system() != "Windows":
+                    raise
         else:
             self.data['mtime'] = modtime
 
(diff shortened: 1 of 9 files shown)


tox.ini
```diff
@@ -92,8 +92,10 @@ commands =
 
 [flake8]
 ignore =
-	E501 # line too long (86 > 79 characters)
-	W503 # line break before binary operator
+	# line too long (86 > 79 characters)
```
ikus060, line 95: Ha ha, I had to do the same thing in Rdiffweb, Did you figure out why ? I guess it's related to a recent version of flake8..

src/rdiff_backup/rpath.py
```diff
@@ -766,7 +766,11 @@ def setmtime(self, modtime):
         except OSError:
             # It's not possible to set a modification time for
             # directories on Windows.
-            if self.conn.os.name != 'nt' or not self.isdir():
+            if Globals.get_api_version() < 201:  # compat200
```
ikus060, line 769: As another tough. Should we not have a fs_abilities to check if say directory support modification date.

So we have a test at the beginning only and use the result for the rest of the backup. Current implementation will make the same call to `self.conn.platform.system()` for every directory. Right ?

tox.ini
ericzolf: Yes, definitely: https://github.com/PyCQA/flake8/issues/1760

src/rdiff_backup/rpath.py
ericzolf: The idea is good. Actually, I should at least check first if it's a directory, and then only check which OS it is, that would already reduce the number of calls across the line.
Your suggestion requires more changes, let me think about it.

ericzolf: So, I've implemented the more efficient approach, checking first for a directory and _then_ only for Windows.

Additionally, I've created two enhancement requests #795 and #796 - but it's something for 2.4 not for now.

PR #823, opened 2023-01-12
allow user to use curly braces for shell purposes in remote-schema

### LINKED ISSUE #818: [BUG] remote-schema doesn't allow {} use in shell
## Bug summary

My remote-schema has had { } curly-braces in it forever.  2.2.0 won't run with my remote-schema it seems.  Gives schema error.  tcsh which I use (not sure about bash) allows wrapping multiple commands in {} kind of like a subshell, but different from ().

I fixed the bug by changing in SetConnections.py in def _fill_schema the first findall to:
        if ((re.findall(b"(?<!{){[^{}]*}", cmd_schema)

And doubling-up my shell {} into {{}} in my remote-schema.  This "just works" because python string format allows escaping the {} by doubling up as such.  The only thing required to make it work is the regex change above to exclude checking the double-upped {}'s.

---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - /bin/rdiff-backup
  - info
  parsed: null
  version: 2.2.0
python:
  executable: /usr/bin/python3
  name: cpython
  version: 3.10.9
system:
  fs_encoding: utf-8
  platform: Linux-6.0.15-200.fc36.x86_64-x86_64-with-glibc2.35
...

## rdiff-backup call

The remote-schema with the {} is (after I doubled them up and now working):

--remote-schema 'ssh -x -C -q -o StrictHostKeyChecking=no -o CheckHostIP=no -c aes256-gcm@openssh.com {h} '\''(/usr/local/sbin/netbak-smbmount '\''//192.168.100.2/Users'\'' '\''/mnt/netbak'\'' '\''netbak'\'' '\''0'\'' && ( {{ ok=1;  (( `find /mnt/netbak/Joe/Favorites | head -2 | wc -l` == 2 )) && ( umask 077; chmod --quiet 600 /tmp/rdb.log; rnice=/usr/bin/nice; [ -e /bin/nice ] && rnice=/bin/nice; $rnice -n19 /usr/bin/ionice -c2 -n7 rdiff-backup  server 2>>/tmp/rdb.log ) && ok=0; }} ; /usr/local/sbin/netbak-smbumount /mnt/netbak; rm -f /home/netbak/.netbak-smbmount.cred; exit $ok )) | /usr/bin/pv -L 200000 -rbWf -i2592000'\'

It immediately fails when checking the command line options with a bad remote-schema error.  Should work ok with my old remote-schema.

tecordes:
## Changes done and why

FIX allow user to use curly braces for shell purposes in remote-schema

Change the are-there-any-unrecognized-curly-braces test to exclude doubled-up curly braces so that the user can include in their remote-schema curly-braces that are meant to be passed as-is to the shell.

format() will then change the doubled-up curly braces into a single one, and that is what will be passed to the remote shell.

bash and tcsh allow curly braces in command lines for various purposes.  My main use is in bash to group commands for && conditional execution without starting a subshell, because I need to obtain the results of a variable within the group outside the group.  A subshell would not allow the return of variables, only an exit code.

The easiest test case that illustrates my goal is below.  The test case might be achieved without the {{}}, but it somewhat reflects what I need to do in any case.

rdiff-backup --remote-schema 'ssh -x -C {h} "( {{ ret=1; ls /tmp/dobackup >/dev/null 2>&1 && rdiff-backup server && ret=0; }} ; exit \$ret ) | pv -L 50000"' backup me@foo.baz::/tmp/Z /tmp/RD

Fixes #818

## Self-Checklist

- [X] changes to the code have been reflected in the documentation
- [X] commit contains a description of changes

### DIFF
--- a/docs/rdiff-backup.1.adoc
+++ b/docs/rdiff-backup.1.adoc
@@ -486,7 +486,9 @@ For example, if you have rdiff-backup 2.1.5 and 2.2.1 installed in virtual envir
 The client will then use the correct version of rdiff-backup based on its own version '[.code]``x.y.z``'.
 You'll find more explanations in the *migration* file in the documentation.
 
-And finally, to include a literal '[.code]``%``' in the string specified by *--remote-schema*, quote it with another '[.code]``%``', as in '[.code]``%%``' (this is due to the compatibility with the deprecated host placeholder '[.code]``%s``', which you shouldn't use anymore).
+If you need to include a literal '[.code]``%``' in the string specified by *--remote-schema*, quote it with another '[.code]``%``', as in '[.code]``%%``' (this is due to the compatibility with the deprecated host placeholder '[.code]``%s``', which you shouldn't use anymore).
+
+And finally, if you need to include literal '[.code]``{ }``' (curly braces) in the the string specified by *--remote-schema*, quote them (both) by doubling each of them up, as in '[.code]``{{ foo=0; }}``'.
 
 Although ssh itself may be secure, using rdiff-backup in the default way presents some security risks.
 For instance if the server is run as root, then an attacker who compromised the client could then use rdiff-backup to overwrite arbitrary server files by "backing up" over them.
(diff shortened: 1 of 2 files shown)


ericzolf:
Good one, thanks for taking care. Never mind the test, I've added #824 to take care a bit more holistically. But volunteers are always welcome :smirk: 
Regarding the prefixed message, I'll clarify the expectations, thanks for the feedback. Check  the output of `git log`, you should see what I mean.

PR #827, opened 2023-01-15
Fix CloseConnections doesn't wait for child to die

### LINKED ISSUE #819: [BUG] CloseConnections doesn't wait for child processes to die, losing output
## Bug summary

rdiff-backup will exit while the ssh process is still running for .5 to 1.5 seconds.  If I check with ps command at the right time I can see that the ssh command has been taken over by ps 1 as its ppid.

This screws up the capture of the output from my pv pipe-monitoring program at the end of my remote-schema chain.  The output goes to nowhere, since the ppid is now 1.  rdiff-backup never sees/passes on this output.

I've fixed it with a few minor mods to SetConnections.py and connection.py.  Just need to track the child and wait for it to exit.

```yaml
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - /bin/rdiff-backup
  - info
  parsed: null
  version: 2.2.0
python:
  executable: /usr/bin/python3
  name: cpython
  version: 3.10.9
system:
  fs_encoding: utf-8
  platform: Linux-6.0.15-200.fc36.x86_64-x86_64-with-glibc2.35
...
```
## rdiff-backup call

remote-schema ends in:

| /usr/bin/pv -L 200000 -rbWf -i2592000

pv final status output disappears when connection is slow and/or you get unlucky with the race condition.  I expect the final pv status output line to show up so my script can capture it from the rdiff process.

## More information

my fix in brief:

connection.py:
```python
class LowLevelPipeConnection(Connection):
...
    def __init__(self, inpipe, outpipe, process = 0):
        self.process = process

    def _close(self):
        self.outpipe.close()
        self.inpipe.close()
        if self.process!=0: self.process.wait()

class PipeConnection(LowLevelPipeConnection):
    def __init__(self, inpipe, outpipe, conn_number = 0, process = 0 ):
...
	LowLevelPipeConnection.__init__(self, inpipe, outpipe, process)
```
SetConnections.py:
```python
def _init_connection(remote_cmd):
    conn = connection.PipeConnection(stdout, stdin, conn_number, process)
```

Note:
The Python docs on popen says wait could deadlock, but since the rdiff-backup protocol is so precise, I doubt there could ever be remaining data on the pipe at that point.   Otherwise they say to use communicate() but that seems more complicated than what I'm after.  (And Python is not my language of choice nor maximum aptitude.)

I've run this fix since 2019 and never had a problem at all.  I'm tired of manually editing the 2 source files everytime rdiff-backup gets upgraded :-)

tecordes:
## Changes done and why

FIX: CloseConnections doesn't wait for child prcesses to die, losing output, Fixes #819

Keep track of the process object within the LowLevelPipeConnection so we can wait() for it properly when closing the connection, thus avoiding possibly lost output from commands run in the remote-schema after the rdiff-backup server process.

Remove the old _process global that tracked the process purely for avoiding the "ResourceWarning: subprocess N is still running". As long as we are pretty sure the process is dead/killed before the connection destructor, the warning should not recur.  The extra terminate/kill and sleeps should not run in the normal case, but if delays occur, should help to ensure the process is marked dead before destruction.

Note: I have started working on a test case but have no way to easily test it yet(?), so I will post the code ideas in the original bug https://github.com/rdiff-backup/rdiff-backup/issues/819 and request feedback before doing another PR for the testing.

## Self-Checklist

- [X] changes to the code have been reflected in the documentation
- [X] commit contains a description of changes

### DIFF
--- a/src/rdiff_backup/connection.py
+++ b/src/rdiff_backup/connection.py
@@ -29,6 +29,8 @@ import shutil  # noqa: F401
 import socket  # noqa: F401
 import tempfile  # noqa: F401
 import types  # noqa: F401
+import time
+import subprocess
 
 # The following EA and ACL modules may be used if available
 try:  # compat200
@@ -157,10 +159,11 @@ class LowLevelPipeConnection(Connection):
 
     """
 
-    def __init__(self, inpipe, outpipe):
+    def __init__(self, inpipe, outpipe, process=None):
         """inpipe is a file-type open for reading, outpipe for writing"""
         self.inpipe = inpipe
         self.outpipe = outpipe
+        self.process = process
 
     def __str__(self):
         """Return string version
@@ -363,6 +366,23 @@ class LowLevelPipeConnection(Connection):
         """Close the pipes associated with the connection"""
         self.outpipe.close()
         self.inpipe.close()
+        # reap the pipe child with wait() to ensure any final output from
+        # the pipe by commands that run after rdiff-backup server; otherwise
+        # a race condition occurs where final output is sometimes lost;
+        # Python>=3.3 gives the timeout option to wait: set to a modestly
+        # small value here to minimize possibility of introducing bugs/delays
+        if self.process:
+            try:
+                self.process.wait(timeout=5)
+            except subprocess.TimeoutExpired:
+                # if some longer delay occurred, just kill it, try hard so
+                # we avoid the ResourceWarning if it's still running when
+                # the subprocess destructors hit; yes ugly sleeps, but rare
+                self.process.terminate()
+                time.sleep(1)
+                if (self.process.poll() is None):
+                    self.process.kill()
+                    time.sleep(1)
 
 
 class PipeConnection(LowLevelPipeConnection):
@@ -377,7 +397,7 @@ class PipeConnection(LowLevelPipeConnection):
 
     """
 
-    def __init__(self, inpipe, outpipe, conn_number=0):
+    def __init__(self, inpipe, outpipe, conn_number=0, process=None):
         """Init PipeConnection
 
         conn_number should be a unique (to the session) integer to
@@ -386,7 +406,7 @@ class PipeConnection(LowLevelPipeConnection):
         number to route commands to the correct process.
 
         """
-        LowLevelPipeConnection.__init__(self, inpipe, outpipe)
+        LowLevelPipeConnection.__init__(self, inpipe, outpipe, process)
         self.conn_number = conn_number
         self.unused_request_numbers = set(range(256))
 
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -35,10 +35,6 @@ from rdiffbackup.utils import safestr
 # The first is None because it is the local connection.
 __conn_remote_cmds = [None]
 
-# keep a list of sub-processes running; we don't use it, it's only to avoid
-# "ResourceWarning: subprocess N is still running" from subprocess library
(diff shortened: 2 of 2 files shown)


ericzolf:
The code looks pretty good to me, let me sleep over it and think if we really need an additional tests, perhaps with a dummy `sleep 10` on the server side to make sure the killing happens.

src/rdiff_backup/connection.py
```diff
@@ -363,6 +366,23 @@ def _close(self):
...
+        # Python>=3.3 gives the timeout option to wait: set to a modestly
+        # small value here to minimize possibility of introducing bugs/delays
+        if self.process:
+            try:
+                self.process.wait(timeout=5)
+            except subprocess.TimeoutExpired:
+                # if some longer delay occurred, just kill it, try hard so
+                # we avoid the ResourceWarning if it's still running when
+                # the subprocess destructors hit; yes ugly sleeps, but rare
+                self.process.terminate()
```
ericzolf, line 381: Add please a warning log message that "the server process {sp} is being terminated".format(sp=self.process.args)

src/rdiff_backup/connection.py
```diff
@@ -363,6 +366,23 @@ def _close(self):
...
+                time.sleep(1)
+                if (self.process.poll() is None):
+                    self.process.kill()
```
ericzolf, line 384: Add please a warning log message that "the server process {sp} is being killed".format(sp=self.process.args)

src/rdiff_backup/connection.py
```diff
@@ -363,6 +366,23 @@ def _close(self):
...
         self.outpipe.close()
         self.inpipe.close()
+        # reap the pipe child with wait() to ensure any final output from
+        # the pipe by commands that run after rdiff-backup server; otherwise
+        # a race condition occurs where final output is sometimes lost;
+        # Python>=3.3 gives the timeout option to wait: set to a modestly
+        # small value here to minimize possibility of introducing bugs/delays
+        if self.process:
+            try:
+                self.process.wait(timeout=5)
```
ericzolf, line 376: I think an INFO level message of the kind `"Server process {sp} has exited with return code {rc}".format(sp=self.process.args, rc=self.process.returncode)` could be helpful in certain circumstances.

src/rdiff_backup/connection.py
```diff
@@ -157,10 +159,11 @@ class LowLevelPipeConnection(Connection):
 
     """
 
-    def __init__(self, inpipe, outpipe):
+    def __init__(self, inpipe, outpipe, process=None):
```
ericzolf, line 162: Checking again `testing/connectiontest.py`, I think that `self.process` should stay at `PipeConnection` level and shouldn't be known by LowLevelPipeConnection. Basically, this is the difference between both that one is only two input and output pipes, whereas the other really has a process in-between.

That means that a _close method of PipeConnection should override the LowLevel one and contain your code. Does it make sense?

ericzolf (changes requested):
Looks very good to me, I just think that a few more messages might help understanding what's happening, and that LowLevelPipeConnection shouldn't know about the underlying process.

Regarding testing, I think that adding `self.p` to the call of PipeConnection on line 112 should be sufficient.

ericzolf:
@tecordes any news and/or ETA?

ericzolf:
@tecordes do you still plan to finish this PR according to my comments or should I finish it myself?

ericzolf:
This PR can be considered obsolete, just check #837

PR #829, opened 2023-01-15
Include latest actual build dependencies in Debian packaging control file

ottok:
DEV: Ensure the native Debian packaging correctly builds by including new build dependencies asciidoctor and PyYAML. Without these the man page generation would fail and program would fail to start with Python erroring about missing module 'yaml'.

## Testing

Before this the Debian/Ubuntu builds failed locally and at https://launchpad.net/~otto/+archive/ubuntu/ppa/+builds?build_text=&build_state=all

When this branch is pushed to Launchpad builds pass.

image

Same changes were also done downstream in
* https://salsa.debian.org/python-team/packages/rdiff-backup/-/commit/2338b1fc6083c0621cc701fe4a28120dd0d4c51c
* https://salsa.debian.org/python-team/packages/rdiff-backup/-/commit/b3789b60094f52fc6b7ecbdf606427421547a47b

## Self-Checklist

- [N/A] changes to the code have been reflected in the documentation
- [x] commit contains a description of changes

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,14 +2,16 @@ Source: rdiff-backup
 Section: utils
 Priority: optional
 Maintainer: Otto Kekäläinen <otto@debian.org>
-Build-Depends: debhelper (>= 11),
+Build-Depends: asciidoctor,
+               debhelper (>= 11),
                dh-python,
                librsync-dev,
                python3-all-dev,
                python3-pylibacl,
                python3-pyxattr,
                python3-setuptools,
-               python3-setuptools-scm
+               python3-setuptools-scm,
+               python3-yaml
 Standards-Version: 4.4.0
 Homepage: http://rdiff-backup.net/
 Vcs-Git: https://github.com/rdiff-backup/rdiff-backup.git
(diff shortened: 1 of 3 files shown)


debian/control
```diff
@@ -2,14 +2,16 @@ Source: rdiff-backup
 Section: utils
 Priority: optional
 Maintainer: Otto Kekäläinen <otto@debian.org>
-Build-Depends: debhelper (>= 11),
+Build-Depends: asciidoctor,
```
ericzolf, line 5: Since 2.2.2 asciidoctor is an optional dependency on request of some other packagers, to avoid pulling too big of a dependency (ruby, etc), hence the man pages are pre-built and part of the sdist package (but they are not in the Git repo, on purpose, so not sure it applies to your case because I think you build from Git).

ericzolf (approved):

debian/control
ottok: Well this dependency is required for a full build, thus having it as a build dependency is correct.

This is not a run-time dependency.

ericzolf: Fine wwith me, you already had my approval, I'll merge once the tests have gone through.

PR #841, opened 2023-02-07
[DOC] Add FAQ for No space left on device

### LINKED ISSUE #838: [ENH] improve handling of disk full situations

If the backup repository disk is full, nothing works anymore and it's not even possible to rollback and/or remove old increments, because rdiff-backup also needs space to do so.

Apparently, using a separate temp directory makes regress/remove possible. This needs to be confirmed, then a better error message should hint at this and an FAQ entry added to explain the exact rescue procedure.

Forecasting the used disk could be an option but is very difficult thing to do, and even databases don't do this, and get full if nobody monitors their disk space, so there must be reasons...

## Additional context

Issue was reported through
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=310442 resp. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=397161

ericzolf:
## Changes done and why
DOC: add FAQ on how to handle 'No space left on device' messages, closes #838

## Self-Checklist

### DIFF
--- a/docs/FAQ.adoc
+++ b/docs/FAQ.adoc
@@ -458,3 +458,61 @@ If the server side is v2.2+, then the warning message will appear.
 Call rdiff-backup with the higher API and the message will disappear.
 
 Calling for example `rdiff-backup --api-version 201 backup loc1 server2::loc2` will make sure that `rdiff-backup server` is being called, getting rid of the warning message.
+
+== How to handle a "No space left on device" message while doing a backup?
+
+If you're doing a backup and get an exception `[Errno 28] No space left on device: ...` (or something similar under Windows), it means exactly what it states, there is not enough space to create your backup.
+You need to first fix the issue, then recover, then make sure that it doesn't happen again.
+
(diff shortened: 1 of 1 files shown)


docs/FAQ.adoc
```diff
@@ -458,3 +458,58 @@ If the server side is v2.2+, then the warning message will appear.
 Call rdiff-backup with the higher API and the message will disappear.
 
 Calling for example `rdiff-backup --api-version 201 backup loc1 server2::loc2` will make sure that `rdiff-backup server` is being called, getting rid of the warning message.
+
+== How to handle a "No space left on device" message while doing a backup?
```
ikus060, line 462: Some note of that. This error could also be raised if the "/tmp" folder is full. That happen to some one on the mailing list. I just don'T recall if this was during a backup or a restore.

ericzolf: I am the opinion that we addressed the issue at some point in time, but it's not a guarantee that it can't happen so I've added a note to this regard. Thanks.

PR #849, opened 2023-02-13
Refresh Python versions

ericzolf:
## Changes done and why

CHG: remove support for Python below 3.8 and add proper Python 3.11 support, also under Windows
CHG: remove support for 32 bits Linux by lack of working manylinux image (you can probably still compile it yourself locally)

## Self-Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,7 +16,7 @@ defaults:
     shell: bash
 
 env:
-  WIN_PYTHON_VERSION: 3.10.7
+  WIN_PYTHON_VERSION: 3.11.2
   WIN_LIBRSYNC_VERSION: v2.3.2
 
 jobs:
@@ -26,14 +26,14 @@ jobs:
     strategy:
       matrix:
         # we could do it also as matrix but it takes longer
-        #python-version: ["3.11","3.10","3.9","3.8","3.7"]
+        #python-version: ["3.11","3.10","3.9","3.8"]
         #--- Build and deploy Linux wheels using manylinux containers ---
         # - build manylinux2010 x64 and i686
         # - build manylinux2014 x64
         # - build manylinux_2_28 x64
         # avoiding manylinux2014_i686 because does not provide librsync-devel
         # manylinux_2_24 is already deprecated
-        many-linux: [2010_x86_64, 2010_i686, 2014_x86_64, _2_28_x86_64]
+        many-linux: [2014_x86_64, _2_28_x86_64]
     steps:
     - uses: actions/checkout@v3
       with:
@@ -47,7 +47,7 @@ jobs:
           if [[ ${{ matrix.many-linux }} != *64 ]]; then PRE_CMD=linux32; fi
           #py=$(echo ${{ matrix.python-version }} | tr -d .)
           plat=manylinux${{ matrix.many-linux }}
-          $PRE_CMD /ws/tools/build_wheels.sh /ws ${plat} /opt/python/cp3{11,10,9,8,7}*/bin
+          $PRE_CMD /ws/tools/build_wheels.sh /ws ${plat} /opt/python/cp3{11,10,9,8}*/bin
     - name: Upload wheel artifacts
       uses: actions/upload-artifact@v3
       with:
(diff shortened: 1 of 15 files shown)


ericzolf:
@ikus060 this is the kind of changes required to make 3.11 work, it is for me too extensive for a patch version, hope this is OK for you.

.github/workflows/deploy.yml
```diff
@@ -26,14 +26,14 @@ jobs:
...
-        #python-version: ["3.11","3.10","3.9","3.8","3.7"]
+        #python-version: ["3.11","3.10","3.9","3.8"]
         #--- Build and deploy Linux wheels using manylinux containers ---
         # - build manylinux2010 x64 and i686
         # - build manylinux2014 x64
         # - build manylinux_2_28 x64
         # avoiding manylinux2014_i686 because does not provide librsync-devel
         # manylinux_2_24 is already deprecated
-        many-linux: [2010_x86_64, 2010_i686, 2014_x86_64, _2_28_x86_64]
+        many-linux: [2014_x86_64, _2_28_x86_64]
```
ikus060, line 36: I'm not fluent with ManyLinux. Correct me if I'm wrong. I tought the new convention is to use _2_28_x86_64 where _2_28 is compatible with glibc >=2.28. Is it required to keep 2014_x86_64 ?

.github/workflows/deploy.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           if [[ ${{ matrix.many-linux }} != *64 ]]; then PRE_CMD=linux32; fi
           #py=$(echo ${{ matrix.python-version }} | tr -d .)
           plat=manylinux${{ matrix.many-linux }}
-          $PRE_CMD /ws/tools/build_wheels.sh /ws ${plat} /opt/python/cp3{11,10,9,8,7}*/bin
+          $PRE_CMD /ws/tools/build_wheels.sh /ws ${plat} /opt/python/cp3{11,10,9,8}*/bin
```
ikus060, line 50: We should probably keep compatibility with python 3.7.

Debian Buster is supported until June 30th, 2024 and is release with python 3.7
If it's not too hard. https://wiki.debian.org/LTS

.github/workflows/test_windows.yml
```diff
@@ -17,7 +17,7 @@ defaults:
     shell: bash
 
 env:
-  WIN_PYTHON_VERSION: 3.10.7
+  WIN_PYTHON_VERSION: 3.11.2
```
ikus060, line 20: I guess there is not such thing as manywindow !? It's a shame that only one binary version is distributed for Windows.

requirements.txt
```diff
@@ -28,7 +27,7 @@ pywin32 ; platform_system=="Windows"
 twine
 wheel
 #py2exe; platform_system=="Windows"
-pyinstaller==5.4; platform_system=="Windows"
+pyinstaller; platform_system=="Windows"
```
ikus060, line 30: I strongly suggest to fix the version here. I had a couple of problem from time to time with pyinstaller. They keep changing the bootloader and the behavior is not the same. e.g.: the latest version was not resolving call to symbolic link. So creating a symlink of rdiff-backup in /usr/bin was not working.

.github/workflows/test_windows.yml
ericzolf: Well, the difference between Windows and Linux is that Python comes with Linux, so you're dependent on the version delivered (more or less). On Windows, you anyway need to install Python and can decide which version (and we deliver anyway a compiled binary).

requirements.txt
ericzolf: Yeah, on the other hand, it led to having a shit old version of it, being a few versions behind, so not optimal.

.github/workflows/deploy.yml
ericzolf: No, sorry, the rule is to have 4 supported versions of Python, that takes enough of my time. Debian people will have to stay on 2.2, if they are clever enough to install their own version of rdiff-backup, they will probably be clever enough to install their own version of Python, else they go with what's delivered with their platform.

.github/workflows/deploy.yml
ericzolf: My understanding is, yes, for slightly older platforms: https://github.com/pypa/manylinux (e.g. CentOS 8 with Python3.8).

requirements.txt
ericzolf: After thinking about it, I prefer to cope with small incompatibilities on a regular basis, than a bunch of incompatibilities all at once, not knowing which version of PyInstaller introduced it, so I stick to my decision to remove the fixed version.

ericzolf:
All these comments answered, I'll have to throw away this PR anyway, so that I can fix #850 which is a bit too blunt to my taste to keep it in a stable version *sigh*

Issue #867, opened 2023-04-12
[BUG] rdiff-backup.1 has incorrect remote-schema example


frankcrawford:
## Bug summary

The example for using the remote-schema in rdiff-backup.1 has `--remote-schema` listed as a sub-option, but it is a global-option.

## Version, Python, Operating System

```yaml
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - /usr/bin/rdiff-backup
  - info
  parsed: null
  version: 2.2.4
python:
  executable: /usr/bin/python3
  name: cpython
  version: 3.11.2
system:
  fs_encoding: utf-8
  platform: Linux-6.2.9-200.fc37.x86_64-x86_64-with-glibc2.36
...
```

## rdiff-backup call

How did you call rdiff-backup, with which parameters:

```
man rdiff-backup
```

## What happened and what did you expect?

Example listed is
```
rdiff-backup backup --remote-schema 'cd /usr; {h}' \
                               foo 'rdiff-backup server'::bar
```
but I believe should be
```
rdiff-backup --remote-schema 'cd /usr; {h}' backup \
                               foo 'rdiff-backup server'::bar
```
## More information

Other examples do look to be correct.

ericzolf:
You're absolutely correct, do you provide a PR, or shall I do it?

frankcrawford:
I'll submit one shortly.

frankcrawford:
Sorry, been busy, will submit it in the next day or so.

PR #871, opened 2023-04-30
[DOC] Minor doc to fix to example argument order

frankcrawford:
## Changes done and why

Correct example to have the arguments in the example in the right order, as given in issue #867.

## Self-Checklist

NA


ericzolf:
I missed the `DOC: bla bla, closes #NNN` in your commit message, but else fine with me.

Issue #875, opened 2023-05-08
How to set SSH key identity


notklaatu:
I do remote backups of my system. To enable passwordless login, I use an SSH key dedicated to my backup server. This is how I used to do it with rdiff-backup before the new syntax:

```bash
$ rdiff-backup \
--remote-schema "ssh -C -i ~/.ssh/backup_rsa %s rdiff-backup --server" \
/home/klaatu/ klaatu@example.local::/blah
```

This no longer works.

What's the new way to set which SSH key to use when logging rdiff-backup into the remote machine?

notklaatu:
I guess it must have been some environment issue after upgrading rdiff-backup with pip. Now that my workstation has been rebooted, it's working exactly as expected and I can no longer reproduce the error.

Thank you for the support, and the suggestion. I'll look into ssh-add as a way of doing this, should I need it in the future.

notklaatu:
No, I'm wrong. I just forgot overnight how to replicate the issue.

Here's the error:

```bash
$ rdiff-backup -v2 backup --print-statistics --exclude-globbing-filelist /home/klaatu/.excludes \
--remote-schema "ssh -C -i ~/.ssh/backup_rsa %s rdiff-backup --server" \
/home/klaatu/ klaatu@10.1.2.3::/blah/
usage: rdiff-backup [-h] [--api-version API_VERSION] [--current-time CURRENT_TIME] [--force]
                    [--fsync | --no-fsync] [--null-separator] [--new | --no-new] [--chars-to-quote CHARS]
                    [--parsable-output] [--remote-schema REMOTE_SCHEMA] [--remote-tempdir DIR_PATH]
                    [--ssh-compression | --no-ssh-compression] [--tempdir DIR_PATH]
                    [--terminal-verbosity {0,1,2,3,4,5,6,7,8,9}] [--use-compatible-timestamps]
                    [-v {0,1,2,3,4,5,6,7,8,9}] [-V]
                    {backup,calculate,compare,complete,info,list,regress,remove,restore,server,test,verify}
                    ...
rdiff-backup: error: unrecognized arguments: --remote-schema klaatu@10.1.2.3::/blah/
```

The old syntax IS working, but I'm getting warnings that it's deprecated and that I should use the new syntax.

Is using `ssh-add` the only new way to authenticate with a unique SSH key? If so, does anyone have a sample command to demonstrate how that works? I've never used `ssh-add` before.

ericzolf:
[...]

PR #877, opened 2023-05-09
[DOC] Clarify what kind of option is a generic option that must come before Action

notklaatu:
## Changes done and why

This adds a clearer heading to the "Options" section so that a user knows that everything in that section is classified as "generic" and must come before the Action.

I think there's still room to better define what a "Sub-action" is, but that's not within the scope of this pull request.

This helps solve Issue #875


ericzolf:
Fair enough, thanks for taking the effort :smile:

PR #888, opened 2023-06-15
Fix return code when arguments can't be parsed

ericzolf:
## Changes done and why

FIX: failed parsing of arguments would return code 2 for warnings instead of 1 for errors

Workaround is necessary because argparse doesn't allow for overwriting the return code.

## Self-Checklist


ikus060 (approved):
Make sens

ericzolf:
I added a test case which makes the PR ready for final review. Thanks!

ericzolf:
Thanks for the approval!

PR #913, opened 2023-08-19
Add environment variable to change api version

ericzolf:
## Changes done and why
NEW: add environment variable RDIFF_BACKUP_API_VERSION to artificially modify the API version dictionary, see the man page for details.

As I make sure that everything works, this allows me to more quickly try different API versions, and perhaps later test them all in the pipeline.

## Self-Checklist


ericzolf:
OK, ready for review!

src/rdiff_backup/Globals.py
```diff
@@ -54,6 +55,9 @@
     "max": 201,
     "actual": 0
 }
+# Allow overwrite from the environment variable RDIFF_BACKUP_API_VERSION
```
ikus060, line 58: Could we consider a try catch block in case the variable contains garbage ?
```
try:
   ...
catch ScannerError:
   print("a nice user friendly message instead of a stack trace")
   exit(?)
```

ericzolf: As it's an advanced feature, I wouldn't mind a traceback, especially as I would anyway raise the exception. But well, why not?

ericzolf: OK, I checked why not: because the loading happens too early for handling log messages, and I'd want anyway to avoid  adding a circular dependency between Globals and log. So I only made it clearer that's a developer option.

PR #914, opened 2023-08-20
Fix issues with API 201

ericzolf:
FIX: quoting variables where not correctly propagated when set from the command line.
FIX: extended variables were not correctly written if content of file hadn't changed and file was read-only.
FIX: some new functions were needed with higher security level and had to be sorted accordingly.

There were also a few testing issues that needed fixing:

- fix timetest to use function instead of API dependent variables
- when using the RepoShadow class multiple times, as object, it needs to be initiliazed
- fix sequence in restore skipping owners mapping initiation
- add stringify function __str__ to ExtendedAttributes object
- skip restore and cmdline tests with too many internals for new API
- fix test return codes for rdiff-backup-delete

## Self-Checklist


ericzolf:
Ready for review. And once this PR is merged, together with the other ones, I'll release the last 2.2 version aka 2.2.6.

src/rdiff_backup/rpath.py
```diff
@@ -1677,6 +1677,8 @@ def copy_attribs(rpin, rpout):
     if Globals.change_ownership:
         rpout.chown(*map_owners.map_rpath_owner(rpin))
     if Globals.eas_write:
+        if not rpin.issym():  # make sure EAs can be written
```
ikus060, line 1680: I don'T understand this change

ikus060: Is it related to an issue ?

ikus060: Ok, so it was not reported as a bug.

Thanks for the update !

ericzolf: No, I changed the code to force API 201 and uncovered afew bugs while running all tests.

ericzolf:
I guess, that means that I can merge?

PR #946, opened 2023-11-15
Fix sdist building without name and version

ericzolf:
## Changes done and why

Older versions of setuptools and setuptools-scm don't know how to handle properly pyproject.toml, hence enforcing newer versions in the pipeline, and adding accordingly constraints.

Also librsync package in Debian/Ubuntu is now called librsync2.

## Self-Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -59,7 +59,9 @@ jobs:
         sudo apt-get --assume-yes update
         sudo apt-get --assume-yes install asciidoctor
     - name: create source dist package
-      run: python setup.py sdist
+      run: |
+        pip install --upgrade setuptools setuptools-scm  # workaround for old versions
+        python setup.py sdist
     - name: Upload sdist artifact
       uses: actions/upload-artifact@v3
       with:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm"]
+requires = ["setuptools>=67.0.0", "setuptools-scm>=7.1.0"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -54,7 +54,9 @@ jobs:
         sudo apt-get --assume-yes update
         sudo apt-get --assume-yes install asciidoctor
     - name: create source dist package
-      run: python setup.py sdist
+      run: |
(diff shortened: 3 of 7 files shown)


ericzolf:
OK, works as expected and can be reviewed.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm"]
+requires = ["setuptools>=67.0.0", "setuptools-scm>=7.1.0"]
```
ikus060, line 2: Any specific reason to use v67. pyproject.toml is supported by setuptools starting with v61.

.github/workflows/deploy.yml
```diff
@@ -59,7 +59,9 @@ jobs:
         sudo apt-get --assume-yes update
         sudo apt-get --assume-yes install asciidoctor
     - name: create source dist package
-      run: python setup.py sdist
+      run: |
+        pip install --upgrade setuptools setuptools-scm  # workaround for old versions
```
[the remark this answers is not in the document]
ericzolf: So I thought also but latest means 22.04 for github and it comes with setup tools v53 or so.

pyproject.toml
ericzolf: Not entirely specific but I went through https://setuptools.pypa.io/en/latest/history.html and version v66.1.0

.github/workflows/deploy.yml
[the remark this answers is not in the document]
ikus060: Wow Ubuntu is shipping an ancient setuptools version. Thanks for the clarification.

pyproject.toml
ikus060: Ok it make sens then.

frankcrawford:
I will just add a comment here that Fedora 39, i.e. released last week, is the only one that support setuptools>=67.0.0.  Fedora 38 supports 65.5.1 and I don't know what earlier ones handle, but will also go away very soon.

RHEL9 looks to be 53.0.0 and others would be older.

Having said that, it does seem to build on F38 fine, as rpmbuild seems to ignore that version value, although I don't know what will happen with the RHEL versions.

ericzolf:
Noted but as the current changes are not planned to be released for a long time, and I'm currently really trying to be future prone, I'm not too concerned for now. Once we're nearer to release date, we can discuss workarounds and options.

frankcrawford:
Yes, I understand and agree, although I will do a copr build of all the likely OSs soon and see how it goes.

PR #951, opened 2023-11-23
Improve regression speed

ikus060:
## Changes done and why
To enhance the speed of regression, consider verifying the original file's metadata instead of restoring the data. This approach accelerates the regression process by validating the file signature without the need for a full restoration. Given that reading is typically faster than writing, this significantly improves the efficiency of the regression process.

## Self-Checklist


ikus060:
Hello @ericzolf 

Here is a very nice improvement to the regression process to avoid running the restoration process on every file during the regression. This provide a huge advantages for slow drives (e.g.: USB Stick). With this code, the regression speed is 10 time faster.

As I'm not confident enough in my changes, plz kindly review the modification.

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -2306,11 +2306,15 @@ def _restore_orig_regfile(self, rf):
             "Metadata path '{mp}' can only be regular file.".format(
                 mp=rf.metadata_rorp))
         if rf.mirror_rp.isreg():
-            tf = rf.mirror_rp.get_temp_rpath(sibling=True)
-            tf.write_from_fileobj(rf.get_restore_fp())
-            tf.fsync_with_dir()  # make sure tf fully written before move
-            rpath.copy_attribs(rf.metadata_rorp, tf)
-            rpath.rename(tf, rf.mirror_rp)  # move is atomic
+            # Before restoring file from history, let check if the file are already identical.
+            if rf.metadata_rorp.has_sha1() and rf.mirror_rp.getsize() == rf.metadata_rorp.getsize() and rf.metadata_rorp.get_sha1() == hash.compute_sha1(rf.mirror_rp):
```
ericzolf, line 2310: I would prefer to have the size comparison before all the hash related ones, just cosmetic.

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -2306,11 +2306,15 @@ def _restore_orig_regfile(self, rf):
...
+                rpath.copy_attribs(rf.metadata_rorp, rf.mirror_rp)
```
ericzolf, line 2311: This isn't atomic anymore, I'm hesitating how far it's a big issue, and if the behavior is acceptable in all cases, or if we should make it optional.

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -2306,11 +2306,15 @@ def _restore_orig_regfile(self, rf):
             "Metadata path '{mp}' can only be regular file.".format(
                 mp=rf.metadata_rorp))
         if rf.mirror_rp.isreg():
-            tf = rf.mirror_rp.get_temp_rpath(sibling=True)
-            tf.write_from_fileobj(rf.get_restore_fp())
-            tf.fsync_with_dir()  # make sure tf fully written before move
-            rpath.copy_attribs(rf.metadata_rorp, tf)
-            rpath.rename(tf, rf.mirror_rp)  # move is atomic
+            # Before restoring file from history, let check if the file are already identical.
```
ericzolf, line 2309:
```suggestion
            # Before restoring file from history, check if the versions are already identical.
```

ericzolf (changes requested):
Few comments but the approach looks sound to me.

Also consider that your commit doesn't contain any of NEW or CHG (also hesitating which one fits), and didn't we have an issue around this topic? Your commit message should close it.

ericzolf:
You did something strange with your commits. I think, I fixed it, but whatever you did, it wasn't good.
The patch itself was very good though :smile: Thanks, you can or I'll merge once the last round of tests is done.
If you do it, please make sure the sqashed commit message is proper (and contains a `NEW:` line with an understandable message fit for a lambda user).

PR #952, opened 2023-11-26
Make sure tox.ini doesn't write if called without venv

### LINKED ISSUE #949: [ENH] make tox.ini only write to readable places with --current-env
## What is it about?

When trying to add a `%check` option to RPM spec file, using the current `tox.ini` file it attempts to write to a directory that it does not have permission to do, i.e. `open("{envsitepackagesdir}/coverage.pth","w")`

Current github version, i.e. 2.4.1+

## rdiff-backup call

Adding the following to rdiff-backup RPM fails:

```
%check
%tox
```

Causes a failure on writing:

```
+ /usr/bin/python3 -m tox --current-env -q --recreate -e py311
rdiff-backup 2.4.1.dev14+gf27110de3
  File "<string>", line 1, in <module>
PermissionError: [Errno 13] Permission denied: '/usr/lib/python3.11/site-packages/coverage.pth'
py311: exit 1 (0.01 seconds) /builddir/build/BUILD/rdiff-backup-weekly> python -c 'with open("/usr/lib/python3.11/site-packages/coverage.pth","w") as fd: fd.write("import coverage; coverage.process_startup()\n")' pid=1235
  py311: FAIL code 1 (0.19 seconds)
  evaluation failed :( (0.21 seconds)
error: Bad exit status from /var/tmp/rpm-tmp.HDMejv (%check)
    Bad exit status from /var/tmp/rpm-tmp.HDMejv (%check)

```

The question is `tox.ini` suitable for use for an RPM `%check`, or what can be used, and should it be attempting to write to that directory?

ericzolf:
## Changes done and why
NEW: tox.ini supports --current-env as normal user, avoiding to write in root-protected paths, closes #949

## Self-Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ MANIFEST
 *[._]local[._]*
 
 # setup.py installation files
-src/rdiff_backup.egg-info/
+*.egg-info/
 .eggs/
 
 # Windows build file
--- a/testing/api_test.py
+++ b/testing/api_test.py
@@ -17,17 +17,19 @@ class ApiVersionTest(unittest.TestCase):
         Globals.api_version["actual"] = 201
         info = Globals.get_runtime_info()
 
-        # because the current test will have a different call than rdiff-backup itself
-        # we can't compare certain keys
+        # because the current test will have a different call than rdiff-backup
+        # itself, we can't compare certain keys
         self.assertIn("exec", out_info)
         self.assertIn("argv", out_info["exec"])
         out_info["exec"].pop("argv")
         info["exec"].pop("argv")
-        # info['python']['executable'] could also be different but I think that
-        # our test environments make sure that it doesn't happen, unless Windows
-        if os.name == "nt":
-            info["python"]["executable"] = info["python"]["executable"].lower()
-            out_info["python"]["executable"] = out_info["python"]["executable"].lower()
+
+        # info['python'] could also be different, under special conditions
+        self.assertIn("python", out_info)
+        self.assertIn("executable", out_info["python"])
+        out_info.pop("python")
+        info.pop("python")
+
         self.assertEqual(info, out_info)
 
     def test_default_actual_api(self):
@@ -41,7 +43,7 @@ class ApiVersionTest(unittest.TestCase):
         self.assertEqual(out_info["exec"]["api_version"]["actual"], api_version["max"])
 
     def test_debug_output(self):
-        """we use verbosity 9 only to cover debug functions in logging"""
+        """Use verbosity 9 to cover debug functions in logging"""
         output = subprocess.check_output(
             [RBBin, b"-v", b"9", b"--terminal-verbosity", b"9", b"info"]
         )
--- a/testing/coverage_pth.py
+++ b/testing/coverage_pth.py
@@ -1,0 +1,31 @@
+# write the hook file which will make sure that coverage is loaded
+# also for sub-processes, like for "client/server" rdiff-backup
+
+import os
+import site
+import sys
+
+for site_path in sys.path:
+    coverage_pth = os.path.join(site_path, "coverage.pth")
+    if os.path.isfile(coverage_pth):
+        print("The file {} exists already".format(coverage_pth))
+        sys.exit(0)
+
+if site.ENABLE_USER_SITE:
+    site_packages = [site.getusersitepackages()]
+else:  # we're probably in a virtualenv
+    site_packages = site.getsitepackages()
+
+for site_path in site_packages:
+    coverage_pth = os.path.join(site_path, "coverage.pth")
+    try:
+        os.makedirs(site_path, exist_ok=True)
+        with open(coverage_pth, "w") as fd:
+            fd.write("import coverage; coverage.process_startup()\n")
+            print("The file {} has been written".format(coverage_pth))
+            sys.exit(0)
+    except OSError as exc:
+        print("Couldn't write the file {} due to {}".format(coverage_pth, exc))
+
+print("The coverage.pth file could be written nowhere")
+sys.exit(1)
--- a/testing/coverage_pth.py
+++ b/testing/coverage_pth.py
@@ -14,7 +14,7 @@ for site_path in sys.path:
 if site.ENABLE_USER_SITE:
     site_packages = [site.getusersitepackages()]
 else:  # we're probably in a virtualenv
-    site_packages = site.getsitepackages()
+    site_packages = reversed(site.getsitepackages())
 
 for site_path in site_packages:
     coverage_pth = os.path.join(site_path, "coverage.pth")
--- a/tools/coverage_pth/MANIFEST.in
+++ b/tools/coverage_pth/MANIFEST.in
@@ -1,0 +1,1 @@
+include **/*.pth
--- a/tools/coverage_pth/pyproject.toml
+++ b/tools/coverage_pth/pyproject.toml
@@ -1,0 +1,21 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "coverage-pth"
+version = "0.0.1"
+authors = [
+    {name = "The rdiff-backup project", email = "rdiff-backup-users@nongnu.org"}
+]
+description = "write the hook file which will make sure that coverage is loaded"
+
+[tool.setuptools]
+package-dir = {""="src"}
+include-package-data = true
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.setuptools.package-data]
+"*" = ["*.pth"]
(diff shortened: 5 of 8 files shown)


frankcrawford:
@ericzolf okay, good news is that that patch seems to work fine and it runs %check step.
Bad news is that it fails one or more of the checks.
```
Error processing line 1 of /builddir/.local/lib/python3.11/site-packages/coverage.pth:

    File "<frozen site>", line 186, in addpackage
    File "<string>", line 1, in <module>
  ModuleNotFoundError: No module named 'coverage'

Remainder of file ignored
test_utils_simpleps (__main__.UtilsPsutilTest.test_utils_simpleps)
verify that hardlinked files can be rotated, see issue #272 ... ERROR

ERROR: test_utils_simpleps (__main__.UtilsPsutilTest.test_utils_simpleps)
verify that hardlinked files can be rotated, see issue #272
  File "/builddir/build/BUILD/rdiff-backup-weekly/testing/utils_simpleps_test.py", line 31, in test_utils_simpleps
    self.assertIn(simpleps._get_pid_name_ps(os.getpid()).split(".")[0],
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builddir/build/BUILDROOT/rdiff-backup-2.4.1.dev14+gf27100d-1.fc38.x86_64/usr/lib64/python3.11/site-packages/rdiffbackup/utils/simpleps.py", line 54, in _get_pid_name_ps
    output = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 1026, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.11/subprocess.py", line 1950, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'ps'

Ran 1 test in 0.001s

FAILED (errors=1)
/usr/lib64/python3.11/site-packages/coverage/inorout.py:507: CoverageWarning: Module rdiff_backup was never imported. (module-not-imported)
  self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")
py311: exit 1 (0.08 seconds) /builddir/build/BUILD/rdiff-backup-weekly> coverage run testing/utils_simpleps_test.py --verbose --buffer pid=11059
  py311: FAIL code 1 (30.18 seconds)
  evaluation failed :( (30.21 seconds)
error: Bad exit status from /var/tmp/rpm-tmp.IJHwNx (%check)
    Bad exit status from /var/tmp/rpm-tmp.IJHwNx (%check)
```

However, I also noticed that the `ModuleNotFoundError: No module named 'coverage'` occurs in most of the tests, but does not cause a fatal error.

ericzolf:
Mmh, of course, coverage must be installed...

frankcrawford:
Okay, I'll guess `coverage` would come in as a `BuildRequires`.

I've never played with `tox`, etc before, so am learning about how this testing works.

However, the rest of the test failure does seem to indicate an issue about something else (`ps`) not being found, which is strange, because it exists, maybe another RPM that needs to be pulled in.

ericzolf:
OK, this raises a few points:

1. tox.ini calls `coverage` which each command, and that doesn't seem to be an issue, but then coverage.pth doesn't find the `coverage` module. How does this difference happen?
2. add coverage to the BuildRequires, yes, probably necessary.
3. the `ps` command isn't found, which means that
    a. the (optional) `psutil` module isn't installed, which is fine (it's optional)
    b. but also that the `ps` utility isn't in the PATH, what is its value in a Fedora build environment? Or is `ps` really not installed in the environment?
    c. you need to "BuildRequire" either python3-psutil or procps-ng and/or add it to the PATH (or I need to call `/usr/bin/ps` instead of simply `ps`).
4. you really need to tell me how to progress, there seems to be a lot of assumptions not holding in your packaging environment.
5. if you prefer, or additionally, I'm welcoming Fedora packaging instructions for dummies to reproduce your packaging environment so that I can test myself what's exactly happening.

frankcrawford:
Sorry, still working through this, but:

1. `coverage` is automatically included when I included `tox` for checking.  I am trying to work out why coverage.pth doesn't fine the module.  I think it may be a lib vs lib64 issue.
2. Not actually necessary, it is done automatically
3. Yes, I've added a `BuildRequires: /bin/ps` as it was not automatically pulled into the `chroot`'ed environment.
4. Working on why `/builddir/build/BUILD/rdiff-backup_testfiles/stattest1` doesn't exist.  It seems to be done externally to the rest of the setup.
5. Probably what I currently need to find is a successful `tox` run, so I can compare it to the results I am getting.

I'm fine just working through it, I'm learning a lot.

frankcrawford:
@ericzolf well, I kind of know what is wrong with `coverage` now.  When `coverage.pth` is run, not all site paths are setup.  If I replace it with `import sys; print('\n'.join(sys.path))` I get
```
/usr/lib64/python311.zip
/usr/lib64/python3.11
/usr/lib64/python3.11/lib-dynload
/builddir/.local/lib/python3.11/site-packages
```
but at a later stage I get
```
/usr/lib64/python311.zip
/usr/lib64/python3.11
/usr/lib64/python3.11/lib-dynload
/builddir/.local/lib/python3.11/site-packages
/usr/lib64/python3.11/site-packages
/usr/lib/python3.11/site-packages
```

I don't know when or how the extras get added, but will explore that further.

And obviously `coverage` is in `/usr/lib64/python3.11/site-packages`.

ericzolf:
Few things and thoughts (in a time crush) before I forget:

1. the file `/builddir/build/BUILD/rdiff-backup_testfiles/stattest1` needs to be unpacked, check `.github/workflows` on how we do it.
2. I could place the pth file at the end of the paths list, instead of the start, but I'm not sure it would solve all possible combinations.
3. https://github.com/fedora-python/tox-current-env states that dependencies aren't installed, which explains a lot of our issues

I'm thinking if we shouldn't create a tox_smoke.ini without coverage and dependency on https://github.com/rdiff-backup/rdiff-backup-filesrepo

frankcrawford:
Okay, will look at the workflow and can probably workout some spec file setup equivalent, but I am also thinking that the full test suite may be overkill for the RPM build anyway.  A cut-down version may be a better option.

Also, this isn't really too urgent as it is more for either the weekly or future dev builds, not the current release.

ericzolf:
OK, I added a `tox_smoke.ini` file, so that you should be able to call `tox -c tox_smoke.ini ...` to have only tests without coverage and without need for pre-existing test files. Let me know if this works.

frankcrawford:
Thanks, I will try it in the next few days.

frankcrawford:
@ericzolf  I tried it, and in general it works, but there are some issues with `tox_smoke.ini`.

Firstly, the script `testing/verbosity_actions_test.sh` still uses `coverage run` for the tests (it works, but doesn't propagate the failures.

Secondly, `testing/api_test.py` fails as follows:

```
AssertionError: {'exec': {'version': '2.4.1.dev14+gf27110de3', 'api_[296 chars]-8'}} != {'exec': {'api_version': {'actual': 201, 'default': [249 chars]38'}}
  {'exec': {'api_version': {'actual': 201,
                            'default': 201,
                            'max': 201,
                            'min': 201},
            'parsed': None,
            'version': '2.4.1.dev14+gf27110de3'},
-  'python': {'executable': '/builddir/build/BUILD/rdiff-backup-weekly/.tox/py312/bin/python',
+  'python': {'executable': '/usr/bin/python3',
              'name': 'cpython',
              'version': '3.12.0'},
   'system': {'fs_encoding': 'utf-8',
              'platform': 'Linux-6.6.5-x86_64-with-glibc2.38'}}
```

Which is correct, since it is using a local version of `python` rather than the system one.

ericzolf:
I've removed the shell script from the ini file, it doesn't add much to smoke tests.
Regarding the python executables, it's a bit weird because I'd expect the test script to run with the same Python as rdiff-backup, but in your case it's a wrong assumption. Anyway, I've simplified the test so that it should pass, as long as the python executable contains `python`. And as I write these lines, I realize that it might not be a good idea under Windows, because of PyInstaller... Running out of time now, I'll check later.

ericzolf:
My suspicion with Windows proved true, hence yet another fix. Let me know if it works now @frankcrawford

frankcrawford:
Okay, updated my spec file with the new tests and it completes fine.  Note, that it basically applies the diff between your test branch and a weekly version from a few weeks ago.

I've attached the build log for you to quickly look through to see if there is any issue you can see that I missed.

[build.log](https://github.com/rdiff-backup/rdiff-backup/files/13778026/build.log)

Issue #862, opened 2023-03-01
[ENH] Add an option to exclude a filesystem by type (primarily targeted at NFS)


danci1973:
## Is your feature request related to a problem? Please describe.

We often use 'temporary' NFS mounts for ease of work and people tend to forget to un-mount at the end of the day, so with the next nightly backup, these mounts are backed up as well - and if these are large (they usually are), this is a problem.

## Describe the solution you'd like

An extra option to tell rdiff-backup not to backup files that are on NFS (or some other specified filesystem type) would be useful.

Something like `--exclude-nfs` or `--exclude-filesystem-type xyz`.

## Describe alternatives you've considered

None really.

ikus060:
Hello @danci1973 

Did you give a try to `--exclude-other-filesystems`

According to the man page:
```
Exclude files on file systems (identified by device number) other than the file system the root of the source directory is on.
```

That might fit your need.

danci1973:
Often servers have different paths like/opt, /home, /srv etc. mounted as separate (yet local) filesystems and one would want to backup all of them. That easiest way is by specifying the source directory as '/' - not needing to list each mount point separately.

Using `--exclude-other-filesystems` would skip all of those, whereas using `--exclude-filesystem-type` one could exclude special filesystems (proc, sysfs, tmpfs) and also NFS, while not needing to worry about listing all 'data' filesystems.

ericzolf:
We can think about it, in the meantime, excluding the mount point should do the job.

ericzolf:
Well, we can agree, but Windows also knows the concept of mount points, if not as common as under Linux. See https://learn.microsoft.com/en-us/windows-server/storage/disk-management/assign-a-mount-point-folder-path-to-a-drive

ikus060:
> Well, we can agree, but Windows also knows the concept of mount points, if not as common as under Linux. See https://learn.microsoft.com/en-us/windows-server/storage/disk-management/assign-a-mount-point-folder-path-to-a-drive

Thanks for the reference, I will give it a shot. I tried with other mean, like OneDrive and Google Drive, and the st_dev was the same.

PR #953, opened 2023-11-29
Add an option to exclude a filesystem by type #862

ikus060:
## Changes done and why

NEW: Add option `--exclude-filesystem-type`

## Self-Checklist


src/rdiff_backup/selection.py
```diff
@@ -193,9 +193,11 @@ def _init_parsing_mapping(self):
             "--exclude": (self._glob_get_sf, Select.EXCLUDE),
             "--exclude-regexp": (self._regexp_get_sf, Select.EXCLUDE),
             "--exclude-if-present": (self._presence_get_sf, Select.EXCLUDE),
+            "--exclude-filesystem-type": (self._filesystem_type_get_sf, Select.EXCLUDE),
             "--include": (self._glob_get_sf, Select.INCLUDE),
             "--include-regexp": (self._regexp_get_sf, Select.INCLUDE),
             "--include-if-present": (self._presence_get_sf, Select.INCLUDE),
+            "--include-filesystem-type": (self._filesystem_type_get_sf, Select.EXCLUDE),
```
ericzolf, line 200: I'm not through the whole code but shouldn't it be a Select.INCLUDE?

src/rdiff_backup/selection.py
```diff
@@ -457,6 +459,26 @@ def _filelist_globbing_get_sfs(self, filelist_fp, inc_default, list_name):
...
+    def _filesystem_type_get_sf(self, fstype, include):
+        """Return selection function matching filesystem type."""
+        assert include == Select.EXCLUDE or include == Select.INCLUDE, (
+            "Include is {ival}, should be 0 or 1.".format(ival=include))
+        # FIXME Raise an exception on Windows ?
+        # Since it's too costly to get the filesystem type of each file, will identify the mountpoint using devloc
+        psutil = self.rpath.conn.psutil
+        assert psutil, "psutil must be installed"  # FIXME Raise an exception is psutil is not installed ?
+        partitions = self.rpath.conn.psutil.disk_partitions(all=True)
+        devlocs = [self.rpath.conn.os.lstat(p.mountpoint).st_dev for p in partitions if p.fstype == fstype]
```
ericzolf, line 471: I don't want new direct remote use of Python functions. They need to be wrapped in a new rpath function, which breaks the API 201 BTW (but so does the remote call to psutil). The remote call needs to be marked as API function, it needs to be documented, and the value bumped.

ericzolf (commented):
I'm not yet finished with my review, but I'm a bit questioning if it's worth the effort.

src/rdiff_backup/selection.py
ericzolf: If possible, wrap both lines into one remote function, then you'll have only one remote call instead of two (psutil and lstat), this should improve the performance.

src/rdiff_backup/selection.py
```diff
@@ -457,6 +459,26 @@ def _filelist_globbing_get_sfs(self, filelist_fp, inc_default, list_name):
             else:
                 yield self._glob_get_sf(line, inc_default)
 
+    def _filesystem_type_get_sf(self, fstype, include):
+        """Return selection function matching filesystem type."""
+        assert include == Select.EXCLUDE or include == Select.INCLUDE, (
+            "Include is {ival}, should be 0 or 1.".format(ival=include))
```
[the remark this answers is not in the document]
ericzolf:
```suggestion
            "Include is {ival}, should be {exc} or {inc}.".format(ival=include, exc=Select.EXCLUDE, inc=Select.INCLUDE))
```
That's what I roughly meant.

ikus060:
Maybe we could discuss about ways to avoid breaking API all the time. I know we brought this discussion a couple of time but it seams to be one of the biggest request we receive from users. On my side, I don't need it this feature because I decided to use a different approach where each version are installed side by side.

My understanding is rdiff-backup make direct calls to python libraries (os.unlink, os.chmod) most of the time and other calls to rdiff-backup modules (conn.rpath.make_file_dict, rpath.copy_reg_file). The list of function calls and their behaviour is what make the API. Every time we change or add a new function or change the behaviour of one function, it break the API.

So as long as we keep those function "stable", interoperability should work across different rdiff-backup version. 

For refactoring reason, if too many changes occur, we could create shim layer for compatibility. To allow moving stuff around, we could use annotation similar to what you did with `@API`. 

```
@API(200, 'rpath.setdata_local')
def setdata_local(rp):
   ...
```

Using the "agreed" API version, we could identify the API functions available or not on the other side. Like for this PR, we could check if the API version >= 203, if not we report to the user this feature is missing and only available with API version 203.

In addition, the `@API` annotation could also be improve to integrate the security feature.

If you agree with the direction, I could try to do a proof-of-concept.

ericzolf:
I'm puzzled now, because I think we've had those discussions already in the past, and they have been decided (and I don't re-opening discussions without very good reasons):

1. yes, there is already today the possibility to put gates and validate the API version before using a specific function.
This is what has been done in version 2.2 (API 201) to be compatible with 2.0 (API 200).
2. as per discussion on the mailing list and vote, we've decided to break the API one more time in order to allow me to not get hindered by the past, and create a clean/modern/simpler base to do a fresh start. This is what I'm currently doing in three steps:
    a. remove the old CLI keeping the new one only (tag v2.4.0.unreleased, done!)
    b. remove the API 200 and modernize the build infra (pyproject.toml, etc), while keeping API 201 (tag v2.6.0.unreleased, in progress). Idea is to make the code as simple and modern as possible without killing the compatibility, making hence the risk to break something in this and the next step smaller.
    c.  break API 201 (probably v3.0.0 and API 300), to get better encapsulation, clearer API, parallelization and/or AsyncI/O implemented or at least readyness, etc... My objective is really to make extension of rdiff-backup as simple as possible for the future.
3. I offered to keep the compatibility at the price of slower development (and more complex code), but a majority decided against it. That there is still a minority against it, is an issue but nothing we can account for, it's an either/or decision. This doesn't mean that we shouldn't strive to keep the API compatibility post-v3, at least for the foreseeable future. It was also part of the considerations that not releasing v2.4 and v2.6 was a gesture towards the people seeking for stability over innovation, as v3.0 will come out much later than those 2.x versions would have, keeping 2.2 the stable version until then.
4. That `@API` looks like a decorator isn't by chance, that was always the idea to use it to implement the Security module, but it's something for v3, not before.
5. your PR is also something for v3 as I don't see the point of creating an API 202 which will never get released.

Does it all make sense or did I miss some of your arguments?

ericzolf:
No pressure at all but I just wanted to note that you could now rebase and continue your PR before the changes in the code base become too big.

ikus060:
@ericzolf Yeah I Know. I'm a bit busy lately with other projects. :(

ikus060:
Closing as won't do.

PR #954, opened 2023-11-30
Fix regression of long filename

ikus060:
## Changes done and why

This Pull request include two changes.

One change is not related to regression of long filename but rather to robustness of the regression when text file already exists.

The other one is the actual fix that return the "index" of the filename based on metadata which seams more appropriate since we are traversing the tree of files based on the ordering of the file in metdata.

The change also include a testcase to reproduce the problem and get it fixed.

## Self-Checklist

- [-] changes to the code have been reflected in the documentation


ericzolf (approved):

ericzolf:
I made the fix message more understandable but else got job!

ikus060:
@ericzolf Thanks for your help !

PR #1019, opened 2024-09-30
Flush stderr or stdout when logging

ikus060:
## Changes done and why

I have this scenario where output of rdiff-backup is scramble because, it seams, the stdout and stderr are printed to the console in different order. To get this fixed, I recommand to flush stderr and/or stdout on every log. This also push the lines to the console in real-time.

## Self-Checklist

- [-] changes to the code have been reflected in the documentation: N/A
- [-] changes to the code have been covered by new/modified tests: N/A
- [ ] commit contains a description of changes relevant to users prefixed by DOC:, FIX:, NEW: and/or CHG: I'm open to recommendation.

### DIFF
--- a/src/rdiff_backup/log.py
+++ b/src/rdiff_backup/log.py
@@ -121,6 +121,7 @@ class Logger:
             )
         else:
             termfp.write(tmpstr)
+        termfp.flush()
 
     def conn(self, direction, result, req_num):
         """Log some data on the connection
--- a/testing/server.py
+++ b/testing/server.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import sys
+from rdiffbackup.singletons import specifics
 
 __doc__ = """
 
@@ -26,5 +27,6 @@ except (OSError, ImportError):
     print_usage()
     raise
 
+specifics.server = True
 Security._security_level = "override"
 sys.exit(connection.PipeConnection(sys.stdin.buffer, sys.stdout.buffer).Server())


ikus060:
@ericzolf I will need your help on this. I don't understand why my changes is making all the test fail.

ericzolf:
Let me check, the weekly build also failed, it could be that the error isn't yours!

ericzolf:
This add nothing to do with the weekly issue. See #1020 

It appears that the flush breaks the communication between server and client (which happens through stdin _and_ stdout of the server). I don't fully understand why, because the place you modified has only to do with stderr of the server, but there is probably a side effect I don't get.

This said, I'm not sure I get your issue: isn't it a mismatch between client and server output, rather than between stderr and stdout? We're talking about messages being in the wrong order, not messages being intermingled (which I've never seen)? Furthermore I am the opinion that a newline flushes anyway the buffer (which explains why messages aren't intermingled).

Sorting stderr of the server process, which goes directly to stderr of the client through the SSH channel, would require to pass it as a proper message to the client process (through stdout/stdin), pickling it and then unpickling it. This would have probably a lot of impact on the time required, compared to the current unprocessed handling through stderr.

ikus060:
Hello Eric,

Thanks for you feedback. I will try to debug this issue a bit more based on you statement.

The issue I faced:

While executing rdiff-backup on a computer, where stdout is redirected to a file and stderr is redirected to stdout. Sometime, the stderr of the client get printed before the stderr of the server. I assumed is a flush problem and I assumed it doesn't hurt to force the flush of stderr.

After some testing, it seams the change I made is fixing the issue, but the problem is hard to reproduce.

I will do further testing.

ikus060:
@ericzolf This should be ready for review. I fixed the unit test.

ericzolf:
The change looks OK, I'll probably approve but this timeout issue on our pipeline is killing me (and this is a bad joke, because I suspect the issue in the `kill_test.py`). I'll try to fix this before merging this PR.

Before I do so, one more question: the flushing doesn't impact the performance?

ikus060:
Thanks for merging.

ikus060:
Seams you fix for `kill_test` is not working after all. :(

ericzolf:
Two things:

1. you didn't answer  my question about performance
2. and, yeah, disappointing: I think, I found the right issue/place, but not the right solution yet. Apparently the backup step gets so stuck (for reasons I don't know but seems rather new) that it can't even get killed (SIGKILL), so that the regress step after fails because the backup is still apparently running. Before, the pipeline was waiting forever for the first step to get terminated (SIGTERM), I added the kill step in my last change, but that even wasn't enough.

PR #1022, opened 2024-10-06
Add timeout when reading files not on disk

ikus060:
## Changes done and why

When using cloud file storage, some time the file is not present on disk and the software is supposed to fetch the data from the cloud in real-time. When this process is failing, this cause rdiff-backup to be stuck trying to read a file indefinitely. To avoid this, a timeout is added to read the first bytes of the file.

I'm creating this pull request to start a discussion about the change. Technically, the change I've done is working. Rdiff-backup is no longer stuck on file read when OneDrive on MacOS is failing. But:

1. The additional os.lstat() double the number of IO call which I would prefer to avoid, but when I try adding "blocks" to `make_file_dict` everything fall apart in comparison. And I'm not 100% sure if adding "blocks" to the dict is backward compatible.
2. I think we could make this timeout value a settings. What do you suggest ? 300sec (5min) might not be enough in all condition. If the file is large and OneDrive need to download the file, it might take more then 5 min. Not sure how to detect this. We could do some timeout calculation based on the file-size reported. 

## Self-Checklist

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -47,6 +47,8 @@ from rdiffbackup.utils import usrgrp
 from rdiffbackup.locations.map import owners as map_owners
 from rdiffbackup.meta import acl_posix, acl_win, ea
 
+import signal
+
 try:
     import win32api
     import win32con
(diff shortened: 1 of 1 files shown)


ikus060:
Closing. After extensive testing, this doesn't do any good.

PR #1029, opened 2024-10-13
Provide more information in tox calls

ericzolf:
## Changes done and why

Replace rdiff-backup --version with rdiff-backup info, especially because I had some doubts which version of Python is actually used in our pipeline


ericzolf:
As I feared, something changed and the root user doesn't see by default the newly installed python, meaning we're currently only testing against the system Python (3.10 at time of writing).

PR #1030, opened 2024-11-05
Split Globals into different singletons

ericzolf:
## Changes done and why

Split the Globals module into different singletons depending on the scope of the variables:

1. consts for constants (renamed to be all caps)
2. specifics for variables where the value can be different depending on client vs server
3. generics for variables which need to have the same value across all instances of rdiff-backup

This led to uncovering a few unused variables, which were removed.

## Self-Checklist

### DIFF
--- a/src/rdiff_backup/SetConnections.py
+++ b/src/rdiff_backup/SetConnections.py
@@ -522,23 +522,28 @@ def _test_connection(conn_number, rp):
     print("Testing server started by: ", __conn_remote_cmds[conn_number])
     conn = specifics.connections[conn_number]
     if conn is None:
-        sys.stderr.write("- Connection failed, server tests skipped\n")
+        log.Log("Connection failed, server tests skipped", log.ERROR)
         return False
-    # FIXME the tests don't sound right, the path given needs to pre-exist
-    # on Windows but not on Linux? What are we exactly testing here?
     try:
-        remote_time = conn.Globals.get("current_time")
-        assert (
-            remote_time == generics.current_time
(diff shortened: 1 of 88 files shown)


ikus060:
Wow, so many changes. I did not review each individual changes. it's basically a refactoring of Globals ?

ericzolf:
Correct, that's what it is. Do you think, you'll find the time to review?

src/rdiff_backup/SetConnections.py
```diff
@@ -517,25 +519,30 @@ def _test_connection(conn_number, rp):
     # the function doesn't use the log functions because it might not have
     # an error or log file to use.
     print("Testing server started by: ", __conn_remote_cmds[conn_number])
-    conn = Globals.connections[conn_number]
+    conn = specifics.connections[conn_number]
     if conn is None:
-        sys.stderr.write("- Connection failed, server tests skipped\n")
+        log.Log("Connection failed, server tests skipped", log.ERROR)
```
ikus060, line 524: Any reason to use `log.Log()` instead of `sys.stderr.write()` ? 

I wonder why we have a mixture with `print()` too...

ikus060 (approved):

src/rdiff_backup/SetConnections.py
ericzolf: Only consistency, I want to have everything going through the log framework. And the `print`'s were an oversight from my side, fixed in the just added commit. Thanks.

ericzolf:
These randomly failing tests are really annoying, but anyway... Merging.

PR #1033, opened 2024-11-17
Continue simplifying singletons

ericzolf:
## Changes done and why

- get rid of preserve_atime which wasn't used
- introduce typing in singletons
- mypy added to test dependencies for typing check

## Self-Checklist


ericzolf:
A bit of ranting as I wait for the test results...
The recurrent issue with the pipeline taking too long under Linux seems to be due to some effect regarding the killing and/or subprocesses done in the `testing\kill_test.py`.
Where I feel like ranting is that there seems to be a different handling of such topics in the pipeline vs. locally, which makes extremely hard to properly debug in reasonable time. Being able to recreate the conditions of the pipeline would be a big help but I haven't seen any considerations that it can be possible.
And https://github.com/orgs/community/discussions/141132 hasn't be helpful either...

ericzolf:
I can't really believe that I've actually fixed the whole issue but the pipeline did run successfully twice without timeout, so who knows.

Issue #1027, opened 2024-10-11
[BUG] Need to handle OSError EDEADLK/BlockingIOError on MacOS


ikus060:
## Bug summary

I'm facing a situation where the filesystem on MacOs is raising an OS Error 11 on read operation. It's most likely an issue with OneDrive on MacOS. Such a pain.

```
WARNING: Exception '[Errno 11] Resource deadlock avoided' raised of class '<class 'BlockingIOError'>':
  File "rdiff_backup/robust.py", line 94, in check_common_error
  File "rdiff_backup/rpath.py", line 1586, in copy
  File "rdiff_backup/rpath.py", line 1620, in copy_reg_file
  File "rdiff_backup/rpath.py", line 1152, in write_from_fileobj
  File "rdiff_backup/rpath.py", line 1537, in copyfileobj
  File "rdiff_backup/rpath.py", line 1514, in read
  File "rdiff_backup/iterfile.py", line 495, in read

* Sending back exception '[Errno 11] Resource deadlock avoided' of type <class 'BlockingIOError'> with traceback   File "rdiff_backup/connection.py", line 509, in _answer_request
  File "rdiff_backup/backup.py", line 197, in patch
  File "rdiff_backup/rorpiter.py", line 146, in __call__
  File "rdiff_backup/backup.py", line 600, in fast_process_file
  File "rdiff_backup/backup.py", line 655, in _patch_to_temp
  File "rdiff_backup/backup.py", line 702, in _patch_snapshot_to_temp
  File "rdiff_backup/robust.py", line 94, in check_common_error
  File "rdiff_backup/rpath.py", line 1586, in copy
  File "rdiff_backup/rpath.py", line 1620, in copy_reg_file
  File "rdiff_backup/rpath.py", line 1152, in write_from_fileobj
  File "rdiff_backup/rpath.py", line 1537, in copyfileobj
  File "rdiff_backup/rpath.py", line 1514, in read
  File "rdiff_backup/iterfile.py", line 495, in read

* Cleaning up
```

Reading more about this error code, I understand that we should better handle this error and effectively try again when this is raised.

## Version, Python, Operating System

rdiff-backup: 2.2.6
Python 3.
MacOS

## rdiff-backup call

Called trough Minarca. But it should not make a difference.

## What happened and what did you expect?

[...]

PR #1035, opened 2024-11-19
Handle Cross-Platform OSError.errno Mismatch #1027

ikus060:
FIX: issue with OSError.errno discrepancies when the source and destination are on different platforms.

## Changes done and why

Serialize and deserialize OSError differently to handle variations in error number values across platforms. Use the error code name to ensure cross-platform compatibility when transporting the error code.

@ericzolf Let me know what you think about this way to implement the fix.
I will need some time to reproduce the problem on a MacOS to ensure everything is working as expected.

## Self-Checklist

### DIFF
--- a/src/rdiff_backup/connection.py
+++ b/src/rdiff_backup/connection.py
@@ -24,6 +24,7 @@ import subprocess
 import sys
 import time
 import traceback
+import errno
 
 from rdiff_backup import (
     iterfile,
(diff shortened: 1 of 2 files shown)


src/rdiff_backup/connection.py
```diff
@@ -588,7 +594,11 @@ def _extract_exception(self):
                 ),
                 log.INFO,
             )
-        return sys.exc_info()[1]
+        result = sys.exc_info()[1]
```
ericzolf, line 597: This part should be _before_ the call to `robust.is_routine_fatal` because the condition might also raise an OSError, which should be extended with errno_str as well. I don't think it'll make a big difference, because the error is fatal, but just to be consistent.

src/rdiff_backup/connection.py
```diff
@@ -588,7 +594,11 @@ def _extract_exception(self):
...
+        # OSError code are specificto the platform. Send back code as string.
+        if isinstance(result, OSError) and result.errno:
+            result.errno_str = errno.errorcode.get(result.errno, "EUNKWN")
```
ericzolf, line 600:
```suggestion
            result.errno_orig = result.errno
```
Just to facilitate a potential cross-platform debugging session, so that it's clear if a conversion of errno happened.

ericzolf (changes requested):
I originally would have preferred an own version of OSError, but that would have been too complicated, so this is fine. Just two small comments.

ericzolf (approved):

PR #1037, opened 2024-11-24
Yet another trying to fix the pipeline stuck during killing tests

ericzolf:
## Changes done and why

As the subject says...

## Self-Checklist


ericzolf:
OK, I don't know if I've fixed it but it doesn't appear again. Possibly adding a few waiting times in the kill tests was all what was needed. We'll see, ready for review.

testing/kill_test.sh
```diff
@@ -0,0 +1,6 @@
+# script to test with specific verbosity and parameters
```
ikus060, line 1: Do you still need this file ? I'm not sure what is the use of it.

ericzolf: Good point, I've renamed the file and added some explanations to make it more generically usable, I think it could be useful again in the future.

ikus060: Great and your changes seams to be working. Yeah !

PR #1063, opened 2025-04-25
NEW: Add selection parser "--exclude-unreadable"

ikus060:
## Changes done and why

- Add implementation for a new selector to ignore unreadable files.
- Make use of this new selector in fs_abilities to avoid raising permissions error

## Self-Checklist

### DIFF
--- a/docs/rdiff-backup.1.adoc
+++ b/docs/rdiff-backup.1.adoc
@@ -351,6 +351,8 @@ This doesn't apply recursively though so that the _filename_ must be present in 
 
 --min-file-size _sizeinbytes_:: Exclude files that are smaller than the given size in bytes.
 
+--exclude-unreadable:: Exclude files that are not readable by rdiff-backup based on their ACLs, to avoid permission errors during backup.
+
 == STATISTICS OPTIONS
 
 --file-statistics, --no-file-statistics::
--- a/src/rdiff_backup/selection.py
+++ b/src/rdiff_backup/selection.py
@@ -191,6 +191,7 @@ class Select:
             "exclude-device-files": (self._devfiles_get_sf, Select.EXCLUDE),
             "exclude-sockets": (self._sockets_get_sf, Select.EXCLUDE),
             "exclude-fifos": (self._fifos_get_sf, Select.EXCLUDE),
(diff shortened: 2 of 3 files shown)


ikus060:
@ericzolf If you agree with the principle, I will write a unit test to cover the new "--exclude-unreadable" selector.

ikus060:
@ericzolf Sorry for the long list of small commits. At least everything is split into manageable chunks.

The general idea here is to avoid calling os.lstat() on files that are excluded by --exclude ... and not re-included by any other pattern.

1. This reduces the number of I/O calls.
2. When done properly, it prevents permission errors or access denied issues when trying to retrieve metadata from excluded folders.

Your input is appreciated, as always.

Just a heads-up: I’m heading off on vacation and will be back in late May.

ericzolf:
Before I review this PR, can you clarify why the PR merges into a branch which isn't master?

ericzolf:
I need to take a deeper look at the changes you've made, they're quite extensive, but one thought goes through my mind: why do we need an exclude-unreadable option? If a file is unreadable, we won't anyway be able to backup it (and rdiff-backup wouldn't choke on it, just send a warning). Is it only about avoiding warnings? Then I'd like another option and even approach, but we can talk about once you've clarified the purpose.

ikus060:
> I need to take a deeper look at the changes you've made, they're quite extensive, but one thought goes through my mind: why do we need an exclude-unreadable option? If a file is unreadable, we won't anyway be able to backup it (and rdiff-backup wouldn't choke on it, just send a warning). Is it only about avoiding warnings? Then I'd like another option and even approach, but we can talk about once you've clarified the purpose.

Thanks, Eric, for taking the time to review.

To answer your question—yes, the main goal is to avoid raising unnecessary warnings during the backup process. That was the initial motivation for this change.

However, while working on it, I noticed that rdiff-backup calls lstat() even on files that are excluded. This shouldn't be necessary when the exclusion is based on filename patterns. In my scenario, this has a negative impact on performance because lstat() is still called on every file, even those that are excluded.

ericzolf:
Could we close this PR and wait for the implementation of #1078 ?

PR #1068, opened 2025-05-01
Use surrogateescape for list files & compare output

ikus060:
## Changes done and why

- Change implementation of log.py to send unicode directly to stdout using surrogateescape, fallback to backslashreplace.
- Use surrogateescape in list files output
- Use surrogateescape in compare output

## Self-Checklist


ericzolf:
I have only one general request: use everywhere the utils functions `to_safe_bytes` and `to_safe_str` and make sure we consistently use the same escape mechanism(s). It'll be more work but will simplify maintenance.

ikus060:
I've rebase my changes on Master.

PR #1087, opened 2025-11-30
Merge statistics into calculate action

ericzolf:
## Changes done and why

As the subject states, get rid of `rdiff_backup.run_stats` aka `rdiff-backup-statistics` and have instead `rdiff-backup calculate statistics`.

At the same time, I had to:

* Add StoredRPath class to cover increments methods in repository, removing them from RPath
* Split rdiffbackup.singletons.stats in sstats (session) and fstats (file)
* Add most of the run_stats code to fstats

## Self-Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -39,7 +39,7 @@ psutil
 
 tox
 flake8
-black
+black < 26
 bandit
 coverage
 mypy
--- a/requs/test.txt
+++ b/requs/test.txt
@@ -3,7 +3,7 @@
 
 tox
 flake8
-black
+black < 26
 bandit
 coverage
 mypy


ericzolf:
Documentation has been uploaded, ready for review!

src/rdiffbackup/locations/increment.py
```diff
@@ -20,13 +20,93 @@
...
+        """
+        Returns list of increments whose name starts like the current file
+        """
+        dirname, basename = self.dirsplit()
+        parent_dir = self.__class__(self.conn, dirname, ())
+        if not parent_dir.isdir():
+            return []  # inc directory not created yet
+
+        inc_list = []
+        for filename in parent_dir.listdir():
```
ikus060: I'm raising the point here. Maybe we could tackle this later on. But calling `listdir()` on repository with a lot of increment files will make forever to execute the calculate command.

My brain is not a compiler, but this gets called at least 2 times in a row to compute the statistics. It should preferably be called a single time to reduce I/O.

ericzolf: I want first to get it merged as is, we always can improve at a later stage. Please  create an issue for this if you don't mind.

Issue #1071, opened 2025-05-25
[ENH] don't return a warning when there are no increments to remove


TimothyPMann:
## Bug summary

I don't need a warning if I ask to remove old increments and there are no increments that old. A note is okay, but a note and two warnings is rather excessive. If some people need this warning for some reason, perhaps there could be an option to suppress it. I'd use that so that my cron job doesn't send me an email every day full of useless warnings.

## Version, Python, Operating System

Call `rdiff-backup info` and replace the following line with the output,
repeat for each environment impacted, especially if using remote operations:

```yaml
---
exec:
  api_version:
    actual: 0
    default: 200
    max: 201
    min: 200
  argv:
  - /usr/bin/rdiff-backup
  - info
  parsed: null
  version: 2.2.6
python:
  executable: /usr/bin/python3
  name: cpython
  version: 3.12.3
system:
  fs_encoding: utf-8
  platform: Linux-5.15.0-140-lowlatency-x86_64-with-glibc2.39
...
```

## rdiff-backup call

How did you call rdiff-backup, with which parameters:

```
sudo rdiff-backup --api-version 201 remove increments --older-than 1Y /extra/backup/boot
```

## What happened and what did you expect?

$ sudo rdiff-backup --api-version 201 remove increments --older-than 1Y /extra/backup/boot
NOTE:    No increments older than Sat May 25 14:56:07 2024 found, exiting.
WARNING: No increment is older than '1Y'
WARNING: Action remove emitted warnings, see previous messages for details

I would have expected no warnings, as the NOTE is enough.

## More information

If you can, please repeat the action leading to the error, adding the option `-v9` and attach the output to this bug report.

ericzolf:
Works as designed, so it's not a bug, and I don't promise that it'll get implemented, but valid enhancement request.

ikus060:
That applies to all warnings. My honest opinion: I don't care about the warning in my use case. I usually completely ignore the error code 2. In Minarca or any other automation, I simply swallow the error code 2.

That said, a good middle ground would be a new flag `--dont-fail-on-warnings` or the other way around a `--fail-on-warning` for all action.

baikiet:
As follow up on [tentative fix](https://github.com/rdiff-backup/rdiff-backup/pull/1101), could the maintainer(s?) please clarify what shape a satisfactory fix would look like?

[...]

PR #1101, opened 2026-03-29
Remove spurious warning when attempting to delete non-existing increment.

baikiet:
Both `RepoShadow._get_removal_time` and `RepoShadow.remove_increments_older_than` emit similar messages when attempting to delete an increment that does not exist. The former is only ever called by the latter. In one case it's a WARNING and in the other a NOTE.

## Changes done and why

Removes the NOTE from `RepoShadow._get_removal_time` and turns the WARNING from `RepoShadow.remove_increments_older_than` into a NOTE as requested in https://github.com/rdiff-backup/rdiff-backup/issues/1071
This is normal operation and is not worthy of a WARNING, let alone a WARNING and a NOTE.

AFAICT the return value or log messages in this situation are not documented so I have not modified the documentation but I have ticked the relevant box in this PR description as per developer documentation.

## Self-Checklist

### DIFF
--- a/src/rdiffbackup/locations/_repo_shadow.py
+++ b/src/rdiffbackup/locations/_repo_shadow.py
@@ -1230,7 +1230,7 @@ class RepoShadow(location.LocationShadow):
         if removal_time < 0:  # no increment is old enough
             log.Log(
                 "No increment is older than '{ot}'".format(ot=time_string),
-                log.WARNING,
+                log.NOTE,
             )
             return consts.RET_CODE_WARN
 
@@ -1411,12 +1411,6 @@ class RepoShadow(location.LocationShadow):
             ]
         times_in_secs = [t for t in times_in_secs if t < action_time]
         if not times_in_secs:
-            log.Log(
-                "No increments older than {at} found, exiting.".format(
-                    at=Time.timetopretty(action_time)
-                ),
-                log.NOTE,
-            )
             return -1
 
         times_in_secs.sort()
--- a/testing/action_remove_test.py
+++ b/testing/action_remove_test.py
@@ -206,7 +206,7 @@ class ActionRemoveIncsTest(ActionRemoveTest):
                 b"remove",
                 ("increments", "--older-than", "30000"),
             ),
-            consts.RET_CODE_WARN,
+            consts.RET_CODE_OK,
         )
         self.assertRegex(
             comtst.rdiff_backup_action(


ericzolf (changes requested):
Thanks for your submission, first, it's very much appreciated.

I don't think it works as you expect it currently but:

1. I need  to do a more thorough review
2. I'm not yet sure I agree with the principle of the change

baikiet:
> Thanks for your submission, first, it's very much appreciated.
> 
> I don't think it works as you expect it currently but:
> 
>     1. I need  to do a more thorough review
> 
>     2. I'm not yet sure I agree with the principle of the change

Thanks for checking. I have not been able to setup a proper development environment for rdiff-backup (tests are meant to run within Docker?!) and have not been able to properly test this change. But if you disagree with the principle, that's pointless anyway. I will follow up on the bug.

ericzolf:
You can use containers and the Makefile, but I simply use a virtual environment (virtualenv or venv).
The whole document could use some refactoring but this should get you started: https://github.com/rdiff-backup/rdiff-backup/blob/master/docs/DEVELOP.adoc#install-for-test-and-development

And, yeah, I'm not yet convinced: people who want a change are more vocal about it, but there might be more user happy with the current situation.

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -1230,7 +1230,7 @@ def yield_files(rp):
         if removal_time < 0:  # no increment is old enough
             log.Log(
                 "No increment is older than '{ot}'".format(ot=time_string),
-                log.WARNING,
+                log.NOTE,
             )
```
ericzolf, line 1234: You can actually remove this log entry, it's duplicate with the one in `_get_removal_time`

src/rdiffbackup/locations/_repo_shadow.py
```diff
@@ -1411,12 +1411,6 @@ def _get_removal_time(cls, time_string, show_sizes):
             ]
         times_in_secs = [t for t in times_in_secs if t < action_time]
         if not times_in_secs:
-            log.Log(
-                "No increments older than {at} found, exiting.".format(
-                    at=Time.timetopretty(action_time)
-                ),
-                log.NOTE,
-            )
```
ericzolf, line 1419: I would keep this one because the function has all log messages anyway, so better to keep them in one place

src/rdiffbackup/locations/_repo_shadow.py
ericzolf, line 1403: This is not related to the issue at hand, but `_get_removal_time` returning `None` isn't handled in the calling function: rdiff-backup should return in error.
And we should have tests covering those use cases, I'm surprised that we don't (but I didn't check explicitly).

ericzolf (changes requested):
Do you mind taking care of the additional issues I found? Let me know if you need help with the testing.

ericzolf:
In order to help newcomers like you, I've added a specific section in #1106 - feel free to review it and give me your feedback.

ikus060:
@ericzolf what is the status of this PR. I recall you ask the mailing list about this change: WARNING vs no-WARNING.

I still think returning a WARNING error code if the correct way to handle this. I would honestly not merge this PR.

That said, I think we could make everybody happy by adding a simple `--success-on-warning` to the cli to return RET_CODE_OK instead of RET_CODE_WARN. Simple to implement and cover all use cases.

ericzolf:
I must admit that I lost track but let me check, I won't merge anyway without proper testing from the author, this is the only way IMHO to avoid AI slop (not that I would think that it is the case here).

PR #1103, opened 2026-04-10
Avoid remote system and rpath calls

ericzolf:
## Changes done and why

All system and most rpath functions are now called locally, not through the connection.
I didn't have to do a lot, most calls were unnecessarily remote, as showed by a (now commented) debug message.
It definitely makes the code simpler and should slightly improve performance.

## Self-Checklist


ikus060 (approved):
Bravo !

ericzolf:
Thanks for the quick review, merging!

PR #1105, opened 2026-04-18
Reduce remote calls in rpath

ericzolf:
## Changes done and why

`make_file_dict` becomes `get_rpath_data` and is the only one being called remotely, whereas `setdata_local` becomes `_add_rpath_metadata` called locally to improve performance when gathering metadata.
Note that the __fspath__ for QuotedRPath was wrong as it should represent the (quoted) path on the file system.

## Self-Checklist

### DIFF
--- a/src/rdiff_backup/rpath.py
+++ b/src/rdiff_backup/rpath.py
@@ -743,9 +743,7 @@ class RPath(RORPath):
 
     def setdata(self):
         """Set data dictionary using the wrapper"""
-        self.data = self.conn.rpath.make_file_dict(self.path)
-        if self.lstat():
-            self.conn.rpath.setdata_local(self)
+        self.data = self.conn.rpath.get_rpath_data(self.path, self.base, self.index)
 
     def chmod(self, permissions, loglevel=log.WARNING):
         """Wrapper around os.chmod"""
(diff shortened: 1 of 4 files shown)


src/rdiff_backup/rpath.py
```diff
@@ -1704,26 +1702,21 @@ def rename(rp_source, rp_dest):
...
-# @API(make_file_dict, 200)
-def make_file_dict(filename):
-    """Generate the data dictionary for the given RPath
-
-    This is a global function so that os.name can be called locally,
-    thus avoiding network lag and so that we only need to send the
-    filename over the network, thus avoiding the need to pickle an
-    (incomplete) rpath object.
+# @API(get_rpath_data, 300)
+def get_rpath_data(rp):
```
ikus060: This implementation is confusing to me.

As it's a getter, I would expect it to return a value. Period. But it also has a side effect of updating "rp". I'm not sure if it's the best implementation for that. I believe the function should either return a dict() with the new data. Or update rp.data in place.

src/rdiff_backup/rpath.py
```diff
@@ -743,9 +743,7 @@ def __setstate__(self, rpath_state):
 
     def setdata(self):
```
ikus060, line 744: I've raised this point before, and since you're updating code in this area, it might be a good time to address it.
During backup execution, RPath objects are frequently created without ever querying the file metadata. This is particularly common during filename filtering (include/exclude), where paths are evaluated purely on their name.
The problem is that setdata() is called in __init__(), meaning every RPath instantiation triggers an I/O call to retrieve file metadata — even when that metadata is never used. This results in a significant number of wasteful I/O calls over the course of a backup.

Two approaches worth considering:

1. Lazy loading: defer the setdata() call until the metadata is first accessed, for example via a property or a __getattr__ hook ?
2. Explicit loading: remove the setdata() call from __init__() entirely and require callers to invoke it explicitly when metadata is needed.

Either approach would reduce unnecessary I/O.

src/rdiff_backup/rpath.py
ericzolf: Good point. I tried to optimize too much. Reverted!

src/rdiff_backup/rpath.py
ericzolf: I haven't forgotten but this isn't a small thing and will rather complicate the code. I'm not ready for that and wanted first anyway to use asyncIO to optimize the metadata reading.

PR #1112, opened 2026-05-29
FIX: log level in fs_abilities chmod() calls

gpziemba:
Calls to chmod() specified loglevel "4," which is not a valid log level, so logging caused an exception in _format() when it tried to access _LOG_PREFIX[4].

This fix replaces instances of "4" with "log.INFO"

- [*] changes to the code have been reflected in the documentation
- [*] commit contains a description of changes relevant to users prefixed by DOC:, FIX:, NEW: and/or CHG:

### DIFF
--- a/src/rdiffbackup/locations/fs_abilities.py
+++ b/src/rdiffbackup/locations/fs_abilities.py
@@ -535,7 +535,7 @@ class FSAbilities:
         test_rp = rp.append("dir_inc_check")
         test_rp.touch()
         try:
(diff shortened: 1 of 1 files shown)


gpziemba:
I don't think there are any new test cases needed for this fix, but please let me know if otherwise. Cheers!

ericzolf:
I agree that tests aren't required. Thanks for your contribution, I'll merge once the pipeline has gone through.

PR #1117, opened 2026-08-05
WIP: Refactor tox files

ikus060:
## Changes done and why

To follow a more conventional approach, I want to reduce duplication in our tox.ini files. The target is to use more conventional target names like "lint", "format", "test", etc. And also get rid of the makefile which essentially only call tox.

## Self-Checklist

### DIFF
--- a/src/rdiff_backup/selection.py
+++ b/src/rdiff_backup/selection.py
@@ -765,7 +765,15 @@ probably isn't what you meant""".format(se=self.selection_functions[-1].name))
         """Return list of regexps equivalent to prefixes of glob_str"""
 
         glob_parts = glob_str.split(b"/")
-        if b"" in glob_parts[1:-1]:  # "" OK if comes first or last, as in /foo/
+
+        # Allow a leading empty part for UNC paths like //server/share/foo
+        # on Windows, but still catch consecutive slashes anywhere else.
+        if os.name == "nt" and glob_str.startswith(b"//"):
+            check_parts = glob_parts[2:-1]
+        else:
+            check_parts = glob_parts[1:-1]
+
+        if b"" in check_parts:  # "" OK if comes first or last, as in /foo/
             raise GlobbingError(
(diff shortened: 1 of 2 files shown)


ericzolf:
Few things:

1. You seem again to have forked off an early version of my previous PR.  Are you sure you didn't break your local master? In all cases, this makes review difficult...
2. I didn't look into the details (see above) but I'm not sure I want to change, it "works for me" and I work with the setup the most. I would appreciate if you would talk to me before doing such changes. e.g. create an issue first before investing time and effort?
3. This said, it doesn't mean I won't like it, it's just that I haven't understood (yet) what the advantages and disadvantages of what you suggest would be.

ikus060:
@ericzolf 
As describe, the goal is to merge all the tox files that are redundant. In addition, I want to remove the make file which is very obsolete in modern project. The goal is to simplify the tooling to run the test and make it more aligned with the convention.

ericzolf:
I got this, my question was more, why should I care? The best standard is the one you're used to and why should I change my habits?

ikus060:
@ericzolf The idea is really to simplify the developement of rdiff-backup.

As I'm working on it once in a while, every time I come back working on rdiff-backup it's painful. Here are some of the pain point.

1. The presence of the make file.
     * It's not cross-platform
     * required non-python tooling 
2. The make file is configure by default to run with Docker. Why ?
3. Missing target to "format" to source code

I care less about various version of tox.ini, but I think it make sens to merge everything into a single tox file in order to minimize duplication.

ericzolf:
OK, PR is closed but just to complete the discussion:

1. Makefile and Docker: I don't mind removing them, I don't use them and they are not even declared as dependency, but those were introduced by @ottok - I would check with him if those are pre-requisites for Debian/Ubuntu packaging.
2. Target to format - you mean calling `black`? Why not, as non-default.
3. Various tox files - I don't think there is a lot of overlap, and one big file is generally more difficult to understand and maintain than multiple smaller ones, so there is a tradeoff.

PR #1119, opened 2026-08-06
Use sys.platform instead of os.name to help mypy

ericzolf:
## Changes done and why

Use sys.version_info and sys.platform to differentiate respectively Python versions and operating systems because this is what Mypy understands according to https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks
The trigger for this was that check-static fails under Windows.

## Self-Checklist


ikus060 (approved):

ikus060:
@ericzolf 

I did not change the call to ` os.name == "nt"` in my commit. This probably belong in your commit.

See selection.py line 771.

PR #1122, opened 2026-08-30
Fix spurious re-transfer of files whose owner is unknown on the mirror

### LINKED ISSUE #1121: [BUG] Files with owner names unknown on the destination are re-transferred when metadata is read from the filesystem
## Bug summary

When the destination has to be read from the file system because no `mirror_metadata` is available (`WARNING: Could not find mirror metadata file. Metadata will be read from filesystem instead`), every file whose owner or group name cannot be resolved on the destination host is considered changed: rdiff-backup computes a signature, transfers a (tiny) delta, rewrites the whole mirror file and stores a ~1 KB `.diff` increment for it. In my setup the source is a NAS whose files belong to LDAP accounts (e.g. `alice@example.org`, group `staff@example.org`) and the destination has no LDAP client, so almost the whole repository (~13 TB) was being rewritten; only `root:root` files compared equal. The same repository with metadata present behaves normally.

Cause: `RORPath.__eq__` (`src/rdiff_backup/rpath.py`, the `uname`/`gname` branch) only tolerates an unresolved name on `other`, not on `self`. In `_sigs_iterator` the expression is `src_rorp == dest_rorp`; when the destination iterator comes from the file system, `dest_rorp` is an `RPath`, a sub-class of `RORPath`, and Python evaluates the reflected `__eq__` of the sub-class first (this happens even though `RPath` only inherits `__eq__`), i.e. with the destination as `self`. Its name is `None`, the source name is a string → "different". With metadata, both sides are `RORPath`, `self` is the source, and the destination rorp carries the name recorded from the source at the previous session, so the comparison passes.

Minimal reproduction in Python with the shipped classes:

```python
from rdiff_backup import Globals, rpath, connection
Globals.process_uid = 0
dest = rpath.RPath(Globals.local_connection, b"/path/to/mirror/file")   # RPath from the file system
rpath.setdata_local(dest)                                                # uname/gname -> None if unresolvable
src = rpath.RORPath(dest.index, dict(dest.data, uname="alice@example.org", gname="staff@example.org"))
print(src == dest)                       # False  (used by _sigs_iterator)
print(rpath.RORPath.__eq__(src, dest))   # True   (what the metadata path effectively does)
```

Destination (repository), Debian 12:

```yaml
exec:
  api_version: {actual: 0, default: 200, max: 201, min: 200}
  version: 2.2.2
python: {executable: /usr/bin/python3, name: cpython, version: 3.11.2}
system: {fs_encoding: utf-8, platform: Linux-6.1.0-21-amd64-x86_64-with-glibc2.36}
```

Source (client), Synology DSM, SynoCommunity package:

```yaml
exec:
  api_version: {actual: 0, default: 200, max: 201, min: 200}
  version: 2.2.6
python: {executable: /usr/local/rdiff-backup/env/bin/python3, name: cpython, version: 3.12.11}
system: {fs_encoding: utf-8, platform: Linux-4.4.302+-x86_64-with-glibc2.36}
```

The code in current `master` (`b286548`) is unchanged in this respect.

## rdiff-backup call

```
rdiff-backup --exclude-other-filesystems --include-symbolic-links --preserve-numerical-ids \
    --no-compression-regexp "..." --exclude-device-files --exclude-fifos --exclude-sockets \
    --exclude-regexp "..." \
    --remote-schema 'ssh -i KEY %s sudo /usr/bin/rdiff-backup --exclude-other-filesystems --restrict /path/to/repo --server' \
    /path/to/source backup@backuphost::/path/to/repo
```

After an interrupted session had left the repository without usable `mirror_metadata`, the next backup fell back to reading the mirror from the file system and started rewriting practically every file (tens of GB written per minute locally, ~1 KB `.diff.gz` increments for multi-MB files whose size, mtime, uid, gid, permissions and content were identical). Expected: only files that actually differ are transferred, as happens when metadata is present. Files owned by `root:root` (name resolvable on both hosts) were correctly left alone, which is how the owner-name comparison was identified.

## More information

Similar symptom reported in #1080, but there the names resolve on both sides and metadata is present, so the cause seems different.

A fix making the name comparison symmetric (only compare when both sides resolved the name), with a regression test, follows as a pull request.

ramok:
Fixes #1121.

When a mirror is read from the file system (no `mirror_metadata` available), the destination side of the comparison in `_sigs_iterator` is an `RPath`, a sub-class of `RORPath`, so Python evaluates `src_rorp == dest_rorp` with the destination as `self` (the reflected-operand rule applies even though `RPath` does not override `__eq__`). The `uname`/`gname` branch of `RORPath.__eq__` only tolerated an unresolved name on `other`, so every file whose owner or group name resolves on the source but not on the destination (e.g. LDAP accounts) was considered changed and fully re-written on every backup — ~13 TB out of 13.5 TB in the setup where this was found.

The fix makes the name comparison symmetric: names are only compared when both sides could resolve them (`None` or the legacy string `"None"` counts as unresolved on either side). `equal_loose` (which ignores names by design) and the test-only `_equal_verbose` are intentionally unchanged.

The new `RORPEquality` test class covers the unresolved-name case in both dispatch orders (`RORPath == RPath` and against a plain `RORPath`), the legacy `"None"` string, resolved-but-different names, and that ignoring names does not hide real mtime/size/uid/perms changes. `testUnresolvedName` fails without the fix.

- [x] The changes are covered by new tests (`testing/rpath_test.py`, class `RORPEquality`)
- [x] The commit contains a user-relevant description prefixed with `FIX:` (single line, `closes #1121`)
- [ ] Documentation: not applicable (no user-visible interface change)

`tox -e check-static` passes (flake8, black, bandit, mypy).


src/rdiff_backup/rpath.py
```diff
@@ -163,8 +163,20 @@ def __eq__(self, other):
         for key in frozenset(self.data.keys()) - ignored_keys:
             if key == "uname" or key == "gname":
                 # here for legacy reasons - 0.12.x didn't store u/gnames
+                # names are only compared if both sides could resolve them,
+                # whichever side is self: Python evaluates the comparison
+                # with a sub-class instance (e.g. an RPath mirror read from
+                # the file system) as self even though RPath doesn't
+                # override __eq__
```
ericzolf, line 170:
```suggestion
                # names are only compared if both sides can resolve them.
```
I wouldn't comment on derived classes, a parent shouldn't know or care about derived classes.

testing/rpath_test.py
```diff
@@ -41,6 +41,71 @@ def testPickle(self):
...
+
+    def _get_source_rorp(self, **changes):
+        """Return a rorpath as sent by the source, with resolved names"""
+        data = dict(self.mirror_rp.data)
+        data["uname"] = "someone@ldap.example.com"
+        data["gname"] = "somegroup@ldap.example.com"
+        data.update(changes)
+        return rpath.RORPath(self.mirror_rp.index, data)
+
+    def testUnresolvedName(self):
```
ericzolf, line 65:
```suggestion
    def test_unresolved_name(self):
```
Let's try to be consistent even though I know that we aren't yet...

testing/rpath_test.py
```diff
@@ -41,6 +41,71 @@ def testPickle(self):
...
+            self.assertTrue(b == a)
+            self.assertFalse(a != b)
+            self.assertFalse(b != a)
+        # the legacy string "None" (stored by 0.12.x) counts as unresolved too
+        self.mirror_rp.data["uname"] = "None"
+        self.mirror_rp.data["gname"] = "None"
+        self.assertTrue(src_rorp == self.mirror_rp)
+        self.assertTrue(self.mirror_rp == src_rorp)
+
+    def testResolvedNames(self):
```
ericzolf, line 83:
```suggestion
    def test_resolved_names(self):
```

testing/rpath_test.py
```diff
@@ -41,6 +41,71 @@ def testPickle(self):
...
+        # a mirror file read from the file system, e.g. when metadata is missing
+        self.mirror_rp = out_rp.append("rorp_equality")
+        self.mirror_rp.write_string("some content")
+        # the mirror side can't resolve the names, like a host without LDAP
+        self.mirror_rp.data["uname"] = None
+        self.mirror_rp.data["gname"] = None
+
+    def _get_source_rorp(self, **changes):
+        """Return a rorpath as sent by the source, with resolved names"""
+        data = dict(self.mirror_rp.data)
```
ericzolf, line 59:
```suggestion
        data = self.mirror_rp.data.copy()
```
Slightly more explicit, and that's how getRORPath() does it.

testing/rpath_test.py
```diff
@@ -41,6 +41,71 @@ def testPickle(self):
...
+        self.mirror_rp.data["uname"] = "root"
+        self.mirror_rp.data["gname"] = "root"
+        src_rorp = self._get_source_rorp()
+        self.assertFalse(src_rorp == self.mirror_rp)
+        self.assertFalse(self.mirror_rp == src_rorp)
+        src_rorp = self._get_source_rorp(uname="root", gname="root")
+        self.assertTrue(src_rorp == self.mirror_rp)
+        self.assertTrue(self.mirror_rp == src_rorp)
+
+    def testOtherAttribs(self):
```
ericzolf, line 94:
```suggestion
    def test_other_attribs(self):
```

testing/rpath_test.py
```diff
@@ -41,6 +41,71 @@ def testPickle(self):
...
+        self.mirror_rp.write_string("some content")
+        # the mirror side can't resolve the names, like a host without LDAP
+        self.mirror_rp.data["uname"] = None
+        self.mirror_rp.data["gname"] = None
+
+    def _get_source_rorp(self, **changes):
+        """Return a rorpath as sent by the source, with resolved names"""
+        data = dict(self.mirror_rp.data)
+        data["uname"] = "someone@ldap.example.com"
+        data["gname"] = "somegroup@ldap.example.com"
```
ericzolf, line 61: Not super important in the context but just curious: the uname and gname wouldn't contain the `@ldap.example.com`, would they? Wouldn't they get resolved to a plain Linux user or group name, like someone and somegroup?

ericzolf (changes requested):
Looks good to me and sorry for the (holiday driven) delay to react. Few rather cosmetic changes required. I'll start now the pipeline...

